Issue #64, opened 2014-11-07
Conversion of Json::Value to long int is ambiguous


orodbhen:
Trying to copy a Json::Value object to a long int fails to compile, because Json::Uint64 and Json::Int64 are mapped to unsigned long int and long int, respectively.

On Linux/Unix systems, at least, these should instead be mapped to uint64_t and int64_t, which will always be 64 bit.

Also note that these typedefs (int64_t and uint64_t) don't work either with the Json::Value constructor, because they are typdefed to long int when __WORDSIZE == 64. They do work when __WORDSIZE == 32, because then they map to long long int, just like Json::Int64.

orodbhen:
just checking back to see if any work has been done on this. I find jsoncpp to be very unfriendly to programming with templates.

Currently, I use a wrapper I created that maps the type of the output buffer to the correct jsoncpp type. It's original purpose was to provide a C interface to jsoncpp, but I use it in C++ as well.

It would be nice, though, to have a more generalized interface for C++ that works with templates. I'd like to contribute, but first I need to know how the maintainers want to implement it.

cdunn2001:
Maybe some sort of wrapper? You can derive your own CharReader and StreamWriter to operate on your data-types. But if you want to templatize `Json::Value`, you might be better off forking the whole project.

TuNguyen89:
Today I'm get the different result of function ```Value::UInt64 Value::asUInt64() const```

I have ```Value["key"] = 18446744073709551616```   (Max unsigned 64 bit + 1)
Then the Value["key"] converted to real.
When I use the function value.asUInt64() I get different value for the same compiler but different architect (amd64 vs arm64)

In amd64, 2^64 is returned.
In arm64,  0 is return.

PR #66, opened 2014-11-07
Json::Value constructor: Accept any arithmetic type.

BillyDonahue:
Allow construction of Json::Value from size_t, short, etc.
Fixes issue #64


cdunn2001:
Since the main branch is strictly C++11 now, will this solution work? What about MSVC?

BillyDonahue:
In C++11 a much more elegant version of this code would be possible.
So I'd rewrite it for the main branch using the C++11 common subset supported by MSVC.
How far back into MSVC do you need to support? Their C++11 compliance is a moving target.

cdunn2001:
2013 at the earliest. I'm even fine with 2015. Folks with a problem can always use an older release.

At some point, you might have to break some API-compatibility. In that case, put it on a 2.y.z branch. I expect most C++11 changes will break binary compatibility anyway.

Hmmm. I really need to get the hidden implementation done quickly, so it will not conflict with other new work. But just do what you think best.

hjmjohnson:
@BillyDonahue I'm closing this 4 year old PR.  I think that the move to C++11 makes this PR not consistent with the current code base.

If there are still changes that are needed, please feel free to re-open the PR (preferably rebased on top of the current master branch).

PR #101, opened 2014-12-27
Fix build issues related to Solaris and older GCC

dominicpezzuto:
Fixed two build issues:
 - JsonCPP currently doesn’t compile for Solaris due to platform differences with ‘isfinite’ function.  Fixed by adding proper include and define for Solaris.
 - JsonCPP currently doesn’t compile for GCC version 4.1.2 and earlier due to use of ‘-Werror=*’ compile flag, which was introduced in a later version.  Fixed by adding version check to only add this flag on supported versions of GCC.


cdunn2001:
Thanks. We'll want those changes cherry-picked to `0.7.0` also.

Issue #105, opened 2015-01-05
`target_include_directories` cmake error


godbyk:
Building jsoncpp under Jenkins results in the following output from cmake:

```
+ CC=/usr/bin/clang CXX=/usr/bin/clang++ cmake .. -DCMAKE_BUILD_TYPE=RelWithDebInfo -DJSONCPP_WITH_CMAKE_PACKAGE=ON -DJSONCPP_LIB_BUILD_SHARED=OFF -DCMAKE_CXX_FLAGS=-fPIC -DCMAKE_INSTALL_PREFIX=/var/lib/jenkins/jobs/jsoncpp-linux/workspace/install
-- The C compiler identification is Clang 3.5.0
-- The CXX compiler identification is Clang 3.5.0
-- Check for working C compiler: /usr/bin/clang
-- Check for working C compiler: /usr/bin/clang -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/clang++
-- Check for working CXX compiler: /usr/bin/clang++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- JsonCpp Version: 1.0.0
-- Found PythonInterp: /usr/bin/python2 (found suitable version "2.7.3", minimum required is "2.6") 
-- Configuring done
CMake Error in src/lib_json/CMakeLists.txt:
  Target "jsoncpp_lib" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "/var/lib/jenkins/jobs/jsoncpp-linux/workspace/src/lib_json/../../include"

  which is prefixed in the source directory.
```

This is with cmake version 3.1.0.  There may be a potentially helpful explanation at http://stackoverflow.com/a/25681179/4214.

Gachapen:
Does this only happen while using JSONCPP_WITH_CMAKE_PACKAGE=ON? I think the stackoverflow solution might work. I'll test it soon.

godbyk:
Testing on my local machine instead of the build server: yes, using `JSONCPP_WITH_CMAKE_PACKAGE=OFF` circumvents the issue. After applying your patch from issue #106, cmake stopped complaining even with `JSONCPP_WITH_CMAKE_PACKAGE=ON`.

cdunn2001:
[...]

Issue #104, opened 2015-01-05
Minimum cmake version is too low


godbyk:
The minimum cmake version required is currently set to 2.8.5. However, the `target_include_directories` command (found in `src/lib_json/CMakeLists.txt` lines 53 and 60) requires at least cmake version 2.8.11.

The first occurrence (line 53) was recently worked around by a pull request (commit 8dd32e1e2eb05530d265286a88f5ba9cabf01b3c), but the second occurrence still causes problems for folks using cmake versions between 2.8.5 and 2.8.10.

Options to resolve this are to either increase the minimum cmake version required (see `CMakeLists.txt`, line 1) or add a workaround to line 60 in `src/lib_json/CMakeLists.txt`.

PR #106, opened 2015-01-06
Fix cmake_package install being broken because of wrong include path.

Gachapen:
The TARGET_INCLUDE_DIRECTORIES from inside the
IF(JSONCPP_WITH_CMAKE_PACKAGE) block was removed, since it only needs to
be set once.

JSON_INCLUDE_DIR was also replaced with JSONCPP_INCLUDE_DIR since the former is not defined.

In addition the CMAKE_VERSION check was simplified.

Fixes issue #105 and #104


Gachapen:
Maybe the install command should be tested with the build bot?

cdunn2001:
Sweet.

If you add `install` to [.travis.yml](https://github.com/open-source-parsers/jsoncpp/blob/master/.travis.yml), we'll pull that too.

PR #113, opened 2015-01-15
Integration

datadiode:
Side note: The integration branch is just for my convenience. I don't mean to propose creation of a new branch in the mainline repo.

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -187,7 +187,6 @@ private:
 
   typedef std::deque<ErrorInfo> Errors;
 
-  bool expectToken(TokenType type, Token& token, const char* message);
   bool readToken(Token& token);
   void skipSpaces();
   bool match(Location pattern, int patternLength);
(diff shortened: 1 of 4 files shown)


cdunn2001:
> Side note: The integration branch is just for my convenience. I don't mean to propose creation of a new branch in the mainline repo.

Of course! You're using github properly. Thanks! I'll look at this soon...

cdunn2001:
Please rebase to `master`. You'll just have to fix a couple small conflicts.

datadiode:
`readToken()` has already consumed the first character - be it a sign or a digit - by the time `readNumber()` gains control. A leading `+` is illegal and triggers `readToken()`'s `default:` case.

The goal is to not read beyond a valid token. `sscanf()` tolerates extra characters at the end of its input, so the malformedness remains undetected when `readNumber()` silently skips them.

Existing tests run without issues. It would make some sense to add a test that checks parser behavior with malformed input.

About rebase: I'm currently on subversion which does not seem to offer me that option.
Seems like a good time to start with git.

cdunn2001:
> `readToken()` has already consumed the first character

Oh! Of course! I should have looked at that.

cdunn2001:
I've integrated your fix for `fail31.json`. Let's make a new issue for other changes.

Issue #47, opened 2014-09-26
Value::swap does not swap some members


cout:
Looks like it fails to swap itemIsUsed_, memberNameIsStatic_, and comments_.  If this is intentional, it deserves a comment.

datadiode:
Found this comment in [old code at sourceforge](http://sourceforge.net/p/jsoncpp/code/HEAD/tree/trunk/jsoncpp/include/json/value.h):
/// Swap values.
/// \note Currently, comments are intentionally not swapped, for
/// both logic and efficiency.

cdunn2001:
@datadiode, that comment is [still here](https://github.com/open-source-parsers/jsoncpp/blob/master/include/json/value.h#L241). Maybe that just needs to be expanded.

BillyDonahue:
If it were up to me, I would like the swap to be changed to swap
everything, and not leave the comments behind. The comments are part of the
object's state.

I mean, if I have a std::vector<Value> and I do some sort on it with a
series of swaps, I'd want the values to be completely swapped and not leave
irrelevant comments behind. After the sort, the comments will be attached
to the wrong elements. That's very irregular.

datadiode:
`Value::operator=()` relies on `Value::swap()` to *not* swap the comments, because it is designed to leave comments in place.

The proposed fix is to *not* implement `Value::operator=()` in terms of `Value::swap()`, but in terms of `Value::swapPayload()`. The latter leaves comments (as well as source offsets) in place, while the former is revised to follow standard semantics.

BillyGoto:
operator= should not preserve old comments in the destination. It is
extremely surprising if an algorithm relying on swap produces different
results from an algorithm relying on std::swap, which uses operator=.

sounds like a bug factory.

On Sunday, January 18, 2015, datadiode <notifications@github.com> wrote:

> Value::operator=() relies on Value::swap() to *not* swap the comments,
> because it is designed to leave comments in place.
>
> The proposed fix is to *not* implement Value::operator=() in terms of
> Value::swap(), but in terms of Value::swapPayload(). The latter leaves
> comments (as well as source offsets) in place, while the former is revised
> to follow standard semantics.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/open-source-parsers/jsoncpp/issues/47#issuecomment-70401362>
> .
>


-- 
ǝnɥɐuop ʎllıq

cdunn2001:
[...]

PR #118, opened 2015-01-18
https://github.com/open-source-parsers/jsoncpp/issues/47

datadiode:


cdunn2001:
I don't think we need even a sub-minor bump.

cdunn2001:
@datadiode, you might want to delete your old branches. You can do that in the GUI, or

    git clone git@github.com:datadiode/jsoncpp.git
    cd jsoncpp
    git push origin :integration :47_fix_value_swap

PR #119, opened 2015-01-19
Specialize std::swap() for Json::Value

datadiode:
This addresses [BilliyGoto](https://github.com/BillyGoto)'s comment on https://github.com/open-source-parsers/jsoncpp/issues/47

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -1082,6 +1082,10 @@ public:
 
 } // namespace Json
 
+/// Specialize std::swap() for Json::Value.
+namespace std { void swap(Json::Value&, Json::Value&); }
+inline void std::swap(Json::Value& a, Json::Value& b) { a.swap(b); }
+
 #if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING)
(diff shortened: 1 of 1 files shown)


include/json/value.h
```diff
@@ -1082,6 +1082,10 @@ class JSON_API ValueIterator : public ValueIteratorBase {
 
 } // namespace Json
 
+/// Specialize std::swap() for Json::Value.
+namespace std { void swap(Json::Value&, Json::Value&); }
```
BillyDonahue: This is an illegal declaration according to [namespace.std], because it's declaring a new function in namespace std.

The standard way to define a swap is in the namespace of the type, in this case namespace Json. Then it will be found by ADL, as expected by the Swappable concept.

cdunn2001:
```
commit 9454e687a32bb812092a990e3bc9162077e48e77
Author: datadiode <jochen.neubeck@vodafone.de>
Date:   Tue Jan 20 15:25:04 2015

    Specialize std::swap() for Json::Value in a C++ standard compliant way

    originally from pull #119
```

Issue #116, opened 2015-01-16
Test fails on Windows with a git checkout using dos line endings in comments


rpavlik:
Found this when setting up a new CI system: get these failures in the tests, Windows with VS2013 32-bit building. They pass locally, where the file mentioned has unix endlines, but on the CI system they have dos ones and fail. I think this may have found an actual bug in the comments handling.

cc @godbyk

```
  Failure details:
  * Test C:\jenkins-node\workspace\jsoncpp-win-matrix\BIT\32\VS\12\host\windows\test\data\test_comment_02.json
    Difference in input at line 11:
    Expected: '// Multiline comment cpp-style'
    Actual:   '// Multiline comment cpp-style// Second line'
  
  
  * Test C:\jenkins-node\workspace\jsoncpp-win-matrix\BIT\32\VS\12\host\windows\test\data\test_real_08.json
    Difference in input at line 1:
    Expected: '// Out of 32-bit integer range, switch to double in 32-bit mode. Length the'
    Actual:   '// Out of 32-bit integer range, switch to double in 32-bit mode. Length the// same as UINT_MAX in base 10 and digit less than UINT_MAX's last digit in// order to catch a bug in the parsing code.'
  
  
  * Test C:\jenkins-node\workspace\jsoncpp-win-matrix\BIT\32\VS\12\host\windows\test\data\test_real_09.json
    Difference in input at line 1:
    Expected: '// Out of 64-bit integer range, switch to double in all modes. Length the same'
    Actual:   '// Out of 64-bit integer range, switch to double in all modes. Length the same// as ULONG_MAX in base 10 and digit less than ULONG_MAX's last digit in order// to catch a bug in the parsing code.'
  
  
  * Test C:\jenkins-node\workspace\jsoncpp-win-matrix\BIT\32\VS\12\host\windows\test\data\test_real_10.json
    Difference in input at line 1:
    Expected: '// Out of 32-bit signed integer range, switch to double in all modes. Length'
    Actual:   '// Out of 32-bit signed integer range, switch to double in all modes. Length// the same as INT_MIN in base 10 and digit less than INT_MIN's last digit in// order to catch a bug in the parsing code.'
  
  
  * Test C:\jenkins-node\workspace\jsoncpp-win-matrix\BIT\32\VS\12\host\windows\test\data\test_real_11.json
    Difference in input at line 1:
    Expected: '// Out of 64-bit signed integer range, switch to double in all modes. Length'
[...]

PR #122, opened 2015-01-20
Normalize comment EOLs while reading instead of while writing

mzeren-vmw:
Tests are currently failing when git cloning on Windows with autocrlf = true. In
that setup multiline comments contain \r\n EOLs. The test code assumes that
comments contain \n EOLs and opens the .actual files (etc.) with "wt" which
converts \n to \r\n. Thus we end up with \r\r\n EOLs in the output, which
triggers a test failure.

Instead we should cannonicalize comments while reading so that they contain only
\n EOLs. This approach simplifies other parts of the reader and writer logic,
and requires no changes to the test. It is a breaking change, but probably the
Right Thing going forward.

This change also fixes dereferencing past the end of the comment string in
StyledWriter::writeCommentBeforeValue.

Tests should be added with appropriate .gitattributes for the input files to
ensure that we run tests for DOS, Mac, and Unix EOL files on all platforms. For
now this change is enough to unblock Windows builds.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -152,14 +152,9 @@ bool Reader::readValue() {
   bool successful = true;
 
   if (collectComments_ && !commentsBefore_.empty()) {
-    // Remove newline characters at the end of the comments
-    size_t lastNonNewline = commentsBefore_.find_last_not_of("\r\n");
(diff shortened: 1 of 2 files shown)


mzeren-vmw:
Resolves #116

cdunn2001:
Rebased and merged. Thanks!

Issue #28, opened 2014-08-13
cannot use removeMember because it has only string parameters, not ArrayIndex


cdunn2001:
From Itzik S (privately):
> The following is simple solution.

```
void Value::remove(ArrayIndex index)
{
    if (type_ == arrayValue){
#ifndef JSON_VALUE_USE_INTERNAL_MAP
        CZString key(index);
        ObjectValues::iterator it = value_.map_->find(key);
        if (it != value_.map_->end()){
            ArrayIndex oldSize = size();
            // shift left all items left, into the place of the "removed"
            for (ArrayIndex i=index; i<oldSize-1; i++){
                CZString key(i);
                (*value_.map_)[key] = (*this)[i+1];
            }
            // erase the last one ("leftover")
            CZString keyLast(oldSize-1);
            ObjectValues::iterator itLast = value_.map_->find(keyLast);
            value_.map_->erase(itLast);
        }
#else
        NOT SUPPORTED
#endif
    }
}
```

PR #127, opened 2015-01-20
Remove index

cdunn2001:
See issue #28.
```
/** \brief Remove the indexed array element.

      O(n) expensive operations.
      Update 'removed' iff removed.
      (This is a better pattern than removeMember().)
      \return true iff removed (no exceptions)
  */
  bool removeIndex(ArrayIndex i, Value* removed);
```

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -1019,7 +1019,7 @@ Value Value::removeMember(const std::string& key) {
 }
 
 bool Value::removeIndex(ArrayIndex index, Value* removed) {
-  if (this->type_ != arrayValue) {
+  if (type_ != arrayValue) {
     return false;
   }
 #ifdef JSON_VALUE_USE_INTERNAL_MAP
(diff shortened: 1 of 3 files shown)


cdunn2001:
Any objections?

src/lib_json/json_value.cpp
```diff
@@ -1018,6 +1018,34 @@ Value Value::removeMember(const std::string& key) {
   return removeMember(key.c_str());
 }
 
+bool Value::removeIndex(ArrayIndex index, Value* removed) {
+  if (this->type_ != arrayValue) {
```
BillyDonahue: Is the 'this->' notation really necessary? I find it distracting when accessing data members.
There are some corner cases where you need it, and so seeing it unnecessarily suggests to me that we're in one of those corner cases, and I waste a little attention on it. It's also just making the code a little bloaty, IMO.

src/lib_json/json_value.cpp
```diff
@@ -1018,6 +1018,34 @@ Value Value::removeMember(const std::string& key) {
...
+#else
+  CZString key(index);
+  ObjectValues::iterator it = this->value_.map_->find(key);
+  if (it == this->value_.map_->end()) {
+    return false;
+  }
+  *removed = it->second;
+  ArrayIndex oldSize = this->size();
+  // shift left all items left, into the place of the "removed"
+  for (ArrayIndex i=index; i<oldSize-1; i++){
```
BillyDonahue: Ironically, the prefix ++i notation is more idiomatic in C++, isn't it?
We might want a little whitespace in this expression. Other initializers nearby have spaces around the =, for example.

BillyDonahue:
Looks fine.

src/lib_json/json_value.cpp
cdunn2001: Totally agree. This was copy/paste from a private email. I'll make those changes. Thanks for the review.

src/lib_json/json_value.cpp
cdunn2001: I can understand the preference. However, I've taught myself to use `this` always because
 1. It makes the member access obvious.
 2. More important, I have a lot of experience with re-factoring. 

In fact, at my last start-up, because of some brain-dead decisions before I got there, our code took 10-20 minutes to rebuild, in parallel. (Templates can be very slow, and excessive boost::python compounds that problem, wrapping smart-pointers and boost::serialized classes.) To get work done, I would often write free functions first, to avoid altering headers, and I'd move them into the class later. This was particularly effective because most classes were pure-public (again, not my decision).

Just food for thought. It's not a big deal to me either way. Subjectively, I have a slight preference for consistency between internal and external data-member access. But I can see it your way too. And for this code-base, I'm happy to switch it and go with the team.

I'm just glad that you didn't criticize the return-by-pointer. That *is* [a big deal to me](http://stackoverflow.com/questions/26441220/googles-style-guide-about-input-output-parameters-as-pointers).

cdunn2001:
If it's not an array, should this throw? Or simply return false?

PR #145, opened 2015-01-26
Simplify builder

cdunn2001:
This will make more sense to people.

By using 2 separate Builders, we can keep each API simple. We will agree not to add any fields, so we do not need virtual methods and wrappers. If we need new fields, we can simply provide a new Builder. No big deal.


cdunn2001:
Maybe we can discuss this API before making `1.4.0` an official release?

PR #154, opened 2015-02-04
Ignore empty comments in input.

ya1gaurav:
Empty comments (// or /***/) are unnecessary if present in json input. Current patch ignores empty comments.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -150,7 +150,8 @@ void Value::CommentInfo::setComment(const char* text, size_t len) {
   JSON_ASSERT_MESSAGE(
       text[0] == '\0' || text[0] == '/',
       "in Json::Value::setComment(): Comments must start with /");
-  // It seems that /**/ style comments are acceptable as well.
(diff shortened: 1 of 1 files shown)


cdunn2001:
I think empty comments could be intentional. I'd like to see other opinions before changing this.

ya1gaurav:
Ok. I would also like to know intention behind it :)

cdunn2001:
A placeholder? Consistency?

The comments are meta-data. jsoncpp provides access to the meta-data but does not make strong guarantees about output formatting. I think you're looking for a prettyfier.

ya1gaurav:
During Pretty Print.

cdunn2001:
If you want to supply another pull-request, you could add a setting to `StreamWriterBuilder` which ignores empty comments, or to `CharReaderBuilder` if it's ok to add empty comments programmatically.

But your patch is buggy and insufficient.

1. It crashes when `strlen(text) < 3`
1. It keeps comments which are white-space-only.

I am also not convinced that anyone else would ever use this feature. But if you really want it, that's reason enough. Just submit a better patch.

PR #174, opened 2015-02-15
only throws exceptions JSON_USE_EXCEPTION

kobigurk:
JSON_ASSERT now throws a runtime_error

### DIFF
--- a/include/json/assertions.h
+++ b/include/json/assertions.h
@@ -16,7 +16,7 @@
 #if JSON_USE_EXCEPTION
 #include <stdexcept>
 #define JSON_ASSERT(condition)                                                 \
-  assert(condition); // @todo <= change this into an exception throw
+  if (!(condition)) {throw std::runtime_error( "assert json failed" );} // @todo <= add detail about condition in exception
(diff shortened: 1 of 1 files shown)


cdunn2001:
The only concern is that we could break something somewhere. Seems unlikely, but what do you think?

kobigurk:
I think it's the expected behavior when using JSON_USE_EXCEPTION (it was for me and colleagues, at least). It's also not the "expected behavior" case in which I would consider leaving the assert in production code.
But you have a point here that someone may (although unlikely) rely on it.
Although ugly, maybe we should add another define?

cdunn2001:
I'll go with your advice here. Probably nobody is using `JSON_USE_EXCEPTION` directly, right? I'm not too worried about converting an abort to an exception. I just want to be confident that you and your colleagues have considered all issues, since I don't have time.

kobigurk:
Well, we use it in production for about 1.5 years!
But I think I rushed a bit. I see that you've converted all but one calls to JSON_ASSERT to JSON_ASSERT_MESSAGE which throws exceptions, so it actually might not have a big effect at all (meaning, there was almost nothing to worry about in the first place).

PR #192, opened 2015-03-02
changed length argument of duplicateStringValue from unsigned int to siz...

marklakata:
...e_t, to avoid warnings in Visual Studio.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -52,7 +52,7 @@ const LargestInt Value::maxLargestInt = LargestInt(LargestUInt(-1) / 2);
 const LargestUInt Value::maxLargestUInt = LargestUInt(-1);
 
 /// Unknown size marker
-static const unsigned int unknown = (unsigned)-1;
+static const size_t unknown = (size_t)-1;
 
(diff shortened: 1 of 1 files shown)


cdunn2001:
+1 to eliminating the warnings. However, I am in the middle of a gnarley change to support UTF-8 properly. (Embedded nulls cause problems.) Because we must maintain binary-compatibility, we cannot alter the size of `map<CZString, Value>`. We need to store the length *somewhere* within `CZString` (or, since it's internal, an alternative datatype of the same size). It has a pointer and an enum. `sizeof(enum)` is actually `sizeof(int)`, so we cannot support *keys* of length greater than `UINT_MAX`. We could, in theory, support larger values, but not keys.

Maybe it's ok to use `size_t` for values ([typically 64-bits](http://ivbel.blogspot.com/2012/02/size-of-primitive-types-in-c-language.html)) and to impose the tighter restriction on keys. (In fact, the limit will be `2^30`, because we need 2 bits for something else at the moment.)

I would like to reiterate that these were not my decisions. The basic library was designed long ago, not by me. My goals are backward-compatibilty, bug-fixes, and reasonable new features.

PR #193, opened 2015-03-02
Added MSVC+CMake generated files to .gitignore.

madmaxoft:
These files are generated when compiling in MSVC.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,25 @@
 /doc/doxyfile
 /dist/
 #/include/json/version.h
+
+# MSVC project files:
+*.sln
+*.vcxproj
+*.filters
+*.user
+*.sdf
+*.opensdf
+*.suo
+
+# MSVC build files:
+*.lib
+*.obj
+*.tlog/
+*.pdb
+
+# CMake-generated files:
+CMakeFiles/
+CTestTestFile.cmake
+cmake_install.cmake
+pkg-config/jsoncpp.pc
+jsoncpp_lib_static.dir/


rpavlik:
Usually (in cmake-built projects, I don't know specifically about jsoncpp), an "out of source" build is recommended - one where your source directory and your binary (generated) directory are different, which prevents this from being a problem.

madmaxoft:
Unfortunately that is not possible for MSVC - cmake generates the build files into the source folder even when instructed to make an out-of-source build.

cdunn2001:
Interesting. We'll probably accept that (after rebasing to the new tip). But could you tell us the exact cmake command that you use? I would like to see these files myself.

madmaxoft:
I just re-checked and this time it worked flawlessly, all the build files are in the build folder.

Still, I think this change is useful, since in-source builds are quite usual for MSVC users.

Also, I use this repo as a submodule in another project, which does `add_subdirectory(lib/jsoncpp/)` in its cmake file; and it uses another library that currently doesn't support out-of-source builds (LibEvent), so I do need in-source builds at this moment.

cdunn2001:
Please tell us the cmake command which you are using.

madmaxoft:
For in-source builds: `cmake .`

For out-of-source builds (not working for me due to other libs):
```
mkdir build
cd build
cmake ..
```

cdunn2001:
In general, side-effects of a specific build-system should be ignored in your global `~/.config/git/ignore`. Is there an equivalent on Windows? `C:\Users\MyLogin\.gitconfig`? We don't want to add git-ignores to every possible cmake-based repo everytime cmake supports a new build-system. Do you see the point?

It's easier to make a case for the cmake-generated files. But that includes `Makefile`, and I often use that to aid development. 

To ignore untracked files:
```
git status -uno
```
To remove untracked files:
```
git clean -nfxd
```
(Drop the `n` when you are certain.)

Anyway, I'm fine with your ignores here. @rpavlik, any specific objections?

rpavlik:
No, of course not. Just saw the notification and thought I'd chime in with help

cdunn2001:
Pulled.

PR #206, opened 2015-03-07
Trivial fixes in CZString constructors.

datadiode:
Value::CZString::CZString(char const* str, unsigned length, DuplicationPolicy allocate) with allocate == duplicate does not seem to happen.
It may be favorable to just not support that case, but rather add an assert(allocate != duplicate).

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -196,7 +196,7 @@ void Value::CommentInfo::setComment(const char* text, size_t len) {
 Value::CZString::CZString(ArrayIndex index) : cstr_(0), index_(index) {}
 
 Value::CZString::CZString(char const* str, unsigned length, DuplicationPolicy allocate)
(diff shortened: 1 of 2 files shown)


cdunn2001:
Yes, the first ctor with allocate==duplicate never occurs. I was meaning to drop that. The 2nd one is an embarrassing bug. Thanks.

However, your first test passes without your fix. Looking into the first test, and why zeoresInKeys did not catch this bug....

cdunn2001:
Interesting. Your first test,
```cpp
JSONTEST_ASSERT_EQUAL(Json::Value("object1_[keyWithNulls]"), object1_[strKeyWithNulls]);
```
actually uses the non-const `Value::operator[]`. That test fails only for the const version.

`ValueTest/zeroesInKeys` passed because it did not actually test *embedded* zeroes, though it tested the different behavior for c-strings.

cdunn2001:
Rebased as #207.

Issue #214, opened 2015-03-09
isfinite compilation error


connormanning:
```
Building CXX object myProject/third/CMakeFiles/third.dir/jsoncpp.cpp.o
/home/vagrant/myProject/myProject/third/jsoncpp.cpp: In function ‘std::string Json::valueToString(double)’:
/home/vagrant/myProject/myProject/third/jsoncpp.cpp:4049:21: error: ‘isfinite’ was not declared in this scope
   if (isfinite(value)) {
                     ^
/home/vagrant/myProject/myProject/third/jsoncpp.cpp:4049:21: note: suggested alternative:
In file included from /usr/include/c++/4.8/random:38:0,
                 from /usr/include/c++/4.8/bits/stl_algo.h:65,
                 from /usr/include/c++/4.8/algorithm:62,
                 from /home/vagrant/myProject/myProject/third/jsoncpp.cpp:2398:
/usr/include/c++/4.8/cmath:596:5: note:   ‘std::isfinite’
     isfinite(_Tp __x)
     ^
make[2]: *** [myProject/third/CMakeFiles/third.dir/jsoncpp.cpp.o] Error 1
make[1]: *** [myProject/third/CMakeFiles/third.dir/all] Error 2
make: *** [all] Error 2
```

This is on Ubuntu Trusty, failing [here](https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_writer.cpp#L113) (I am using an amalgamated version of SHA ed495edcc1bda878ef3cabd4c1ddaa1cdf2c57b0).  For me, replacing `isfinite()` with `std::isfinite()` at that line let me compile.  However due to the cross-platform support for `isfinite` [here](https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_writer.cpp#L20-L34) this isn't the real solution.

Dani-Hub:
What is the gcc version you are using? Is it possible that you are a victim of this error:

https://bugs.launchpad.net/ubuntu/+source/gcc-4.3/+bug/512741

connormanning:
Interesting find, wasn't aware of that.  My GCC version is 4.8.2.

```
$ gcc -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/4.8/lto-wrapper
Target: x86_64-linux-gnu
[...]

PR #215, opened 2015-03-09
math.h -> cmath.  Add macro case for isfinite to use std namespace.

connormanning:
Possible solution for #214.  Requires replacing the `math.h` include with `cmath`, so if there was a reason for using that, this might not be an acceptable fix.


connormanning:
Closing in favor of more complete solutions (see discussion at #214).

PR #221, opened 2015-03-14
Added forgotten virtual dtor for Json::CharReader::Factory

btolfa:


cdunn2001:
Ugh! You're right. That's a terrible mistake.

Unfortunately, this change is not binary-compatible. Hmmm. I guess we're better off fixing it quickly.

PR #225, opened 2015-03-19
fix find_program() bug: no result in not-win sys

selaselah:
if str.split() without args, all empty string will drop


cdunn2001:
Yes, that slipped in without testing. Thank you.

PR #227, opened 2015-03-24
Revert "Use std namespace for snprintf."

baruchsiach:
This reverts commit 1c58876185d2a4ed87dac4a54b82f607e74f55fd.

snprintf() is not in the std namespace. This fixes the build error below:

src/lib_json/json_writer.cpp:33:18: error: 'snprintf' is not a member of 'std'


baruchsiach:
As noted by Thomas Petazzoni (http://article.gmane.org/gmane.comp.lib.uclibc.buildroot/110700) the commit message is wrong. std::snprintf is part of the C++11. I'll update and send another pull request.

PR #233, opened 2015-04-01
add solution for msvc2012

fenglc:


lanzkron:
This seems to be misnamed, it looks like this is for Visual Studio 2013 (VS12) not Visual Studio 2012 (VS11).

lanzkron:
@cdunn2001 this compiles OK with VS12 (all unit tests pass).

 I think the PR is fine. In any case it doesn't affect any existing code and can only impact people using VS12 (which didn't have project files before).

fenglc:
Sorry for my typo error.
msvc2012 is a typo of msvc2013.

cdunn2001:
I am concerned about churn. Are these files truly stable? Or will VisualStudio modify them?

Also, I think we want these in `.gitignore`:
```
*.vcproj.*.user
*.ncb
*.aps
*.suo
```

lanzkron:
@cdunn2001  the files here are not modified by VS automatically. VS puts all its changes in `.user` and  `.suo` files. You can see that these files are committed for VS10. 

Here's a list of which files should and should not be committed to source control: http://stackoverflow.com/a/3923352/3848

cdunn2001:
@lanzkron: Ok. But adding those others to `.gitignore` is not a bad idea.

@fenglc: This PR still needs to be rebased, and I don't have time today. If you don't know how, I'll do it this week sometime.

lanzkron:
Keep in mind that vcproj files are no more, current versions of VS use vcxproj files.

PR #235, opened 2015-04-06
fixed compiler warnings

marklakata:
fixed compiler warnings on my end, started adding new documentation (didn't get very far)


cdunn2001:
Don't worry about doc changes. We'll deal with that later.

Rebase to the tip of `master` and push again. That should spawn the Travis CI tests. Call me if you need help with that.

Even then, I will have to apply a revert of all this on `0.y.z` branch, but I'll take care of all that.

But I have a question: What about `<cstdint>`? In fact, maybe instead of using the special new `JSON_USE_STDINT`, why not just check for C++11? Since if we're using C++11, then we *definitely* have `<cstdint>`.

marklakata:
It's fine if you ignore this pull request. They are not that important.

PR #217, opened 2015-03-09
Fix infinite compilation problems on some gcc variants (Issue #214)

Dani-Hub:
The provided patch has been tested on gcc mingw 5.0.0 and for Visual Studio 2012.

PR #247, opened 2015-04-18
Nan

cdunn2001:
Rebased #217 onto appveyor branch. Let's see if appveyor passes this way.


cdunn2001:
Replaced by #339, I think.

PR #263, opened 2015-04-23
Static/shared

cdunn2001:
(created from several separate branches by @ya1gaurav)

Let's see if it builds....


cdunn2001:
Let's discuss this now. What was broken? What's the goal of this change? Why did you add `devtools/agent_*.json`? Why didn't you update `appveyor.yml`?

ya1gaurav:
The goal of these changes is to use standard CMake variable as option for building either shared/static library. Currently it is using JSONCPP_LIB_BUILD_SHARED & JSONCPP_LIB_BUILD_STATIC.
It will avoid extra variables, as there exists standard variable for same purpose.
Also, as I checked these variables are not used for some other purpose.

cdunn2001:
According to a comment elsewhere, using the non-standard variables means that this library does not work well as a subproject. Is that your point?

I tried to learn why the custom variables were chosen. It seems to have been arbitrary, in the original (eafd702a17163160a911368f0a3475b49f673738) commit for **cmake** build-config.

I think AppVeyor failed a build only because of a problem contacting GitHub. That's fine.

ya1gaurav:
Actually my motivation is one of the patch i saw - https://github.com/open-source-parsers/jsoncpp/issues/51
As it is included as work around for standard variable, then why using the other variables like JSONCPP_LIB_BUILD_SHARED, it seems duplicate. Isn't it?

ya1gaurav:
I have one more question regarding flag "JSON_DLL_BUILD".
DLL are for Windows, is this flag generic for shared library or only for windows in Jsoncpp?
If it is for windows only, should be written under "if(WIN32)".

cdunn2001:
> should be written under "if(WIN32)"

Yes. But it's mingled with the CPPTL code, which I cannot test. I am hoping not to break that since @blep (the original author) still uses it. It's a minor issue.

Thanks for fixing these variables names.

ya1gaurav:
In Readme file, i can see below:
If JsonCpp was built as a dynamic library on Windows, then your project needs to define the macro JSON_DLL.
It think it means for windows only.

Issue #223, opened 2015-03-16
Json::Value should support move semantics


georgehdd:
I see that Json::Value does not support move semantics (move constructor and move assignment operator). 

Is there a reason for this? If not would you like any help on this front?

BillyGoto:
the last time we tried this there was an issue of binary compatibility.
Cdunn, is that still a concern?

On Monday, March 16, 2015, georgehdd <notifications@github.com> wrote:

> I see that Json::Value does not support move semantics (move constructor
> and move assignment operator).
>
> Is there a reason for this? If not would you like any help on this front?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/open-source-parsers/jsoncpp/issues/223>.
>


-- 
ǝnɥɐuop ʎllıq

cdunn2001:
Yes. But since we've just introduced a rather serious backward-incompatibility vs. 1.4 (the virtual dtor in the Reader Builder) we could add it immediately on the 1.*.* branch. The 0.*.* branch would exclude it and would remain compatible with 0.6.0-rc2, which is the most important thing.

The real issue is that if 1.*.* goes too far down the C++-11 path, then 0.*.* will not be patchable with bug-fixes common to both. However, I think we're past the worst of the bugs at this point, so I'd be fine with starting 2.*.* right away and letting it diverge. Or even with letting 1.*.* diverge, if that's not too confusing.

I'll be AFK for a couple weeks, so do whatever you think best.

lanzkron:
Would you like me to create a pull request for this? If so please let me know what compilers are supported.

cdunn2001:
Sure.

Officially, we support the versions of gcc an clang used by default in Travis CI, which would be run automatically on your pull request. Unofficially, when users submit patches for others (esp. versions of VisualStudio) we accept them. In general, we support older compilers in the 0.*.* branch than in the 1.*.* branch, so just make your changes on 1.*.* (which is `master`).

lanzkron:
I created a pull request for this but it failed to merge :(
Please take a look 

https://github.com/open-source-parsers/jsoncpp/pull/264

tshepang:
Should this remain open?

cdunn2001:
@tshepang, I guess not. Thanks for noticing.

matsbror:
So there is no plan on adding the move semantics to Value?

cdunn2001:
[...]

PR #264, opened 2015-04-26
Add Move constructors

lanzkron:
Add move constructors to Json::Value and Json::Value::CZString 

As requested by issue #223: Json::Value should support move semantics

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -67,40 +67,21 @@
 // Storages, and 64 bits integer support is disabled.
 // #define JSON_NO_INT64 1
 
-#if defined(_MSC_VER) // MSVC
-	#if _MSC_VER <= 1200 // MSVC 6
-		// Microsoft Visual Studio 6 only support conversion from __int64 to double
-		// (no conversion from unsigned __int64).
-		#define JSON_USE_INT64_DOUBLE_CONVERSION 1
-		// Disable warning 4786 for VS6 caused by STL (identifier was truncated to '255'
-		// characters in the debug information)
-		// All projects I've ever seen with VS6 were using this globally (not bothering
-		// with pragma push/pop).
-		#pragma warning(disable : 4786)
-	#endif // MSVC 6
+#if defined(_MSC_VER) && _MSC_VER <= 1200 // MSVC 6
+// Microsoft Visual Studio 6 only support conversion from __int64 to double
+// (no conversion from unsigned __int64).
+#define JSON_USE_INT64_DOUBLE_CONVERSION 1
+// Disable warning 4786 for VS6 caused by STL (identifier was truncated to '255'
+// characters in the debug information)
+// All projects I've ever seen with VS6 were using this globally (not bothering
+// with pragma push/pop).
+#pragma warning(disable : 4786)
+#endif // if defined(_MSC_VER)  &&  _MSC_VER < 1200 // MSVC 6
 
-	#if _MSC_VER >= 1500 // MSVC 2008
-		/// Indicates that the following function is deprecated.
-		#define JSONCPP_DEPRECATED(message) __declspec(deprecated(message))
-	#endif
-
-	#if _MSC_VER >= 1600 // MSVC 2010
-		#define JSON_HAS_RVALUE_REFERENCES
-	#endif 
-#endif // defined(_MSC_VER)
-
-
-#ifdef __clang__
-	#if __has_feature(cxx_rvalue_references)
-		#define JSON_HAS_RVALUE_REFERENCES
-	#endif 
-#elif defined __GNUC__
-	#if defined(__GXX_EXPERIMENTAL_CXX0X__) 
-		#define JSON_HAS_RVALUE_REFERENCES
-	#endif
-#endif // __clang__ || __GNUC__
-
-
+#if defined(_MSC_VER) && _MSC_VER >= 1500 // MSVC 2008
+/// Indicates that the following function is deprecated.
+#define JSONCPP_DEPRECATED(message) __declspec(deprecated(message))
+#endif
 
 #if !defined(JSONCPP_DEPRECATED)
 #define JSONCPP_DEPRECATED(message)
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -83,6 +83,28 @@
 #define JSONCPP_DEPRECATED(message) __declspec(deprecated(message))
 #endif
 
+#ifndef JSON_HAS_RVALUE_REFERENCES
+
+#if defined(_MSC_VER) && _MSC_VER >= 1600 // MSVC >= 2010
+#define JSON_HAS_RVALUE_REFERENCES 1
+#endif // MSVC >= 2010
+
+#ifdef __clang__
+#if __has_feature(cxx_rvalue_references)
+#define JSON_HAS_RVALUE_REFERENCES 1
+#endif 
+#elif defined __GNUC__ // not clang (gcc comes later since clang emulates gcc)
+#if defined(__GXX_EXPERIMENTAL_CXX0X__) || (__cplusplus >= 201103L)
+#define JSON_HAS_RVALUE_REFERENCES 1
+#endif
+#endif // __clang__ || __GNUC__
+
+#endif // not defined JSON_HAS_RVALUE_REFERENCES
+
+#ifndef JSON_HAS_RVALUE_REFERENCES
+#define JSON_HAS_RVALUE_REFERENCES 0
+#endif 
+
 #if !defined(JSONCPP_DEPRECATED)
 #define JSONCPP_DEPRECATED(message)
 #endif // if !defined(JSONCPP_DEPRECATED)
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -67,21 +67,40 @@
 // Storages, and 64 bits integer support is disabled.
 // #define JSON_NO_INT64 1
 
-#if defined(_MSC_VER) && _MSC_VER <= 1200 // MSVC 6
-// Microsoft Visual Studio 6 only support conversion from __int64 to double
-// (no conversion from unsigned __int64).
-#define JSON_USE_INT64_DOUBLE_CONVERSION 1
-// Disable warning 4786 for VS6 caused by STL (identifier was truncated to '255'
-// characters in the debug information)
-// All projects I've ever seen with VS6 were using this globally (not bothering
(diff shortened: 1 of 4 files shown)


include/json/config.h
```diff
@@ -67,21 +67,40 @@
...
-// with pragma push/pop).
-#pragma warning(disable : 4786)
-#endif // if defined(_MSC_VER)  &&  _MSC_VER < 1200 // MSVC 6
-
-#if defined(_MSC_VER) && _MSC_VER >= 1500 // MSVC 2008
-/// Indicates that the following function is deprecated.
-#define JSONCPP_DEPRECATED(message) __declspec(deprecated(message))
-#endif
+#if defined(_MSC_VER) // MSVC
+	#if _MSC_VER <= 1200 // MSVC 6
```
BillyDonahue: Please don't indent preprocessor directives. It's churning the diff.

include/json/config.h
```diff
@@ -67,21 +67,40 @@
...
+	#endif 
+#endif // defined(_MSC_VER)
+
+
+#ifdef __clang__
+	#if __has_feature(cxx_rvalue_references)
+		#define JSON_HAS_RVALUE_REFERENCES
+	#endif 
+#elif defined __GNUC__
+	#if defined(__GXX_EXPERIMENTAL_CXX0X__)
```
BillyDonahue: This is not a reliable macro. It's going to go away soon.
What I think you want is:

    #if __cplusplus >= 201103L

Which should probably just work on clang, GCC, or MSVC.

include/json/config.h
lanzkron: Don't you think it makes the conditions much more readable? 
I have no problem to adhere to the coding standards of the project if you think that's better

include/json/value.h
```diff
@@ -171,6 +171,9 @@ class JSON_API Value {
     CZString(ArrayIndex index);
     CZString(const char* cstr, DuplicationPolicy allocate);
     CZString(const CZString& other);
+#ifdef JSON_HAS_RVALUE_REFERENCES
+	CZString(CZString&& other);
```
BillyDonahue: A `#if` shouldn't affect indentation of C++ declarations. This should be aligned with the other constructors.
Should be marked `noexcept`.

include/json/value.h
```diff
@@ -171,6 +171,9 @@ class JSON_API Value {
...
+#endif
     ~CZString();
     CZString& operator=(CZString other);
```
BillyDonahue: If you're providing a move ctor, you should also provide a move assignment.

    CZString& operator=(CZString&& other) noexcept;

include/json/config.h
lanzkron: From what I saw __cplusplus isn't dependable. This is how boost are doing it. 
For visual studio 2012 the value of __cplusplus is still 199711.
See http://stackoverflow.com/a/21923116/3848

src/lib_json/json_value.cpp
```diff
@@ -182,6 +182,15 @@ Value::CZString::CZString(const CZString& other)
                      ? noDuplication : duplicate)
                  : other.index_) {}
 
+#ifdef JSON_HAS_RVALUE_REFERENCES
+Value::CZString::CZString(CZString&& other)
+	: cstr_(other.cstr_)
+	, index_(other.index_)
```
BillyDonahue: Move this comma up to the preceding line.

include/json/value.h
lanzkron: Re: indentations: Sure my bad
Re: noexcept: Unfortunately not all compilers that support rvalue references support noexcept

include/json/value.h
lanzkron: I was going to but since CZString provides a copy and swap assignment operator I don't think it's worthwhile, if `other`  refers to an rvalue it will be moved in to the by-value parameter and then swap would be just as efficient.

src/lib_json/json_value.cpp
lanzkron: OK

include/json/value.h
BillyDonahue: Without `noexcept`, the move ctor often degenerates into a copy, like when reallocating a vector.

    std::vector<Json::Value> v = MakeSomeValues();
    v.reserve(v.capacity() + 1);  // realloc: all elements are deep-copied!

If a compiler supports move semantics but not `noexcept`, it needs to be special-cased. I suggest detecting and flagging noexcept support in config.h similarly to how rvalue-reference support is flagged, otherwise we're just making slow move ops where speed matters most.

There's a reasonable-looking `#if` given here:
http://stackoverflow.com/questions/18387640/how-to-deal-with-noexcept-in-visual-studio

include/json/config.h
BillyDonahue: That is not exactly how boost is doing it.
Boost.config is doing this for GCC:

```c++
#if defined(__GXX_EXPERIMENTAL_CXX0X__) || (__cplusplus >= 201103L)
#  define BOOST_GCC_CXX11
#endif
```

http://www.boost.org/doc/libs/1_58_0/boost/config/compiler/gcc.hpp

That's just gcc.hpp, but each compiler is different.

This is the built-in compiler support. Boost can also be reconfigured for unforeseen compilers in which case it tries a little test program to see if rvalue references are supported.

The analogous macros for VisualC++ are here:
http://www.boost.org/doc/libs/1_58_0/boost/config/compiler/visualc.hpp

include/json/config.h
BillyDonahue: I don't think the indents as written improve the clarity of the file overall.

The usual thing is to keep the `#` in column 1. This is historical convention.
The next token is indented by 0, 1, or 2 spaces. In this project, it's 0, but you've chosen 8.

The idea is to keep the preprocessor lines visually segregated from the C++ code.
A reader can eyeball column 1 and see a run of `#`, which is easy.

include/json/value.h
BillyDonahue: Move is the more fundamental operation. I think swap should really be built in terms of it rather than the other way round.

```c++
class Value {
    ...
    Value(const Value&);
    Value(Value&&) noexcept;
    Value& operator=(const Value&);
    Value& operator=(Value&&) noexcept;
    void swap(Value& v) noexcept { std::swap(*this, v); }
    friend void swap(Value& a, Value& b) noexcept { a.swap(b); }
};
```

The copy-and-swap trick is a pre-11 hack. It does extra work to fix up the moved-from source object only to delete it at the end of function. There are considerations of vector- or string capacity as well. A moved-from object can be better equipped for reuse if it is moved directly into its destination rather than moving into a temporary first.

I prefer separating the const T& and the T&& cases into separate functions in case they can be accomplished more efficiently with different strategies. Because of binary compatibility, we can't split them up later if we decide we want to. So really we need two functions.

src/lib_json/json_value.cpp
cdunn2001: @lanzkron, personally I like your `,` under `:`, since it makes adding arguments easier, but we try to stick with google-style-guide to avoid arguments. (Actually, it's best if you reformat with Clang, since @jacobsa, did a lot of work on that.) In general, I trust @BillyDonahue's judgement, so I thank you for taking all his suggestions.

I want to pull this and fix the the merge issues myself, and maybe reformat with Clang. However, I don't like adding pre-processor macros. I wonder if we wouldn't be better off with a `2.0.0-alpha` branch. We can guarantee *source-code* compatibility but completely relax *binary* compatibility. We can require compilers which support all features. We could then add all sorts of things before it is stable. Anybody can use `1.y.z` or even `0.y.z` if they have a problem.

Downside: Chromium would not have move-constructors. But I don't think Chromium cares about binary-compatibility, so maybe they could simply upgrade.

Opinions?

include/json/value.h
lanzkron: I'm not clear on how adding an overload effects binary compatibility. Since Json.cpp is a static library a version that has this code will call the move operator while one which doesn't will use the old copy version. Since having a (non-virtual) overload doesn't effect the size or layout of instance objects how is this a binary compatibility concern?

src/lib_json/json_value.cpp
lanzkron: This is my first contribution to github so I apologize if I've performed any faux pas :o)

As I replied to @BillyDonahue, I don't understand the binary compatibility concern, could you please enlighten me?

Of course I respect the fact that I'm contributing to an existing project and will respect the coding standards here.

You say that you want to pull and fix the merge, does this mean I should not make the suggested changes and submit a new pull request?

include/json/config.h
lanzkron: Perhaps we can `#define JSON_HAS_RVALUE_REFERENCES` to 1 or 0 but only if it's not previously defined. Then we give the end user the option to explicitly `#define` the correct value so that it can be used with all compilers. 
Obviously this would mean using `#if` instead of `#ifdef` when checking conditions.

cdunn2001:
I rebased onto `master` (which is better than merging) your code as #267. Move discussion to there, or simply pull that branch and `git push -f` into your own fork to update *this* PR.

Don't worry about compatibility. We'll deal with that.

lanzkron:
Rebased by @cdunn2001 in #267, closing this pull request

cdunn2001:
I have a question: Why do we need to add the move ctor if the default version is [provided automatically](http://stackoverflow.com/questions/4724125/when-will-adding-a-move-constructor-and-a-move-assignment-operator-really-start)? ... Oh, I guess the rules for this [have evolved](http://stackoverflow.com/questions/4819936/why-no-default-move-assignment-move-constructor).

lanzkron:
Yeah, if you have a destructor or a copy constructor then the move
constructor will not be generated.

Even if this is not the case from a practical point of view I don't think
all compilers currently generate a default move constructor.

I think we could have done something like:

Value(Value&&) = default;


But not all current compilers support this either :(

cdunn2001:
Oh, I sometimes update the comment for clarity. I guess that doesn't send a new e-mail.

Thanks, Motti.

BillyDonahue:
The implicit move ops don't work for resource-managing classes like `Json::Value::CZString`.
It would just move (copy) the data members one by one. `cstr_` and `index_` are just pointer and int, so they'd be copied into the destination, but the source would not be changed and would still think of itself as the owner of its `cstr_` data.

lanzkron:
@BillyDonahue  good point!

PR #250, opened 2015-04-19
in travis, build for osx also

cdunn2001:
test via pull-request

PR #265, opened 2015-04-27
run valgrind in Travis CI

### LINKED ISSUE #222: Integrate valgrind into Travis build
From @cinemast in #157:
> I integrated the valgrind checking to my travis build. Maybe that is a solution for you:
>
> https://github.com/cinemast/libjson-rpc-cpp/blob/master/.travis.yml

cdunn2001:
Because this runs apt-get, it will not work as-is for OSX. So when
we have OSX in Travis, we will have to wrap this somehow. See #250.

Closes #222.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,20 @@
 # http://about.travis-ci.org/docs/user/build-configuration/
 # This file can be validated on:
 # http://lint.travis-ci.org/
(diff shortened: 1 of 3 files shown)


cdunn2001:
10s longer build times in Travis (because of `apt-get update`/`apt-get install valgrind`)  but I think it's worth it.

PR #267, opened 2015-04-27
move ctors

cdunn2001:
* Add move constructor to Value::CZString
* Add unit test for Value move constructor
* Allow includer to specify in advance the value for
JSON_HAS_RVALUE_REFERENCES

rebased #264.

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -56,29 +56,58 @@
 // Storages, and 64 bits integer support is disabled.
 // #define JSON_NO_INT64 1
 
-#if defined(_MSC_VER) && _MSC_VER <= 1200 // MSVC 6
+#if defined(_MSC_VER) // MSVC
+#  if _MSC_VER <= 1200 // MSVC 6
     // Microsoft Visual Studio 6 only support conversion from __int64 to double
     // (no conversion from unsigned __int64).
-    #define JSON_USE_INT64_DOUBLE_CONVERSION 1
+#    define JSON_USE_INT64_DOUBLE_CONVERSION 1
     // Disable warning 4786 for VS6 caused by STL (identifier was truncated to '255'
     // characters in the debug information)
     // All projects I've ever seen with VS6 were using this globally (not bothering
     // with pragma push/pop).
 #    pragma warning(disable : 4786)
-#endif // if defined(_MSC_VER)  &&  _MSC_VER < 1200 // MSVC 6
+#  endif // MSVC 6
 
-#if defined(_MSC_VER) && _MSC_VER >= 1500 // MSVC 2008
+#  if _MSC_VER >= 1500 // MSVC 2008
     /// Indicates that the following function is deprecated.
-#define JSONCPP_DEPRECATED(message) __declspec(deprecated(message))
-#elif defined(__clang__) && defined(__has_feature)
-#if __has_feature(attribute_deprecated_with_message)
-#define JSONCPP_DEPRECATED(message)  __attribute__ ((deprecated(message)))
+#    define JSONCPP_DEPRECATED(message) __declspec(deprecated(message))
+#  endif
+
+#endif // defined(_MSC_VER)
+
+
+#ifndef JSON_HAS_RVALUE_REFERENCES
+
+#if defined(_MSC_VER) && _MSC_VER >= 1600 // MSVC >= 2010
+#define JSON_HAS_RVALUE_REFERENCES 1
+#endif // MSVC >= 2010
+
+#ifdef __clang__
+#if __has_feature(cxx_rvalue_references)
+#define JSON_HAS_RVALUE_REFERENCES 1
+#endif  // has_feature
+
+#elif defined __GNUC__ // not clang (gcc comes later since clang emulates gcc)
+#if defined(__GXX_EXPERIMENTAL_CXX0X__) || (__cplusplus >= 201103L)
+#define JSON_HAS_RVALUE_REFERENCES 1
+#endif  // GXX_EXPERIMENTAL
+
+#endif // __clang__ || __GNUC__
+
+#endif // not defined JSON_HAS_RVALUE_REFERENCES
+
+#ifndef JSON_HAS_RVALUE_REFERENCES
+#define JSON_HAS_RVALUE_REFERENCES 0
 #endif
-#elif defined(__GNUC__) &&  (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5))
-#define JSONCPP_DEPRECATED(message)  __attribute__ ((deprecated(message)))
-#elif defined(__GNUC__) &&  (__GNUC__ > 3 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))
-#define JSONCPP_DEPRECATED(message)  __attribute__((__deprecated__))
-#endif
+
+#ifdef __clang__
+#elif defined __GNUC__ // not clang (gcc comes later since clang emulates gcc)
+#  if (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5))
+#    define JSONCPP_DEPRECATED(message)  __attribute__ ((deprecated(message)))
+#  elif (__GNUC__ > 3 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))
+#    define JSONCPP_DEPRECATED(message)  __attribute__((__deprecated__))
+#  endif  // GNUC version
+#endif // __clang__ || __GNUC__
 
 #if !defined(JSONCPP_DEPRECATED)
 #define JSONCPP_DEPRECATED(message)
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -237,6 +237,15 @@ Value::CZString::CZString(const CZString& other)
   storage_.length_ = other.storage_.length_;
 }
 
+#if JSON_HAS_RVALUE_REFERENCES
+Value::CZString::CZString(CZString&& other)
+  : cstr_(other.cstr_),
+  index_(other.index_)
+{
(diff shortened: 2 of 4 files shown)


cdunn2001:
We still have to consider what to do for the older branch. We maintain strict *binary-compatibility* with ancient `0.6.0-rc2` on that branch. Adding a method is fine. I think even adding an rvalue method is fine. So maybe there are no issues.

lanzkron:
Thanks @cdunn2001 !

I see you also fixed my use of `nullptr`, I only realized it after my PR.

include/json/config.h
```diff
@@ -56,28 +56,46 @@
...
+#ifdef __clang__
+#if __has_feature(cxx_rvalue_references)
+#define JSON_HAS_RVALUE_REFERENCES 1
+#endif
+#elif defined __GNUC__ // not clang (gcc comes later since clang emulates gcc)
+#if defined(__GXX_EXPERIMENTAL_CXX0X__) || (__cplusplus >= 201103L)
+#define JSON_HAS_RVALUE_REFERENCES 1
 #endif
-#elif defined(__GNUC__) &&  (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5))
-#define JSONCPP_DEPRECATED(message)  __attribute__ ((deprecated(message)))
```
cdunn2001: Oh, I meant to ask: Why was JSON_DEPRECATED dropped for GNUC?

lanzkron: Did I do that? If so it was in error, sorry.

cdunn2001:
Asked about backward-compatibility on [stackoverflow](http://stackoverflow.com/questions/29913917/does-adding-a-move-constructor-break-binary-compatibility).

lanzkron:
Answered on stackoverflow <http://stackoverflow.com/a/29914866/3848>
(mostly to get feedback from the community as to whether my understanding
is correct).

akrieger:
Can we add move assignment operators as well, so that we can efficiently do

```
Json::Value root;
root["key"] = functionThatGeneratesJson();
```

cdunn2001:
I agree, as in @BillyDonahue comments at #264:
```cpp
class Value {
    ...
    Value(const Value&);
    Value(Value&&) noexcept;
    Value& operator=(const Value&);
    Value& operator=(Value&&) noexcept;
    void swap(Value& v) noexcept { std::swap(*this, v); }
    friend void swap(Value& a, Value& b) noexcept { a.swap(b); }
};
```
We also need a macro for `noexcept`, for the compilers which do not support. Personally, I'm fine with
```cpp
#if dumb_compiler
#define noexcept
#endif
```

akrieger:
I like the very political references to 'the compilers which do not support'.

I actually dug into it and it seems the existing operator= implementations exhibit move semantics, as of https://github.com/open-source-parsers/jsoncpp/commit/45cd9490cd261da31cef84a44d2c587be7e26e99. Looks like that can be revered to the old copy semantics with the addition of proper move operators. Regardless of the philosophical motivation behind making copy assignment actually be move assignment.

BillyDonahue:
I think the direct `#define` of a keyword is asking for trouble, especially in a header. Let's be conservative and use JSON_NOEXCEPT. On such platforms, I'd imagine there are plenty of libs that might want to #define noexcept and we'd be conflicting if we did the same.

include/json/config.h
```diff
@@ -56,30 +56,59 @@
...
-#define JSONCPP_DEPRECATED(message) __declspec(deprecated(message))
-#elif defined(__clang__) && defined(__has_feature)
-#if __has_feature(attribute_deprecated_with_message)
-#define JSONCPP_DEPRECATED(message)  __attribute__ ((deprecated(message)))
-#endif
-#elif defined(__GNUC__) &&  (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5))
-#define JSONCPP_DEPRECATED(message)  __attribute__ ((deprecated(message)))
-#elif defined(__GNUC__) &&  (__GNUC__ > 3 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))
-#define JSONCPP_DEPRECATED(message)  __attribute__((__deprecated__))
+#if defined(_MSC_VER) // MSVC
```
BillyDonahue: Can the reindent be its own PR? It's hard to separate cosmetics from substance as-is.
It would be nice to emphasize the substantive changes with concise PRs.

src/lib_json/json_value.cpp
```diff
@@ -254,6 +254,15 @@ Value::CZString::CZString(const CZString& other)
   storage_.length_ = other.storage_.length_;
 }
 
+#if JSON_HAS_RVALUE_REFERENCES
+Value::CZString::CZString(CZString&& other)
+  : cstr_(other.cstr_),
```
BillyDonahue: Google style would be to indent this more like:
```c++
Value::CZString::CZString(CZString&& src)
    : cstr_(src.cstr_),
      index_(src.index_) {
  src.cstr_ = nullptr;
}
```

(And join cstr_ and index_ onto a single line actually)

src/lib_json/json_value.cpp
```diff
@@ -442,6 +451,15 @@ Value::Value(Value const& other)
   }
 }
 
+#if JSON_HAS_RVALUE_REFERENCES
+// Move constructor
+Value::Value(Value&& other)
```
BillyDonahue: Braced and indented inconsistently with surrounding code. Should be closer to:

```c++
Value::Value(Value&& src) {
  initBasic(nullValue);
  swap(other);
}
```

cdunn2001:
You're probably right; not good in a header. `JSON_NOEXCEPT` then.

> I like the very political references to 'the compilers which do not support'.

Yes! I have strong opinions.

cdunn2001:
@BillyDonahue, would you pull my code from [this branch](https://github.com/cdunn2001/jsoncpp/tree/moveSemantics) and add whatever other operators we need? I will take of the `JSON_NOEXCEPT` macro.

lanzkron:
@akrieger regarding move assignment operator:
Currently the assignment operator uses **copy and swap**, with the addition of a move constructor it would be able to use **move and swap** for rvalues which is almost as efficient as writing a special move assignment operator.

If you have a look at `Json::Value`'s copy constructor you'll see that it's rather complex and in my opinion duplicating this logic in a move constructor and assignment operator isn't worth the potential improvements.

cdunn2001:
@lanzkron, Do we need  any extra methods, beyond what are already in this pull-request? Do we need the move-ctor for CZString at all? I'm not sure where we are.

BillyDonahue:
@cdunn : I'm still a github noob, but I'm figuring out what you need.
I'm a little confused about what that branch represents.

I'd like to ask about compatibility again. We are not strictly adding new
functions here. My ideal suggestion would be that we'd be REMOVING the
by-value "operator=(Value)" assignment and ADDING the by-cref and by-rref
"operator=(const Value&)" and "operator=(Value&&)" assignments.

Old code expecting to find "operator=(Value)" in libjsoncpp.so wouldn't
find it anymore. I'm not saying this is a problem. I haven't been following
the branching plans or compatibility requirements closely enough. I just
want to make sure this point about the disappearing by-value overload is
understood.


On Wed, Apr 29, 2015 at 12:16 AM, Christopher Dunn <notifications@github.com

lanzkron:
@cdunn2001 
The move-ctor for `CZString` is good to have (it also helps when assigning from an rvalue due to the copy and swap assignment operator in `CZString`).

I'm also new to GitHub, who makes the final call? In my opinion having a move-ctor and keeping the copy/swap assignment operator is good enough. In which case I think we're good to go.

Keep in mind that my implementation of `Value(Value&&)` depends on `swap`. If we do decide to implement move ctor/op= from scratch then we'll have to duplicate the logic of the copy constructor (and keep them synced in future changes).

I don't currently have the capacity to do this. If that's what we want then we can dump my changes and wait for someone else to code support for move semantics.

cdunn2001:
@BillyDonahue, we could remove `CZString::operator=()`, because the likelihood that anyone uses it directly is miniscule. Technically, we shouldn't change `Value::operator=()`, but I don't think anyone currently relies on binary-compatibility for 1.0, so it's ok. That method in particular is a welcome change because that is already different in the 0.y.z branch. See 1c4f6a2d79d466bb501cb84626f269e40e60ba7c:
```
-  Value& operator=(Value other);
+  Value &operator=(const Value &other);
```

We can remove anything in a future 2.0.0 release. At some point, we should move development to a new alpha branch. That's a discussion for another day.

@lanzkron, there is no final call. We can always add new features. It's better to add less, because it's hard to *remove* features.

akrieger:
@lanzkron ah, yes, you're right, it is a copy and swap.

lanzkron:
@cdunn2001 I assume you mean replacing CZString's operator= with lvalue and rvalue versions, not remove it completely

thelema:
Is there any chance of this being merged?  It looks a little dead.  I'm hoping to have move constructors to greatly improve performance of building a very large json object in pieces, and I found this open pull request.

cdunn2001:
I'm concerned about removing symbols. I don't mind removing them from CZString because nobody who cares about binary compatibility would use it. But I don't think we can remove symbols from Value. I think 1.0 is now used in Debian, e.g.

I've just rebased this. Let's take another look now.

include/json/config.h
cdunn2001: Ok. I'll try re-indent first, and rebase this onto that...

cdunn2001:
Ok, I'd be fine with merging this, if there are no objections. I'll wait a couple days...

BillyDonahue:
Looks good. I do wish we could have the move assignment operations.
Is there an advanced branch for that sort of thing free from the binary compatibility concerns?

cdunn2001:
Soon. The old branches are stable enough that we can soon consider `master` as a new pre-release for 2.0.

Issue #184, opened 2015-02-24
Please use std::aligned_storage since C++11 for static JSON null 


tymonx:
This should be replaced with std::aligned_storage to be more portable between compilers and avoid 'clumsy' workaround:
https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_value.cpp#L36

In my own JSON implementation I use it like this:
https://github.com/tymonx/OpenRedfish/blob/dev/json/json.cpp#L52

cdunn2001:
This code was actually copied from a change in Android, where somebody noticed a crash for ARM. I'm reluctant to make a change that we cannot test. We run our Travis CI tests with gcc and clang. What do you use?

If you submit a patch and can verify that it works for some particular arch/compiler, we'd probably accept it on `master`. I'm not sure about the `0.8.z` branch. How would an alignment change impact binary-compatibility?

cdunn2001:
We now have `-std=c++11` on *master*. See #313. So this is worth trying, if we are willing to allow `0.10.z` and `1.6.z` to diverge.

tymonx:
Thanks for information, I will follow. Up to today I have my own JSON C++11 implementation. Maybe you can borrow some tips from there ;) 

Repo: https://github.com/tymonx/json-cxx

Using aligned storage: https://github.com/tymonx/json-cxx/blob/master/src/value.cpp#L56

dota17:
Closing due to inactivity. If anyone has new idea, feel free to reopen it,

It seems that `aligned_*` are harmful to codebases and should not be used. Relative link [this](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1413r1.pdf).

PR #271, opened 2015-05-05
Use std::aligned_storage for static JSON null

ya1gaurav:
Patch for https://github.com/open-source-parsers/jsoncpp/issues/184
All test cases pass for x86 and Arm compiler.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -23,17 +23,15 @@
 
 namespace Json {
 
-// This is a walkaround to avoid the static initialization of Value::null.
-// kNull must be word-aligned to avoid crashing on ARM.  We use an alignment of
-// 8 (instead of 4) as a bit of future-proofing.
-#if defined(__ARMEL__)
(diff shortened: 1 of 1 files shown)


ya1gaurav:
Does travis compiler support C++-11 ?

ya1gaurav:
As seen in travis:
gcc (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3
It does not support C++11.

cdunn2001:
I'd like someone else to look at this. And note that we have to merge this into the `0.y.z` branch somehow too, where C++11 is only a special case.

ya1gaurav:
But std::aligned_storage is since C++-11, so i think it can be merged to 1.y.z branch only.

datadiode:
I dare propose the following:
```
static const union {
  char placeholder[sizeof(Value)];
  double alignment;
} kNull = {""};
```

hjmjohnson:
This issue has become stale due to inactivity.  If you want this issue to be reviewed again, please update the PR resolving the merge conflicts and re-open this issue.

Closing until updated information is provided.

PR #273, opened 2015-05-08
Cmake feature to build .so and .a obj in 1 go

jaiveer82:
This patch applies new feature of cmake program which just require single build to compile both .a and .so objects.


ya1gaurav:
What version of cmake is used in Travis?

cdunn2001:
If it's not even in Travis yet (by default), then it's too soon to expect anyone else to have it. Besides, I have not been convinced of a need for this feature. If you provide both `-DSHARED_LIBS=ON` and `-DBUILD_STATIC_LIBS=ON` to **cmake**, then you get both.

jaiveer82:
There are many platforms/systems which require both type of libraries, static as well as shared.
Moreover this is a new feature of cmake (after cmake version 2.8), If this version is not supported in cmake we can add a check.

I think Opensource communities must accept enhancements done by dependencies they use.
What is your opinion?

cdunn2001:
I like to stay up-to-date. However, I think it's ok to lag Travis CI defaults.

You can build both static and dynamic libraries simultaneously today in jsoncpp with a single cmake/make invocation. If you want that to be the default, that's something we could discuss.

PR #278, opened 2015-05-19
Ensure Json::Value::null{,Ref} are initialized at compile time.

hchunhui:
Demo:
Compiler: gcc 4.9.2
The following code crashs:

```
#include <json/json.h>
#include <string>

class A {
public:
    A() {
        std::string json = "{ \"x\":3 }\n";
        Json::Value root;
        Json::Reader reader;
        reader.parse(json, root);
    }
};

A a;
int main(){}
```

This bug was introduced in commit 48d9a92a1b4621e69533af543c1f0611abc47ddc.

gcc now puts Json::Value::null{,Ref} into bss section:

```
  $ nm libjsoncpp.a | c++filt | grep Value::null
  0000000000000000 B Json::Value::null
  0000000000000008 B Json::Value::nullRef
```

When we access them in constructor of global objects, since the order of
initialization is unpredictable, we may dereference a null pointer.

After applying this patch, gcc puts them into rodata section:

```
  $ nm libjsoncpp.a | c++filt | grep Value::null
  0000000000000008 R Json::Value::null
  0000000000000010 R Json::Value::nullRef
```

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -230,9 +230,12 @@ public:
 #endif // ifndef JSON_USE_CPPTL_SMALLMAP
 #endif // ifndef JSONCPP_DOC_EXCLUDE_IMPLEMENTATION
 
-public:
+private:
   struct StaticInitTag {};
-  Value(StaticInitTag); ///< A special constructor for Value::null.
+  Value(StaticInitTag); ///< A special constructor for Value::kNull.
+  static const Value kNull;
+
+public:
   /** \brief Create a default Value of the given type.
 
     This is a very useful constructor.
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -246,6 +246,8 @@ Json::Value arr_value(Json::arrayValue); // []
 Json::Value obj_value(Json::objectValue); // {}
 \endcode
   */
+  struct StaticInitTag {};
+  Value(StaticInitTag);
   Value(ValueType type = nullValue);
   Value(Int value);
   Value(UInt value);
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -232,7 +232,7 @@ public:
 
 private:
   struct StaticInitTag {};
-  Value(StaticInitTag); ///< A special constructor for Value::kNull.
+  explicit Value(StaticInitTag) {} ///< A special constructor for Value::kNull.
   static const Value kNull;
 
 public:
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -231,6 +231,8 @@ public:
 #endif // ifndef JSONCPP_DOC_EXCLUDE_IMPLEMENTATION
 
 public:
+  struct StaticInitTag {};
+  Value(StaticInitTag); ///< A special constructor for Value::null.
   /** \brief Create a default Value of the given type.
 
     This is a very useful constructor.
@@ -246,8 +248,6 @@ Json::Value arr_value(Json::arrayValue); // []
 Json::Value obj_value(Json::objectValue); // {}
 \endcode
   */
-  struct StaticInitTag {};
-  Value(StaticInitTag);
   Value(ValueType type = nullValue);
   Value(Int value);
   Value(UInt value);
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -23,14 +23,6 @@
 
 namespace Json {
 
-// This is a walkaround to avoid the static initialization of Value::null.
-// kNull must be word-aligned to avoid crashing on ARM.  We use an alignment of
-// 8 (instead of 4) as a bit of future-proofing.
-#if defined(__ARMEL__)
-#define ALIGNAS(byte_alignment) __attribute__((aligned(byte_alignment)))
-#else
-#define ALIGNAS(byte_alignment)
-#endif
 static const Value kNull((Value::StaticInitTag()));
 const Value& Value::null = kNull;
 const Value& Value::nullRef = kNull;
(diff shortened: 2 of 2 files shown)


cdunn2001:
1. Which branch are you using? `master` (aka `1.y.z`) or `0.y.z` (binary-compatible with the old `0-6.0-rc2`).
1. Global `null` was a blunder, we know, we know. But we cannot drop it without breaking binary- and source-code- compatibility. Suggestions are welcome.
1. What about the error mentioned in 48d9a92? Does this fix that too?
1. Even if we want this, we cannot merge it until Travis CI passes.

BillyDonahue:
I believe there's a way out here.
We've said that a Value* filled with 0-valued bytes has the null value. We just need a special constructor that does nothing, and a destructor that does nothing for that null-valued case.


class Value {
  struct StaticInitTag {};
  Value(StaticInitTag) {}  // must do absolutely nothing
};

// then you can declare this at namespace scope:
const Value Value::null((StaticInitTag()));

And it should be the case that the Value::null object has the same bit-pattern before and after static initialization.

hchunhui:
Thank you, @cdunn2001 !
1. I am using `master`.
2. I think this patch will not break souce-code- and binary- compatibility.
3.
48d9a92a1b4621e69533af543c1f0611abc47ddc trys to "fix" the "strict aliasing" warning. 
But with this commit, gcc will not to put `Value::null` into rodata.
So if we use jsoncpp in a constructor of a global object, the program may crash.
4.
I simply "revert" that commit, so gcc shows the warning again.
In release mode, the "-Werror=strict-aliasing" flag is set, that's why Travis CI failed.
I think the "strict aliasing" issue is always there, if we use `kNull` to allocate space for `Value::null`. 48d9a92a1b4621e69533af543c1f0611abc47ddc just cheats the compiler. 
The issue won't cause real problems either, because we will not use `kNull`.
So can we drop the "-Werror=strict-aliasing" flag?

BillyDonahue:
I still think aliasing is a gamble with UB that we don't have to take. (see
above)

hchunhui:
Thank you, @BillyDonahue !
I think your scheme is better, I have modified the code.
In order to keep compatibility, I use an extra `kNull`, and keep `Value::null` as a refrence to `kNull`.

cdunn2001:
> I still think aliasing is a gamble with UB that we don't have to take.

@BillyDonahue, What is UB? Universal Binaries?

> In order to keep compatibility, I use an extra `kNull`, and keep `Value::null` as a refrence to `kNull`.

@hchunhui, I'll take a look.

cdunn2001:
@BillyDonahue, Looks good. But I'm curious: Will the constructor definitely fill-in zeroes? And does it even matter?

hchunhui:
The special constructor does nothing.
Static variable should be put in bss, and filled by zeros.

BillyDonahue:
On Fri, May 22, 2015 at 1:54 AM, Christopher Dunn <notifications@github.com>
wrote:

> I still think aliasing is a gamble with UB that we don't have to take.
>
> @BillyDonahue <https://github.com/BillyDonahue>, What is UB? Universal
> Binaries?
>

"undefined behavior"

 In order to keep compatibility, I use an extra kNull, and keep Value::null

BillyDonahue:
As hchunhui says, the special ctor must do absolutely nothing. You'd only use it with static duration Value objects which would be linker-initialized to be zero-filled. The advantage is that you're never lying to the compiler about the type associated with the underlying storage, and it's completely legitimate code.

include/json/value.h
```diff
@@ -231,6 +231,8 @@ class JSON_API Value {
 #endif // ifndef JSONCPP_DOC_EXCLUDE_IMPLEMENTATION
 
 public:
+  struct StaticInitTag {};
```
BillyDonahue: Actually I think StaticInitTag might need to be private.

src/lib_json/json_value.cpp
```diff
@@ -23,18 +23,9 @@
...
-#if defined(__ARMEL__)
-#define ALIGNAS(byte_alignment) __attribute__((aligned(byte_alignment)))
-#else
-#define ALIGNAS(byte_alignment)
-#endif
-static const unsigned char ALIGNAS(8) kNull[sizeof(Value)] = { 0 };
-const unsigned char& kNullRef = kNull[0];
-const Value& Value::null = reinterpret_cast<const Value&>(kNullRef);
-const Value& Value::nullRef = null;
+static const Value kNull((Value::StaticInitTag()));
```
BillyDonahue: The word 'static' here is redundant. namespace-scope const always have internal linkage by default.
If kNull as a (preferably private) member of Value, it would have access to struct StaticInitTag, even if that struct was also made private. I think that's a good idea to do. We might not want to publish the hackery to client code. We don't need client code relying on the ability to create a null-valued Value from a block of zeros.

hchunhui:
@BillyDonahue 
I have modified the code according to your comments. Is that what you want?

src/lib_json/json_value.cpp
```diff
@@ -308,6 +299,8 @@ bool Value::CZString::isStaticString() const { return storage_.policy_ == noDupl
 // //////////////////////////////////////////////////////////////////
 // //////////////////////////////////////////////////////////////////
 
+Value::Value(Value::StaticInitTag) {}
```
BillyDonahue: Maybe give an in-class definition for Value::Value(StaticInitTag) {}

BillyDonahue:
looks good. I hope it tests well.

src/lib_json/json_value.cpp
cdunn2001: And mark that ctor `explicit`.

cdunn2001:
If this is still worth doing, rebase and reopen.

PR #282, opened 2015-05-20
Clarify which parts of README for users vs devs

keithkml:
When I arrived at the JsonCpp GitHub page, as an intermediate C++ developer, I could not figure out how to include JsonCpp into my project. The changes I propose to the README make this much clearer, and define a clear distinction between which instructions are for those developing and contributing to JsonCpp, and those who are just using it.


cdunn2001:
Note that there are many consumers who do not use amalgamated headers. Many in fact use **cmake**.

I've merged your change, but it's awkward to update the README too often, since we maintain an old branch for backward-compatibility. Plus, it needs to stay in sync with the source code, as we something change the **cmake** variables. But feel free to update the wiki if you think you can offer some clarity. You should have edit permissions already.

keithkml:
Thanks @cdunn2001! I think this will really help users get started faster. After all this is the age of NPM and Bower :)

PR #285, opened 2015-05-31
Removed locale dependency:

jeroenwalter:
- use stringstream instead of sprintf/sscanf
- removed fixNumericLocale hack
- some test still fail, this is due to incorrect python test runner.


cdunn2001:
The tests fail. Do you want to try another way, or should I close this?

jeroenwalter:
This can be closed, I won't update the tests, because the problem lies with the python test runner imo and I don't know python.

PR #292, opened 2015-06-11
Update config.h

eyjian:
include/json/config.h:100: error: ISO C++ does not support 'long long'
include/json/config.h:101: error: ISO C++ does not support 'long long'

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -97,8 +97,8 @@ typedef unsigned int LargestUInt;
 typedef __int64 Int64;
 typedef unsigned __int64 UInt64;
 #else                 // if defined(_MSC_VER) // Other platforms, use long long
-typedef long long int Int64;
-typedef unsigned long long int UInt64;
+__extension__ typedef long long int Int64;
+__extension__ typedef unsigned long long int UInt64;
 #endif // if defined(_MSC_VER)
 typedef Int64 LargestInt;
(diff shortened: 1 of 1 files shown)


cdunn2001:
This is for VisualStudio? Which versions should this work for? We might have to revert if we get complaints.

We appreciate the submission though. If the AppVeyor tests pass, I'll merge this.

eyjian:
$ g++ --version
g++ (GCC) 4.1.0 (SUSE Linux)

$ make --version
GNU Make 3.80

Reference：http://blog.chinaunix.net/uid-20682147-id-5078298.html

cdunn2001:
Oh, the comment mislead me: `#else  // if defined(_MSC_VER) // Other `. It should say the opposite.

So this change is for all compilers *except* MSC? That's a non-starter. `__exception__` is not a universal symbol.

Maybe we just need [`-Wno-long-long`](https://gcc.gnu.org/ml/gcc/2003-08/msg00847.html) for old gcc in cmake?

Anyway, if you're using an old compiler, you're lucky it works. Use the `0.y.z` branch (or a recent `0.*.*` release). That one will maintain backward compatibility. Please let us know whether that branch works for you.

cdunn2001:
The `0.y.z` branch now has `-Wno-long-long`. For *master*, you need a more recent compiler.

Issue #296, opened 2015-07-01
Implicit 64 to 32 bit conversions in json_(value|writer).cpp


fivemicro:
Some users of jsoncpp compile in a very strict environment that uses `-Wshorten-64-to-32` and `-Werror` and thus the current revision of jsoncpp fails their build.

PR #297, opened 2015-07-01
Make 64-to-32 bit conversions explicit when needed (issue #296)

fivemicro:


cdunn2001:
Thanks!

cdunn2001:
I just noticed that AppVeyor began to fail with this build:
* https://ci.appveyor.com/project/cdunn2001/jsoncpp/build/build.92

(I had assumed AppVeyor was down when I merged this change, but actually I just had to view it in another browser.)

And the problem has affected someone else: #300.

So I will have to revert this change, unless you have a more complete fix.

fivemicro:
@cdunn2001 - Sorry, I am new to AppVeyor, but I read one of the failed build logs and it looked like there were missing tools?  Specifically:

https://ci.appveyor.com/project/cdunn2001/jsoncpp/build/build.92/job/i2224pf8kr58y854#L55

says
...
...
```
C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\Microsoft.Cpp.Platform.targets(64,5): error
MSB8020: The build tools for v140 (Platform Toolset = 'v140') cannot be found. To build using the v140 
build tools, please install v140 build tools.  Alternatively, you may upgrade to the current Visual Studio 
tools by selecting the Project menu or right-click the solution, and then selecting "Upgrade Solution...". 
[C:\projects\jsoncpp\ZERO_CHECK.vcxproj]
```

I am not sure how my changes caused that.  Still, if you need to revert, I understand.

cdunn2001:
Yeah, maybe we just need the Toolset.

PR #299, opened 2015-07-03
Fix undefined name "sys"

mnaberez:
`sys.stderr` is used in an `except` clause of `fix_source_eol()` but it will cause `NameError: name 'sys' is not defined` because the `sys` module is never imported.


cdunn2001:
AppVeyor is down, but this is simple.

PR #301, opened 2015-07-05
Gcc warning unused "tokenStart"

ya1gaurav:
In 0.x.y branch, tokenStart is not used.
Comment it as done in other cases.


cdunn2001:
See #307.

PR #309, opened 2015-07-11
dockerize the Travis build, and allow C++11

cdunn2001:
We will not actually set `-std=c++11` until v2, but we might turn it on for builds.


cdunn2001:
Just to be clear: The builds are not faster in Docker, but the machines become available faster. The builds also tend to be less sequential. So overall, less wall-clock time.

PR #316, opened 2015-07-14
Update json_tool.h

filipjs:
Fix a typo in comment.


significant-bit:
I think the comment's last [ was meant to indicate an open range. Either way the comment doesn't match the code! Suggest changing to [1..31] or even better describing what makes control chars deserve special treatment.

Also what about 0x7F (DEL)?

cdunn2001:
You're right: `[1..32)` or `[1..31]` would be better.

I'm agnostic on these kinds of changes. Submit a pull-request, and I'll merge it.

PR #321, opened 2015-07-30
CZString assignment operator - pass by reference

ya1gaurav:
The assignment operator of CZString is not passed by reference. It is reported as issue in static analyser tool.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -213,7 +213,7 @@ private:
     CZString(char const* str, unsigned length, DuplicationPolicy allocate);
     CZString(CZString const& other);
     ~CZString();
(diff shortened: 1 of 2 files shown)


BillyDonahue:
I think this change is a mistake and I'd like to revert it.

What static analyzer are we talking about?

BillyDonahue:
This is an incorrect change. You've left the operator= modifying the source.

    Value::CZString& Value::CZString::operator=(CZString& other) {
       swap(other);
       return *this;
     }

This will modify the source object!

    CZString a = "hello";
    CZString b = "goodbye";
    a = b;
    // now a=="goodbye" (fine) but b="hello" !

PR #323, opened 2015-07-30
Remove -Werror

joerg-krause:
`-Werror` shouldn't be used in released code since it can cause random build failures on moderate warnings. It also depends on the used toolchain since different toolchains may or may not print the same warnings.


cdunn2001:
Agreed. But we should try to inject it into the Travis CI build.

PR #334, opened 2015-08-14
Use a consistent target name for jsoncpp in CMake scripts

cdunn2001:
(See #325.)

The target name for the jsoncpp library would change based on whether
we were building a static or shared library. This inconsistency made
it difficult and unintuitive to pull in jsoncpp as a submodule in
another repository and link to it directly via other CMake scripts.

Having a consistent target name will allow libraries with their own
CMake scripts to reliably refer to jsoncpp as a dependency.

Other Changes:
- `BUILD_SHARED_LIBS` and `BUILD_STATIC_LIBS` removed in favor of
  `JSONCPP_LIBRARY_TYPE`, which allows you to pick either `SHARED`
  or `STATIC` library variations. This change was made to prevent
  both shared and static libraries being built at the same time.
  This isn't allowed anymore since we only generate 1 target for the
  jsoncpp library now.
- `travis.sh` build script updated to perform CMake generation in an
  out-of-source binary directory. This will prevent the temporary
  generated output files from intermixing into the source tree and
  allow for multiple generations with different configurations using
  the same source tree.


cdunn2001:
@cinemast, you can pull `cmake-adjustments` from here or from my fork. Please let us know how your testing goes.

cinemast:
Hi!

I haven't tested it yet. But I think instead of using `JSONCPP_LIBRARY_TYPE` you should use the standard variable, that is provided by CMake: `BUILD_SHARED_LIBS`. 

You than can ommit the STATIC and SHARED specification in the `add_library` command. This would also further minimize the CMakeLists.txt file.

http://www.cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html

cdunn2001:
I like that. @rcdailey, could you submit a new PR with BUILD_SHARED_LIBS.

rcdailey:
@cdunn2001 Didn't know about that one. But I'm confused. I thought you guys had decided my changes could not be accepted. I'm starting to lose track of this discussion because it's split across multiple issues and PRs.

Can you centralize the discussion please? Also let me know if you guys have changed your mind and why. I don't want to invest more time into this if it's not going to work anyway.

Thanks. I will wait for your response before amending my branch.

rcdailey:
Actually I take it back. `BUILD_SHARED_LIBS` isn't a good idea. Because it impacts other targets. So if I do `add_subdirectory()` to pull in multiple CMake-driven third party libraries, setting `BUILD_SHARED_LIBS` to `ON` for jsoncpp will also force others to build as a shared library.

I understand using this makes things slightly cleaner, but ideally the settings driving JSONCPP should be specific to it only.

There *is* the unrelated argument of whether or not using `add_subdirectory()` "is the right thing to do". I'm currently looking into this and will let you guys know what they recommend. Honestly there seems to be 3-4 ways of doing the same thing (as far as fulfilling dependencies), so it's time to find out what is the best / recommended way from the authors.

cinemast:
I will also step back from this discussion (which I should have done earlier), since I am causing more confusion than useful contribution to this project. Whatever you decide to do is fine with me.

cdunn2001:
> I thought you guys had decided my changes could not be accepted.

Nope! :=)  Cinemast found a way to run cmake twice within the Debian build, so your ideas are back on the table. I'd just like to get this "right" in some sense, as simple as possible, and reasonably consistent with common practices. But I am far from a cmake expert. Unless someone else joins this discussion, I'll trust your judgement.

cdunn2001:
Macosx does not (readily) support static libraries, so I tend to agree that static and shared should be built separately.

But your travis.sh builds both shared and static for osx.... I guess that's fine. Just b/c the static archives can be built does not mean that anyone must use them.

Any objections to merging this?

PR #339, opened 2015-09-03
Floating-point NaN or Infinity values should be allowed as a feature …

Dani-Hub:
…#209

Introduce 'allowSpecialFloats' for readers and 'useSpecialFloats' for writers, use consistent macro snprintf definition for writers and readers, provide new unit tests for #209

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -321,6 +321,9 @@ public:
         the JSON value in the input string.
     - `"rejectDupKeys": false or true`
       - If true, `parse()` returns false when a key is duplicated within an object.
+    - `"allowSpecialFloats": false or true`
+      - If true, special float values (NaNs and infinities) are allowed 
+        and their values are lossfree restorable.
 
     You can examine 'settings_` yourself
     to see the defaults. You can also write and read them just like any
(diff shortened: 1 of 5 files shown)


cdunn2001:
I like this change. It seems good as-is. But I do have one idea: `NaN/Infinity/-Infinity` are just conventions, right? They could be any 3 strings. We could use something easier to parse, like `_NaN/_PosInfinity/_NegInfinity`. But I guess this is really for Javascript, yes?

Dani-Hub:
They are "just" conventions, but they are very popular ones, as described in the rationale of my issue #209. Please don't change these values lightheartedly, since doing that would break compatibility to gson, json5, simplejson, and probably several other json libraries.

BillyDonahue:
FYI: I did a little more testing and found a parsing bug.
https://github.com/Dani-Hub/jsoncpp/pull/2

Dani-Hub:
Thanks for bug-finding test, Billy! The revised implementation should honor you test cases now properly.

cdunn2001:
I'll try to cherry-pick this to 0.y.z too, but that's probably not a big deal.

PR #340, opened 2015-09-04
C++11 enhancement for reader.h

ya1gaurav:
Adding override keyword for virtual functions in derived class.
Source : http://en.cppreference.com/w/cpp/language/override


cdunn2001:
Good idea! I'll pull all these and rebase this weekend.

cdunn2001:
#347

PR #347, opened 2015-09-05
C++11: override keyword

cdunn2001:
Source : http://en.cppreference.com/w/cpp/language/override


cdunn2001:
Thanks!

cdunn2001:
Btw, separate commits are fine, but I wanted them all in the same pull-request, for a linear commit-history.

PR #352, opened 2015-09-07
C++11 enhancement - Use range-based for

ya1gaurav:
C++11 provides enhancement for range-based for loops specially for containers. It improves readability.
Reference : http://en.cppreference.com/w/cpp/language/range-for

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -817,10 +817,7 @@ std::string Reader::getFormatedErrorMessages() const {
 
 std::string Reader::getFormattedErrorMessages() const {
   std::string formattedMessage;
-  for (Errors::const_iterator itError = errors_.begin();
-       itError != errors_.end();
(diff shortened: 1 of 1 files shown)


cdunn2001:
Is there any performance improvement? Faster compilation?

ya1gaurav:
1. Easy to code.
2. Improved readability.
3. No need to dereference the iterator to get the value.

cdunn2001:
None of those advantages apply to users.

In general, I don't like altering a code-base for aesthetic reasons. If we accept these changes, I expect a never-ending supply of similar submissions. And in this case, it would increase the disparities between the new and old branches, making the project harder to maintain. The old branch is actually the more popular.

ya1gaurav:
Its ok, these are just c++-11 style coding.

cdunn2001:
Thanks for understanding. I don't mean to offend you. We always appreciate contributions.

PR #363, opened 2015-09-21
Fix missing initialization in Constructor.

ya1gaurav:
OurFeatures() Constructor is not initializing variable rejectDupKeys_.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -909,6 +909,7 @@ OurFeatures::OurFeatures()
     , allowDroppedNullPlaceholders_(false), allowNumericKeys_(false)
     , allowSingleQuotes_(false)
     , failIfExtra_(false)
+    , rejectDupKeys_(false)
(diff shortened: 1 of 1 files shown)


pavel-pimenov:
cppcheck 1.69 
Member variable 'OurFeatures::stackLimit_' is not initialized in the constructor.
Member variable 'OurReader::stackDepth_' is not initialized in the constructor.

BillyDonahue:
pavel has a point. Maybe go a little further and initialize the other fields while we're here.

ya1gaurav:
stackLinit_ initialized in https://github.com/open-source-parsers/jsoncpp/pull/364
It wil make OurFeatures to be free from above cppcheck , other pull request can be done for OurReader.

cdunn2001:
Instead, let's delete the whole ctor. We can accept C++ defaults b/c they are always over-ridden in `newCharBuilder()`.

But yes, let's update the OurReader ctor. I don't want to refactor that class since it is mainly a copy of old code.

PR #364, opened 2015-09-21
Add default value of stackLimit couple of places

ya1gaurav:
stackLimit default value is missing at two places.Adding them.


src/lib_json/json_reader.cpp
```diff
@@ -910,6 +910,7 @@ OurFeatures::OurFeatures()
     , allowSingleQuotes_(false)
     , failIfExtra_(false)
     , allowSpecialFloats_(false)
+    , stackLimit_(1000)
```
cdunn2001: Instead of adding this, let's remove this ctor completely. The default is fine, since we always completely over-ride the values in `newCharReader()`.

PR #366, opened 2015-09-22
parseCommandLine also throws

ya1gaurav:
Catching exceptions thrown by parseCommandLine (std::bad_alloc & std::length_error) also.


cdunn2001:
Can you be more specific about this? What was the problem?

ya1gaurav:
some static analyzer reported missing catch for throw statements in this function. I know its test code but yet it can be fixed.

Issue #350, opened 2015-09-05
__cplusplus value should not be used to decide for std::unique_ptr


Dani-Hub:
Both `json_writer.cpp` and `json_reader.cpp` currently use the following preprocessor test to decide whether to use `std::unique_ptr` instead of `std::auto_ptr`:

`#if __cplusplus >= 201103L`

This heuristics isn't sufficient for currently existing popular C++ Standard Libraries. E.g. the Dinkumware library provided with Visual Studio 2015 does provide a conforming `std::unique_ptr` implementation, but the underlying compiler still reports a `__cplusplus` value of 199711, because of a conservative policy signaling that not *all* parts of the C++ 11 Standard are provided. Nonetheless VS 2015 has an extremely good conformance for practical purposes, therefore I recommend to not rely alone on the value of `__cplusplus` which is not necessarily a sign that `std::unique_ptr` is available or not.

I would recommend to consider a more robust characteristics to accept Libraries (such as Dinkumware) providing a valid `std::unique_ptr`, especially because as of C++11 the type `std::auto_ptr` has been formally deprecated and as of C++14 it has been completely removed from the Standard Library specification. Now Dinkumware officially allows users to disable this type by defining a the macro `_HAS_AUTO_PTR_ETC` equal to `0` (as described here: http://blogs.msdn.com/b/vcblog/archive/2015/04/29/c-11-14-17-features-in-vs-2015-rc.aspx) to respect this.

Boost evaluates whether `_CPPLIB_VER` is either not defined or has a value less than 520 for this (plus some additional preprocessor gymnastics) which seems to be more reasonable way. I'm willing to contribute such a heuristics that would better support Visual Studio 2015 and beyond.

cdunn2001:
Sounds good. Let's see what you come up with. It's unlikely we'll alter the 0.y.z branch, which supports some ancient compilers, but we can probably limit ourselves to fairly recent compilers on master.

Dani-Hub:
I don't understand your remark regarding ancient compilers, because my intended fix would still keep those working using `std::auto_ptr`, or do you mean that it should be necessary to require a specific compiler version as well? This would seem like premature pessimization to me: Why should your library try to fix the craziness of a programmer who is trying to use a Standard Library basically conforming with C++14 with an completely outdated compiler?

Please take a look at my most recent pull request #351, which ended to become simpler than I thought originally.

Dani-Hub:
The replacement for #351 is now #367, which should be properly rebased.

PR #367, opened 2015-09-27
__cplusplus value should not be used to decide for std::unique_ptr #350:

Dani-Hub:
In addition to the C++ language version define __cplusplus also check _CPPLIB_VER for better Dinkumware support.


Dani-Hub:
Hopefully this rebased pull request is more appealing.

cdunn2001:
Sweet! I think we can even apply this to the 0.y.z branch.

PR #371, opened 2015-10-01
json_writer: improve isfinite support on *nix

mathstuf:
Based on a patches to CMake by:

Ådne Hovda <ahovda@openit.com>:

    commit 7b1cdb00279908cacabada92f8a53e4986465423

    jsoncpp: Provide 'isfinite' implementation on older AIX and HP-UX

    Newer AIX and HP-UX platforms provide 'isfinite' as a <math.h> macro.
    Older versions do not, so add the definition if it is not provided.

Michael Scott <michael.scott@gbgplc.com>:

    commit 9217b678b305d7df7471ba476a81bf28961fdfa3

    jsoncpp: Provide 'isfinite' impl on more HP-UX versions (#15576)

    Some versions of HP-UX do not define 'isfinite' or 'finite' in math.h
    for Itanium when preprocessing with C++, so we have to add the
    definition ourselves instead to map to the internal version.

Joerg Sonnenberger <joerg@bec.de>:

    commit 75644dafe54c21902f14cfe58cb8338b553b69d8

    jsoncpp: Fix compilation as C99 on Solaris

    In C99 mode, Solaris variants may already define isfinite, so check for
    the existence first.

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -20,8 +20,25 @@
 #include <float.h>
 #define isfinite _finite
 #elif defined(__sun) && defined(__SVR4) //Solaris
+#if !defined(isfinite)
 #include <ieeefp.h>
 #define isfinite finite
+#endif
+#elif defined(_AIX)
+#if !defined(isfinite)
+#include <math.h>
+#define isfinite finite
+#endif
+#elif defined(__hpux)
+#if !defined(isfinite)
+#if defined(__ia64) && !defined(finite)
+#define isfinite(x) ((sizeof(x) == sizeof(float) ? \
+                     _Isfinitef(x) : _IsFinite(x)))
+#else
+#include <math.h>
+#define isfinite finite
+#endif
+#endif
 #else
 #include <cmath>
 #define isfinite std::isfinite


mathstuf:
@bradking

mathstuf:
Links to the cmake commits: kitware/cmake@7b1cdb00279908cacabada92f8a53e4986465423 kitware/cmake@9217b678b305d7df7471ba476a81bf28961fdfa3 kitware/cmake@75644dafe54c21902f14cfe58cb8338b553b69d8.

If you'd like these split as they came into CMake, let me know and I'll redo the branch.

cdunn2001:
Wow. So much work for a value that is not even part of standard JSON.

Anyway, thanks! Looks good to me. One question: Can `<math.h>` by `<cmath>` on those platforms? Some compilers will warn about non-standard includes. It's not a huge problem. Just asking.

mathstuf:
Yeah, `math.h` is a fun bit of the standard-meets-reality story. As for headers, I don't know :( . Brad?

bradking:
CMake's last import of jsoncpp pre-dates the change from `math.h` to `cmath`, and `cmath` is not included anywhere in the entire CMake source tree, so I don't know whether `cmath` would work on those platforms.  As mentioned in the commit message proposed here, the changes to CMake were all contributed by others.  I don't have access to all their platforms to try it out myself.  Fortunately it is only on AIX and HP-UX that `math.h` is included by this patch.  After we next update CMake's jsoncpp version we may be able to test out `cmath` for this use case.

PR #377, opened 2015-10-06
fix a bug: in value.h, ValueConstIterator can convert to ValueIterato…

ycqiu:
…r, I think that is a bug. the correct way is ValueIterator can convert to ValueConstIterator.


BillyDonahue:
Totally. An implicit conversion allowing mutable access to const elements is the wrong direction. :)
I hope this gets fixed. Can you add a regression test for it?

ycqiu:
#include <json/json.h>
#include <iostream>
#include <sstream>
#include <string>

using namespace std;


int main()
{
	Json::Value value;

	for(int i = 10; i < 20; ++i)
	{
		stringstream sstream;	
		sstream << i;
		string str;
		sstream >> str;
		//cout << str << endl;
		value[str] = str;	
	}	

	//in old code, this will get a compile error
	Json::Value::const_iterator iter = value.begin();
	for(; iter != value.end(); ++iter)
	{
		cout << *iter << endl;
	}
	return 0;
}

ycqiu:
yes. the test is easy, i send it here

#include <json/json.h>
//#include "../jsoncpp/include/json/json.h"
#include <iostream>
#include <sstream>
#include <string>

using namespace std;


int main()
{
	Json::Value value;

	for(int i = 10; i < 20; ++i)
	{
		stringstream sstream;	
		sstream << i;
		string str;
		sstream >> str;
		//cout << str << endl;
		value[str] = str;	
	}	

	//in old code, this will get a compile error
	Json::Value::const_iterator iter = value.begin();
	for(; iter != value.end(); ++iter)
	{
		cout << *iter << endl;
	}
	return 0;
}

ycqiu:
yes. the test is easy, i send it here.  the include file can not put here.

int main()
{
Json::Value value;

for(int i = 10; i < 20; ++i)
{
    stringstream sstream;   
    sstream << i;
    string str;
    sstream >> str;
    //cout << str << endl;
    value[str] = str;   
}   

//in old code, this will get a compile error
Json::Value::const_iterator iter = value.begin();
for(; iter != value.end(); ++iter)
{
    cout << *iter << endl;
}
return 0;
}

ycqiu:
yes. I send code here. the include file can not put here.


int main()
{
	Json::Value value;

	for(int i = 10; i < 20; ++i)
	{
		stringstream sstream;	
		sstream << i;
		string str;
		sstream >> str;
		//cout << str << endl;
		value[str] = str;	
	}	

	//in old code, this will get a compile error
	Json::Value::const_iterator iter = value.begin();
	for(; iter != value.end(); ++iter)
	{
		cout << *iter << endl;
	}
	return 0;
}

BillyDonahue:
Put the test in the PR. In comments it doesn't do much good. :)

ycqiu:
ok, I puted the test.cpp in the PR, please check it.

test/test.cpp
```diff
@@ -0,0 +1,31 @@
+#include <json/json.h>
+//#include "../jsoncpp/include/json/json.h"
+#include <iostream>
+#include <sstream>
+#include <string>
+
+using namespace std;
```
BillyDonahue: Please remove this. It's a bad practice.

test/test.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+
+int main()
```
BillyDonahue: Can you just add this code to an existing test?

ycqiu:
yes. 
sorry,this is my first time to use github.
I try it mybest.

BillyGoto:
Sent from my iPhone

> On Oct 8, 2015, at 10:41, ycqiu <notifications@github.com> wrote:
> 
> yes. 
> sorry,this is my first time to use github.
> I try it mybest.
> 

Sure I just want to make this easier for you. Making a new test file means adding a build rule to build and run it. It's easier to piggyback on an existing target.

cdunn2001:
Thanks for catching this, @ycqiu.

@BillyDonahue, I wonder: What if instead of removing the ctor, we include it but mark it `explicit`?

BillyDonahue:
I don't think it's ever ok to make an `iterator` out of a `const_iterator`.
An explicit constructor could be triggered accidentally by template code or by an initializer and break const-correctness. It really should be removed.

ycqiu:
i pull the code, please check it.

cdunn2001:
You're right, of course. The conversion is a bad idea. But since it already existed, I had to keep it and just make it throw.

PR #379, opened 2015-10-10
Iterator conversions

cdunn2001:


cdunn2001:
I had to undo #378 for binary-compatibility. Even though we do not want ConstIterator -> Iterator, we must keep the symbol. So we can just let it throw.

Issue #284, opened 2015-05-29
Possible to limit precision of number types?


richardeakin:
I have a json-based config where I write some parameters, and when my application closes it uses the `Json::StyledWriter` to save the latest state of my root `Json::Value` object. I understand that round trip floating point precision isn't guarunteed to be exact, but is there a way to avoid numbers such as `0.15` being saved as `0.14999999999999999`? This makes the config file difficult to read.

One solution I was wondering about is whether I could limit the precision of floating point numbers, I rarely in this case want more than 4 significant decimals and that would at least improve the use case.

cdunn2001:
First, please use `StreamWriterBuilder`. That allows us to add extra configuration without breaking binary-compatibility.

Note that, as described in issue #98, you will not have round-trip reproducibility if you have fewer than 17 digits of precision. Despite that caveat, we could add a setting to limit the precision. However, I don't have time right now, but if you submit a PR, we could pull it.

bknecht:
I saw you just closed this issue I'm interested in. What's the resolution? I just had a quick look at the code of StreamWriterBuilder and I think it's not too difficult to do. Is your request for a PR still open? I'll sit down and have a try at it if you're still interested.

cdunn2001:
Great!

Thuenem:
I am out of the office until 25.08.2015.

Dear Business Partner, I will be out of the office starting 17.08.2015 and
will not return until 25.08.2015. Your message will not be processed during
my abscence. I will respond to your message when I return. In urgent cases
please contact Peter Winterberg under +49.5971.80819953 or mail
peter.winterberg@prognost.com. - Thank you.


Note: This is an automated response to your message  "Re: [jsoncpp]
Possible to limit precision of number types? (#284)" sent on 15.08.2015
01:28:21.

This is the only notification you will receive while this person is away.

Thuenem:
I am out of the office until 25.08.2015.

Dear Business Partner, I will be out of the office starting 17.08.2015 and
will not return until 25.08.2015. Your message will not be processed during
my abscence. I will respond to your message when I return. In urgent cases
please contact Peter Winterberg under +49.5971.80819953 or mail
peter.winterberg@prognost.com. - Thank you.


Note: This is an automated response to your message  "Re: [jsoncpp]
Possible to limit precision of number types? (#284)" sent on 15.08.2015
01:28:59.

This is the only notification you will receive while this person is away.

longlene:
[...]

PR #381, opened 2015-10-15
Precision

bknecht:
My suggestion on how to fix issue #284. Sorry, it took so long. Stressful work and vacations delayed this.

StreamWriterBuilder has now a setting `precision` that can take values between 0 and 17. Default is always 17 as well as with deprecated classes where there is no option to limit precision.

Also added a small test to test different issues with precision and with changing the precision setting.

I hope this is solving the issue. Let me know if something is missing in your opinion.

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -133,7 +133,7 @@ std::string valueToString(UInt value) {
 
 #endif // # if defined(JSON_HAS_INT64)
 
-std::string valueToString(double value, bool useSpecialFloats, int precision) {
+std::string valueToString(double value, bool useSpecialFloats, unsigned int precision) {
   // Allocate a buffer that is more than large enough to store the 16 digits of
   // precision requested below.
   char buffer[32];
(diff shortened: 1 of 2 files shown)


cdunn2001:
We'll revert if somebody complains, but I can't imagine what could go wrong. Thanks!

KarateSnowMachine:
Any idea if this is cherry-pickable to 0.y.z?

bknecht:
The change does not use any C++11 features if that's what you mean. I don't know the 0.y.z branch, but if C++11 compatibility is the only issue, the change should be safe to cherry-pick.

cdunn2001:
We don't have any automated tests on that branch. Could you cherry-pick and test it? Then submit a pull-request against that branch?

bknecht:
I'll allocate some time this month to test this.

KarateSnowMachine:
I'm planning on trying to cherry-pick this sometime in the next week or two; I'll let you know how it goes.

KarateSnowMachine:
Welp, "next week or two" certainly didn't pan out, but I've finally cherry-picked this back to 0.x.y and submitted #408

shuhongwang-zzu:
Maybe the version of jsoncpp is inconsistent, mine is
JSONCPP_STRING valueToString(double value,
unsigned int precision,
PrecisionType precisionType) {
return valueToString(value, false, precision, precisionType);
}
I have no ideas.

PR #382, opened 2015-10-19
Support ccache

cdunn2001:
http://stackoverflow.com/a/24305849/263998


ya1gaurav:
Is it showing some compile time improvements in travis automatic build?

cdunn2001:
Probably not. Not all CI systems can keep the cache between builds. But this helps *me* a lot.

PR #386, opened 2015-10-27
Remove -pedantic for gcc

joerg-krause:
`-pedantic` issues all warnings demanded by strict ISO C/C++; rejecting extensions that do not follow ISO C/C++. Without this option, certain GNU extensions and traditional C/C++ features are supported as well. Is this really intended for building jsoncpp?

For instance, building jsoncpp with the `-pedantic` option fails for musl toolchains on x86 because of an incompatible posix_memalign declaration. However, without `-pedantic` there is no error anymore.

[1] https://gcc.gnu.org/ml/gcc-patches/2015-05/msg01425.html


cdunn2001:
Can you think of a way to make this a **cmake** option? ISO compliance should be the default.

Alternatively, here is a simpler idea: I wouldn't mind this being removed completely for the old 0.y.z branch, and if you are using old tool-chains, maybe you should be using that branch.

joerg-krause:
Making this an CMake option is okay for me. I'll close this for now and will come up with a new PR.

PR #387, opened 2015-10-29
Disable -pedantic conditionally; add -pedantic-errors and -Werror conditionally

joerg-krause:
Add an option JSONCPP_WITH_STRICT_ISO to disable compilation with `-pedantic` for GCC. If jsoncpp is build with the JSONCPP_WITH_WARNING_AS_ERROR option `-pedantic-errors` is used instead.

Commit 912d550 disabled `-Werror`. Enable it if jsoncpp is build with JSONCPP_WITH_WARNING_AS_ERROR=ON.


cdunn2001:
Sweet. Thanks.

PR #393, opened 2015-12-05
Change ${CMAKE_BINARY_DIR} to ${CMAKE_CURRENT_BINARY_DIR}

ds283:
- if building as a submodule of another repository, installation of pkg-config files can fail because they may not be in the top-level binary directory

- this seems to happen because the CMake INSTALL command references ${CMAKE_BINARY_DIR}

- changing the INSTALL command to use ${CMAKE_CURRENT_BINARY_DIR} allows CMake to find the files for installation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,7 +129,7 @@ IF(JSONCPP_WITH_PKGCONFIG_SUPPORT)
 		"pkg-config/jsoncpp.pc.in"
 		"pkg-config/jsoncpp.pc"
 		@ONLY)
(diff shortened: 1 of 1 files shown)


cdunn2001:
LGTM, but I need another pair of eyes on this.

rpavlik:
Yes, that is more correct than the previous code. The configure file doing the generation is right above and uses a relative output path, and per https://cmake.org/cmake/help/v3.0/command/configure_file.html relative output paths are interpreted with respect to the current binary directory.

cdunn2001:
Thanks, Ryan! (@rpavlik)

PR #395, opened 2015-12-09
Localize stackDepth to each Reader instance

tekknolagi:
Previously it was a global variable, which renders Json::Reader thread unsafe.

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -240,6 +240,7 @@ private:
   std::string commentsBefore_;
   Features features_;
   bool collectComments_;
+	int stackDepth_;
 };  // Reader
 
 /** Interface for reading JSON from a char array.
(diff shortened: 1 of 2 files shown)


Zigsfi:
Your spelling in the commit message is incorrect; please correct "instace" to "instance"

tekknolagi:
Should be fixed now.

tekknolagi:
I noticed some commentary about `OurReader` vs `Reader` and wanting to keep `Reader` as is for binary compatibility. I may have mis-read. Can a contributor please explain the difference? Or how I can use this change (that already exists in `OurReader`)?

cdunn2001:
On the latest branch (master),`Reader` is deprecated. (Not sure how to make that more clear. `OurReader is fine but it's internal, not visible to users.) Use `CharReaderBuilder`, as described here:
* http://open-source-parsers.github.io/jsoncpp-docs/doxygen/index.html

On the old branch, `Reader` cannot be altered when it would break binary-compatibility.

So we can't use this change. Sorry.

Issue #397, opened 2015-12-18
Zero after decimal point is truncated


ksl67:
I recently upgraded to JsonCpp version 1.6.5.  

I have noticed that floating point numbers where the digit after the decimal point is zero are truncated when printed. For example `6.0` is printed as `6` 

The previous version I was using (v0.6.0) did not do this.

Is this behaviour expected?

**UPDATE**

The following code demonstrates the issue described:

```
Json::Value object;
object["float1"] = 2.0f;
object["float2"] = 2.1f;

Json::StyledWriter styledWriter;
std::cout << styledWriter.write(object);
```

The output when run on OS X using Xcode 6.3.2 is:

```
{
   "float1" : 2,
   "float2" : 2.0999999046325684
}
```

cdunn2001:
I don't have time to address this myself, but if someone submits a small change to write the decimal point for floats always, we're interested.

ksl67:
If you can point to the relevant part of the code I might be able to take a look.

cdunn2001:
Well, I would look at this line:
* https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_writer.cpp#L143
* called from
* https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_writer.cpp#L916

But from that, I see the decimal is *always* written, which means that you are probably using the old API. (We have to maintain old behavior for the old API, even when it seems strange.) Try `StreamWriterBuilder`.

baylesj:
Closing due to inactivity, and looks like it has been fixerd by #401 . If this issue is still relevant, please feel free to comment/reopen/open a new issue as appropriate.

PR #401, opened 2016-01-07
Adding truncated decimal point for floats

ya1gaurav:
https://github.com/open-source-parsers/jsoncpp/issues/397
Fix for above mentioned issue.


cdunn2001:
LGTM. But we have 2 open bug reports. I'd rather not update anything until those are addressed.

okodron:
Maybe it's better to add a separate test case mentioned in #397.

hjmjohnson:
@okodron This is outdated and no longer merges.  It seems that this issue may have been resolved in another patch, but that can not be 100% confirmed. 

If this issue persists, please re-apply your small changes to the current code base to resolve this issue.

PR #244, opened 2015-04-17
Appveyor

cdunn2001:
re: #242

Test auto appveyor run...

PR #231, opened 2015-03-31
Revert "Use std namespace for snprintf."

baruchsiach:
This reverts commit 1c58876185d2a4ed87dac4a54b82f607e74f55fd.

std::snprintf() is only available in C++11, which is not provided by
all compilers. Since the C library snprintf() can easily be used as a
replacement on Linux systems, this patch changes jsoncpp to use the C
library snprintf() instead of C++11 std::snprintf(), fixing the build error
below:

src/lib_json/json_writer.cpp:33:18: error: 'snprintf' is not a member of 'std'

PR #406, opened 2016-01-25
std::snprintf not part of std for MinGW32 using c++11

magnific0:
Unfortunately MinGW32 (GCC 4.8.1-4) refuses to bulid jsoncpp. Giving the following error:

    jsoncpp\src\lib_json\json_reader.cpp:33:18: error: 'snprintf' is not a member of 'std'
       #define snprintf std::snprintf

This issue has been covered before (#231 and #244). Although this is most likely a bug in MinGW32, but as the project is inactive (and it is still shipped with the Windows build of QtCreator), I propose to add the following if clause here. If MinGW32 is detected std::snprintf is not used, but simply the old C library snprintf instead.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -30,7 +30,9 @@
 #elif defined(__ANDROID__) || defined(__QNXNTO__)
 #define snprintf snprintf
 #elif __cplusplus >= 201103L
+#if !defined(__MINGW32__)
 #define snprintf std::snprintf
+#endif
 #endif
 
 #if defined(__QNXNTO__)
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -57,7 +57,9 @@
 #elif defined(__ANDROID__) || defined(__QNXNTO__)
 #define snprintf snprintf
 #elif __cplusplus >= 201103L
+#if !defined(__MINGW32__)
 #define snprintf std::snprintf
+#endif
 #endif
 
 #if defined(__BORLANDC__)  


BillyDonahue:
It's surprising that the #define line is being checked for validity. I don't get that.

Are we even using snprintf for anything in those files?
https://github.com/open-source-parsers/jsoncpp/search?utf8=%E2%9C%93&q=snprintf
I think it would be smart to just remove the #define if it's unused.

If we need standard float formatting, we can probably get it more portably out of the ostream library.

magnific0:
Hi @BillyDonahue, thanks for the quick reply. As you say snprintf is unused in the two files that caused me problems. I did not remove the definitions as I do not presume to know their purpose. Removing them will be an equably agreeable solution. I will update my pull request shortly!

BillyDonahue:
The only way the macros could affect you is if the token 'snprintf' is used in the expansion of a macro defined in one of the headers. But it isn't, so you're good.

cdunn2001:
```cpp
$ git grep -n snprintf
src/lib_json/json_reader.cpp:24:#define snprintf sprintf_s
src/lib_json/json_reader.cpp:26:#define snprintf std::snprintf
src/lib_json/json_reader.cpp:28:#define snprintf _snprintf
src/lib_json/json_reader.cpp:31:#define snprintf snprintf
src/lib_json/json_reader.cpp:33:#define snprintf std::snprintf
src/lib_json/json_reader.cpp:813:  snprintf(buffer, sizeof(buffer), "Line %d, Col
src/lib_json/json_reader.cpp:1817:  snprintf(buffer, sizeof(buffer), "Line %d, Co
...
```
`snprintf` is used there. I don't know why the github search does not see it. Oh, I see:

    Showing the top six matches. Last indexed on Oct 27, 2015.

Issue #411, opened 2016-02-03
GCC6 Support


cinemast:
Hi!

At Debian we are currently preparing the migration to GCC6. The following issues are currently affecting libjsoncpp.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=811958

Could you take a look at it?

cdunn2001:
This weekend.

Any chance you can update to 1.6.x? The link refers to 0.10.5.

cinemast:
Tha transition is currently in progress. It might take a few days until its completed.

https://release.debian.org/transitions/html/auto-libjsoncpp.html

So you think those problems should be fixed in 1.6.x?

cinemast:
I tried rebuilding the latest release with GCC-6, resulting in the following build errors:

http://paste.debian.net/378673/

cinemast:
Also while we are at it, could you take a look at current debian patches and check if any of those could be imported upstream?

http://anonscm.debian.org/cgit/collab-maint/libjsoncpp.git/tree/debian/patches

cdunn2001:
> So you think those problems should be fixed in 1.6.x?

No, but the fix depends on compiler version. (`std::auto_ptr<>` -> `const std::scoped_ptr<>`) I'd rather not fix this at all in the old code. If we can fix the conversion errors from your latest build, that's a better path forward.

> Also while we are at it, could you take a look at current debian patches and check if any of those could be imported upstream?

Sure! This weekend.

cinemast:
Of course, I am working on dropping 0.x.y support in the next Debian release. Thanks for taking the effort.

Greetings

cdunn2001:
Patched 0.y.z and fixed 1.y.z. Let's try again.

cinemast:
Hi Cristopher!
With the latest git head on master, the errors are still mostly the same:

```
mkdir build
root@cinemast-ThinkPad-T410s:/tmp/jsoncpp# cd build/
root@cinemast-ThinkPad-T410s:/tmp/jsoncpp/build# cmake ..
-- The C compiler identification is GNU 6.0.0
-- The CXX compiler identification is GNU 6.0.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
[...]

PR #414, opened 2016-02-06
Fix conversion warnings/errors

cdunn2001:
See #411.
  http://paste.debian.net/378673/

Wow. There are a lot of signed/unsigned oversights in this library. We should have turned on these compiler flags long ago. Hopefully, all the conversions are correct-by-construction (no underflow). LGTM. Comments?


cdunn2001:
I guess we should bump the minor version number b/c the symbols have changed. But the *sizes* of datatypes have not changed, so I don't think we need a major version bump.

Issue #423, opened 2016-02-14
New Thing -- API for 2.0.0


cdunn2001:
Set up pull-requests to discuss the new API, on the temporary `Thing` class.

PR #424, opened 2016-02-14
std::string -> String

cdunn2001:
Template `std::string` as `String`.

Maybe this is pointless? The issue is not so much the representation of the JSON strings in the API, but rather of _serialization_. I want to allow users to specify how we translate between internal `char*` and the public representation. Users should be able to supply their own converters:

``` cpp
template<>
fromJson<String>(char const* beg, char const* end, String* result) {
  ...
}
template<>
toJson<String>(String const& input, (*alloc)(size_t n), char const** result) {
}
```

We can provide defaults for `std::string`, but our default would not handle UTF-32, e.g.

What to you think? (See #423.)

### DIFF
--- a/include/json/thing.h
+++ b/include/json/thing.h
@@ -113,22 +113,6 @@ Json::Thing obj_value(Json::objectThing); // {}
   Thing(double value);
   Thing(const char* value); ///< Copy til first 0. (NULL causes to seg-fault.)
   Thing(const char* begin, const char* end); ///< Copy all, incl zeroes.
-  /** \brief Constructs a value from a static string.
-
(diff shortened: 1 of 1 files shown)


baylesj:
If we wanted to do something fancy like wide strings, wouldn't we want the internal representation to be using `wchar_t`? It's not clear to me what the value add of the String template is if we don't also adjust the backing field from char to a similar template, which seems bug-prone.

BillyDonahue:
Well, the CharType isn't the only thing that can vary on a std::basic_string.

I can see how there would still be value in import and export from strings that use unusual kinds of allocator, for example, even if they are still `char`-based.

PR #425, opened 2016-02-14
Number template

cdunn2001:
See #423 and #424.

### DIFF
--- a/include/json/thing.h
+++ b/include/json/thing.h
@@ -56,6 +56,7 @@ namespace Json {
 template<Integer, Float>
 class JSON_API Thing {
 public:
+  typedef int ArrayIndex; // Is this best?
   typedef std::vector<std::string> Members;
   //typedef ThingIterator iterator;
   //typedef ThingConstIterator const_iterator;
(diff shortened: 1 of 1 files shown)


include/json/thing.h
```diff
@@ -17,6 +17,7 @@
 
 /** \brief JSON (JavaScript Object Notation).
  */
+template<Integer, Float>
```
BillyDonahue: I don't get it. Why is template<Integer,Float> above a namespace?

cdunn2001:
Oops. Updated.

cdunn2001:
Any comments?

Basically, we would name the template something obvious, then typedef `Json::Value<std::string, int, double>` to it. Users would have to change a few number-conversion function calls, but that would probably improve their code.

If anyone wants their own datatypes, they must provide `toJson()` and `fromJson()` converter specializations. We would need to settle on their signatures. `toJson()` is the hardest because it involves memory allocation.

okodron:
Is it still relevant?

cdunn2001:
It's future planning. In my opinion, this is closer to how jsoncpp parsers should look. The user should choose both the internal datatypes and how they are converted to/from JSON data. (We could provide useful defaults.) Not only is that flexible, but it also would eliminate a large fraction of bug reports (here and elsewhere).

PR #426, opened 2016-02-17
Secure alloc compile time

dawesc:
Hi Chris, I hope all's well, here's a version of the secure allocator using typedefs in config similar to the way it's done with UInt; i've checked it in with 1 as secure allocator at the moment as i'd like to see what the regression test thought of it first of all. What would you prefer the default to be though?

Thanks again for your support on this, 

Christopher


include/json/value.h
```diff
@@ -230,7 +230,7 @@ class JSON_API Value {
 
     struct StringStorage {
       unsigned policy_: 2;
-      unsigned length_: 30; // 1GB max
+      size_t length_: 30; // 1GB max
```
cdunn2001: This could break binary compatibility. Was this necessary for compilation?

include/json/value.h
```diff
@@ -323,6 +323,7 @@ Json::Value obj_value(Json::objectValue); // {}
   int compare(const Value& other) const;
 
   const char* asCString() const; ///< Embedded zeroes could cause you trouble!
+  unsigned getCStringLength() const; //Allows you to understand the length of the CString
```
cdunn2001: This would force a minor-version bump. Please put this into an `#ifdef`.

src/lib_json/json_value.cpp
```diff
@@ -614,6 +638,16 @@ bool Value::getString(char const** str, char const** cend) const {
   return true;
 }
 
+#if JSON_USE_SECURE_MEMORY
+inline std::string toStdString(Json::String in) {
```
cdunn2001: Wouldn't the `std::string` copy ctor work?

src/lib_json/json_value.cpp
```diff
@@ -614,6 +638,16 @@ bool Value::getString(char const** str, char const** cend) const {
...
+  return std::string(in.data(), in.data() + in.length());
+}
+#else
+inline std::string toStdString(std::string in) {
+  return in;
+}
+#endif
+
 std::string Value::asString() const {
```
cdunn2001: If we're changing the API to accept Json::String instead of std::string, then we might as well return Json::String too. In fact, that should never be a problem (except the loss of efficiency, since std::string is reference-counted on some platforms). The real problem is when a method is change to *accept* Json::String by const reference. Users will have to alter their code. I'm fine with that (since this is a compile-time choice) but it's something to note.

But for that reason, the default should be off, I think.

cdunn2001:
Whew! Tons of changes, but way less than in that previous monster PR.

include/json/value.h
dawesc: Hi Chris, yes to correctly clean the string on unset it was i'm afraid. I'll put the ifdef around it again like you've suggested elsewhere

src/lib_json/json_value.cpp
dawesc: Perfect, i'll make the changes

src/lib_json/json_value.cpp
dawesc: Sadly no, what happens is that the allocators mismatch and it throws a fit, that's why you have to mess around with the underlying data structure and memory pointers rather than even being able to use the iterators.

include/json/value.h
dawesc: Perfect, will do.

include/json/value.h
```diff
@@ -572,6 +573,7 @@ Json::Value obj_value(Json::objectValue); // {}
     void setComment(const char* text, size_t len);
 
     char* comment_;
+    size_t len_;
```
dawesc: I need to add another #ifdef around this

include/json/value.h
```diff
@@ -594,6 +596,7 @@ Json::Value obj_value(Json::objectValue); // {}
   ValueType type_ : 8;
   unsigned int allocated_ : 1; // Notes: if declared as bool, bitfield is useless.
                                // If not allocated_, string_ must be null-terminated.
+  size_t allocatedLength_ = 0;
```
dawesc: And another ifdef here

include/json/config.h
```diff
@@ -118,6 +120,12 @@
 #  define JSON_USE_INT64_DOUBLE_CONVERSION 1
 #endif
 
+// If non-zero, the library zeroes any memory that it has allocated before
+// it frees its
+#ifndef JSON_USE_SECURE_MEMORY
+#define JSON_USE_SECURE_MEMORY 1
```
cdunn2001: `0` by default -- at least until we measure the runtime/memory cost of using this allocator.

cdunn2001: We'll have to check whether the typedefs alter the symbols. If so, then we'll need to switch those all to macros (poor-man's template). Again, that's to ensure binary-compatibility. It's painful, but our unusual care for binary-compatibility is one reason why a distro like Debain includes JsonCpp.

include/json/value.h
```diff
@@ -323,7 +323,10 @@ Json::Value obj_value(Json::objectValue); // {}
   int compare(const Value& other) const;
 
   const char* asCString() const; ///< Embedded zeroes could cause you trouble!
-  std::string asString() const; ///< Embedded zeroes are possible.
+#if JSON_USE_SECURE_MEMORY
+  unsigned getCStringLength() const; //Allows you to understand the length of the CString
```
cdunn2001: After latest commit, is `getCStringLength()` still needed?

src/lib_json/json_value.cpp
```diff
@@ -104,7 +111,11 @@ static inline char* duplicateStringValue(const char* value,
  */
 static inline char* duplicateAndPrefixStringValue(
```
cdunn2001: Note that this is file-static. That means we can change the signature of this without worrying about binary-compatibility. We don't necessarily need ifdefs for this.

But why do we need to change the signature? I think you've misunderstood the purpose. Because we needed to store string-lengths but we could not add a field, we have encoded the length onto the front of these "allocated" strings. It's a hack. But it should already provide all the information you need. When you release, just call decodePrefix first to get the length.

cdunn2001:
In our defense, I'd like to state that none of us wrote this library. Baptiste, the primary author, had good reasons for everything at the time. We're now maintaining it and hopefully letting it evolve in a positive direction. So please excuse the oddities in implementation.

People were very skeptical when I added the string-prefixing, but it has proven invaluable.

I should have more time this weekend or next. I plan to accept the typedefs (or macros, if necessary). Then you can rebase with a much smaller pull-request.

include/json/value.h
dawesc: Hi Chris, it is saving me in my code from doing 
`auto tmp = myVal.asString(); myNonSecure = std::string(tmp.data(), tmp.data().length()); `
instead i can do
`auto tmp = std::string; myNonSecure = std::string(myVal.asCString(), myVal.getCStringLength());`

I think it's worth having because it saves me doing at least one copy operation.

src/lib_json/json_value.cpp
dawesc: Hi Chris, thanks again, sorry my brain is having problems working through all the use cases, so the reasoning step by step:

1. I had to have a macro in the method prototypes because i cannot add a value anywhere so that in the case of non-secure mode i have nothing to pass in there (couldn't add a static variable because this would change the binary compatibility, could add locally but allocating and deallocating this seemed expensive when unecessary).
1. In Value it's clear, the copy or create methods will prefix the string so no need here to track the length separately
1. CZString does not use the prefixing strategy but you can use storage_.length_ to get the amount of memory that was allocated
1. CommentInfo does not store the length allocated anywhere to be properly zeroed

So what i'll do for the moment is to then use the lengths where i have it available and guess then for CommentInfo i think; would you think this ok?

dawesc:
Thanks so much for all of this Chris, it's really very much appreciated all the effort you've put in on this. Christopher

cdunn2001:
Ok. This is looking pretty good. But I'm swamped this weekend. I'll rebase as soon as I can...

dawesc:
I just wanted to re-iterate my thanks for all your support you've put in an amazing amount of time on this to help me get this through and it's thoroughly appreciated!

cdunn2001:
Would you rebase to the `JSONCPP_STRING` branch and submit a new pull-request?

dawesc:
Hi Chris, I've squashed it to https://github.com/open-source-parsers/jsoncpp/pull/436

Many thanks

Christopher

Issue #410, opened 2016-02-01
Compilation: override specifier 'override' not allowed on a destructor


DaltonSleeper:
Hi,

I'm trying to compile the JasonCPP solution but I'm getting an error about the override specifier on some of the destructors. Does anyone know how to solve this without causing any issues? This does not happen when having VC++ 2012 as target.

Windows 10 x64
Visual Studio 2012

One of many:
Error	2	error C3665: 'Json::CharReaderBuilder::~CharReaderBuilder' : override specifier 'override' not allowed on a destructor	C:\jsoncpp-master\include\json\reader.h	336	1	lib_json (Visual Studio 2010)

Best regards,
DS

cdunn2001:
I suppose we could drop `override` for the dtors. That will not alter any symbols. However, if this is only for VS 2010, I am reluctant to adapt. Which version of jsoncpp are you using?

And could you live with an ifdef for your compiler/version? If so, maybe you could submit a pull-request.

DaltonSleeper:
We are using the latest jasoncpp and are currently targeting VC++ 2012, 2010 and 2008. I would guess that the problem will appear for anything less than 2012. I have not gone through all the code and was not sure if override had any specific purpose. If it is safe to remove them we can do it on our side, but I would guess that there are others that will be affected by this.

Best,
DS

cdunn2001:
I'm fine with dropping `override` for dtors. But I'd like to see `virtual`, just as a reminder.

Could you make that change and submit a pull-request?

BillyDonahue:
I'd rather have an `#ifdef`-able override macro than a 'virtual'.
The virtual will declare a destructor virtual even if no base class has a virtual destructor. It can paper over inheritance mistakes. I'd rather have nothing at all. If it's just for documentation, let's just use /*virtual*/.
But really the compiler-conditional macro feels like a better solution than commenting. If a compiler doesn't have 'override', it serves as documentation anyway.

RelicOfTesla:
not support VS2010 on tag 1.7.0 

\include\json\allocator.h(17): error C2873: “value_type

And bad keyword with override
Json::Exception  ~Exception() throw() override;

cdunn2001:
@RelicOfTesla, I don't understand either of these.

A) Line 17 looks like this:
```
 13 template<typename T>
 14 class SecureAllocator {
 15   public:
 16     // Type definitions
 17     using value_type      = T;
```
How should we change that? Is it a C++11 thing?

[...]

PR #430, opened 2016-02-29
Remove override keyword from dtors

ya1gaurav:
This is to silent issues reported in https://github.com/open-source-parsers/jsoncpp/issues/410

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -333,7 +333,7 @@ public:
   Json::Value settings_;
 
   CharReaderBuilder();
-  ~CharReaderBuilder() override;
+  virtual ~CharReaderBuilder();
 
   CharReader* newCharReader() const override;
 
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -40,7 +40,7 @@ namespace Json {
 class JSON_API Exception : public std::exception {
 public:
   Exception(std::string const& msg);
-  ~Exception() throw() override;
+  virtual ~Exception() throw();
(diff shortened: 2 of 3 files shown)


ya1gaurav:
Is this the required patch ?

BillyDonahue:
A mistake IMO. I thought the master branch was C++11 now?

cdunn2001:
Is this definitely the *only* problem with the `master` branch on VS2012? If so, then I'd favor @BillyDonahue's suggestion of a macro. But if there are other problems, then I'd say let's just not support VS2012. `master` is C++11 now. If you can use an old commit -- maybe an older release of `1.6.*` -- fine. But the `0.10.z` branch is for old systems. We try very hard not to break backward-compatibility on that branch for any system or configuration. `master` is moving forward.

ya1gaurav:
I absolutely agree with master branch as it is C++-11, also adding "override" was patch submitted by me. 0.10.z is already before c++-11.

cdunn2001:
> I absolutely agree with master branch as it is C++-11... 0.10.z is already before c++-11.

But this is submitted to the `master` branch, which *is* C++11.

> also adding "override" was patch submitted by me

Yes, but what difference does that make? It's there now.

ya1gaurav:
I mean i sent this patch to cross check intention for - https://github.com/open-source-parsers/jsoncpp/issues/410
Its not needed,, so please close this pull as invalid.
Thanks.

PR #436, opened 2016-03-06
Support for Secure De-allocation of memory

dawesc:
This is a pull request for the support for de-allocation of memory in a secure manner


dawesc:
Sorry i misconstrued what was going on

PR #437, opened 2016-03-07
Support for Secure De-allocation of memory

dawesc:
This squashes the previous branch of secure allocator to a single commit and then merges JSONCPP_STRING in and finally fixes up any changes (essentially Json::String -> JSONCPP_STRING).


cdunn2001:
Let's move the discussion to #442.

Issue #434, opened 2016-03-06
Support MinGW when building  code as part of DLL


mtangoo:
I use amalgamated code and I found that when building jsoncpp as part of my DLL it have list of undefined reference. Checking the code, it only declare exports/imports only for MSVC. The following code in json.h works.

If I get time I will send PR but for now here is the relevant part of the code in case one can fix it. 

```cpp
#ifdef JSON_IN_CPPTL
#define JSON_API CPPTL_API
#elif defined(JSON_DLL_BUILD)
#if defined(_MSC_VER) || defined(__GNUC__)
#define JSON_API __declspec(dllexport)
#define JSONCPP_DISABLE_DLL_INTERFACE_WARNING
#endif // if defined(_MSC_VER)
#elif defined(JSON_DLL)
#if defined(_MSC_VER)  || defined(__GNUC__)
#define JSON_API __declspec(dllimport)
#define JSONCPP_DISABLE_DLL_INTERFACE_WARNING
#endif // if defined(_MSC_VER)
#endif // ifdef JSON_IN_CPPTL
#if !defined(JSON_API)
#define JSON_API
#endif
```

cdunn2001:
A PR would be helpful. Thanks.

ya1gaurav:
Shall I send PR with above code?

mtangoo:
No it's not copy and paste. This code is from amalgamated file. So you need to go through files and add missing MinGw entries

PR #438, opened 2016-03-08
MinGW support while building as DLL

ya1gaurav:
This is PR for https://github.com/open-source-parsers/jsoncpp/issues/434
It will fix reported build error.


rpavlik:
Wouldn't it make sense to check for _WIN32?

cdunn2001:
@rpavlik, Tests pass, so I'm happy. But if you can simplify the macros, that would be welcome.

PR #442, opened 2016-03-15
Jsoncpp string

cdunn2001:
After `std::string` -> `JSON_CPP`, these are the actual changes. For discussion with @dawesc of ETFlab.


dawesc:
Hi Chris, many thanks, I think it's looking good!

src/lib_json/json_value.cpp
```diff
@@ -229,10 +247,10 @@ Value::CZString::CZString(char const* str, unsigned ulength, DuplicationPolicy a
   storage_.length_ = ulength & 0x3FFFFFFF;
 }
 
-Value::CZString::CZString(const CZString& other)
-    : cstr_(other.storage_.policy_ != noDuplication && other.cstr_ != 0
-                ? duplicateStringValue(other.cstr_, other.storage_.length_)
-                : other.cstr_) {
+Value::CZString::CZString(const CZString& other) {
+  cstr_ = (other.storage_.policy_ != noDuplication && other.cstr_ != 0
```
cdunn2001: I was wondering about this change. In general, it's slightly faster to a initialization list, but not for built-in types like pointers. So I don't care. Just curious.

* http://stackoverflow.com/questions/13214241/c-constructors-vs-initialization-lists-speed-comparison

src/lib_json/json_value.cpp
```diff
@@ -248,8 +266,14 @@ Value::CZString::CZString(CZString&& other)
 #endif
 
 Value::CZString::~CZString() {
-  if (cstr_ && storage_.policy_ == duplicate)
-    releaseStringValue(const_cast<char*>(cstr_));
+  if (cstr_ && storage_.policy_ == duplicate) {
+#if JSON_USE_SECURE_MEMORY
+	  releaseStringValue(const_cast<char*>(cstr_), storage_.length_ + 1u); //+1 for null terminating character for sake of completeness but not actually necessary
```
cdunn2001: The code is the same between these 2 macro cases. I guess the comment is only meant to point out that the length is not even used in the other case. But we could eventually remove the `ifdef`, right?

include/json/value.h
```diff
@@ -323,6 +323,9 @@ Json::Value obj_value(Json::objectValue); // {}
   int compare(const Value& other) const;
 
   const char* asCString() const; ///< Embedded zeroes could cause you trouble!
+#if JSON_USE_SECURE_MEMORY
+  unsigned getCStringLength() const; //Allows you to understand the length of the CString
```
cdunn2001: Nothing calls this new method. Is it really needed? What's it for?

cdunn2001:
Almost ready to merge, but I think we need something in cmake to allow this feature to be enabled.

If it is built one way, but users supply a different setting for the macro, that would be a problem. It might be better for cmake to substitute something in some header, and to let everything rely on that.

    #define JSONCPP_USE_SECURE_ALLOC %%TO_BE_SUBSTITUTED_BY_CMAKE%%

cdunn2001:
Any objections?

cdunn2001:
An error from TravisCI:
```
[ 22%] Building CXX object src/lib_json/CMakeFiles/jsoncpp_lib.dir/json_value.cpp.o
/home/travis/build/open-source-parsers/jsoncpp/src/lib_json/json_value.cpp: In function ‘void Json::releasePrefixedStringValue(char*)’:
/home/travis/build/open-source-parsers/jsoncpp/src/lib_json/json_value.cpp:146:10: error: conversion to ‘unsigned int’ from ‘long unsigned int’ may alter its value [-Werror=conversion]
   length += sizeof(unsigned) + 1;
          ^
cc1plus: some warnings being treated as errors
```

cdunn2001:
Another one:
```
/home/travis/build/open-source-parsers/jsoncpp/src/lib_json/json_value.cpp: In function ‘void Json::releasePrefixedStringValue(char*)’:
/home/travis/build/open-source-parsers/jsoncpp/src/lib_json/json_value.cpp:145:59: error: cannot convert ‘size_t* {aka long unsigned int*}’ to ‘unsigned int*’ for argument ‘3’ to ‘void Json::decodePrefixedString(bool, const char*, unsigned int*, const char**)’
   decodePrefixedString(true, value, &length, &valueDecoded);
```

dawesc:
Hi Chris, thanks for merging this, you had asked some questions e.g.:

> I was wondering about this change. In general, it's slightly faster to a initialization list, but not for built-in types like pointers. So I don't care. Just curious.
> 
> http://stackoverflow.com/questions/13214241/c-constructors-vs-initialization-lists-speed-comparison

However they've disappeared, would you like me to answer them in this pull request or are you happy? (p.s. the above was because i had trouble to initialize the list during one of the iterations in the initialization list due to needing an ifdef in the initialization list and the code was becoming unreadable so i sacrificed cpu for maintainability however apologies it should have gone back again!)

cdunn2001:
I think we're fine. Hopefully, no-one will mind the macros too much.

Issue #389, opened 2015-11-18
NORETURN attribute on throwRuntimeError/throwLogicError


MartinBonner:
I am running into Analysis errors when running Visual Studio 2015 Code Analysis on jsconcpp.

We could do with a conditional NORETURN attribute on the throw functions.  That would:
a) suppress false positives from static code analysis 
b) possibly improve optimization opportunities.

Note that although _I_ am hitting this with VS, other people are quite likely to be running into issues on other platforms.

I am thinking of something like:

``` c++
#if defined(__GNUC__) && __GNUC__ > "version"
    #define JSONCPP_NORETURN __attribute((noreturn))
#elif defined (_MSV_VER) && _MSC_VER > "version"
    #define JSONCPP_NORETURN __declspec(noreturn)
// More cases from people who understand different compilers.
#else
    #define JSONCPP_NORETURN
#endif
...
void JSCONCPP_NORETURN throwRuntimeError(std::string const& msg);
```

It would be nice if we could just use C++11 [[noreturn]], but annoyingly, Microsoft don't support that yet.

cdunn2001:
Sounds good. Submit a pull-request.

baylesj:
Looks fixed, closing. I would generally advocate for excluding third party libraries, such as JsonCpp, from static analysis because we cannot guarantee compatibility with every single static analyzer currently on the market, a problem shared with many other third party libraries.

I'll close this issue and open a new one for C++11 compliance.

PR #445, opened 2016-03-16
Added NORETURN for throw functions.

ya1gaurav:
Resolve Issue - https://github.com/open-source-parsers/jsoncpp/issues/389


cdunn2001:
These are only used internally. It might have been better to turn the `throw*` functions into macros. But if this works, great.

PR #446, opened 2016-03-16
NORETURN for throw functions in 0.x.y branch

ya1gaurav:
Resolve issue - https://github.com/open-source-parsers/jsoncpp/issues/389


cdunn2001:
These are only used internally. It might have been better to turn the `throw*` functions into macros. But if this works, great.

PR #449, opened 2016-03-22
Use macro for `override`

### LINKED ISSUE #410: Compilation: override specifier 'override' not allowed on a destructor

cdunn2001:
b/c MS VS2010 is supposed to be C++11 but does not fulfull
the entire standard.

Resolves #410.
Re: #430.

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -77,6 +77,12 @@
 
 #endif // defined(_MSC_VER)
 
+#if defined(_MSC_VER) && _MSC_VER <= 1600 // MSVC <= 2010
+# define JSONCPP_OVERRIDE
+#else
+# define JSONCPP_OVERRIDE override
+#endif // MSVC <= 2010
+
 
 #ifndef JSON_HAS_RVALUE_REFERENCES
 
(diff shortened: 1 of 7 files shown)


ya1gaurav:
`+#if defined(_MSC_VER) && _MSC_VER <= 1600 // MSVC <= 2010
 +# define JSONCPP_OVERRIDE`
This is wrong, It should be marked as virtual. Otherwise we are changing the behaviour of class.

cdunn2001:
Are you sure it's wrong? If the base-class function is `virtual`, then so is the derived, whether we say so or not. And `override` is used only on the derived.

This was @BillyDonahue's suggestion, and I think it makes sense. Have you seen it actually fail somewhere?

ya1gaurav:
I agree with your explanation, I think I made some wrong assumption. Thanks for clarification.

PR #467, opened 2016-04-25
json/config.h: Generalize setting of JSONCPP_OVERRIDE to all compilers

jcfr:
This commit has been adapted from InsightSoftwareConsortium/ITK@1c86090

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -77,12 +77,14 @@
 
 #endif // defined(_MSC_VER)
 
-#if defined(_MSC_VER) && _MSC_VER <= 1600 // MSVC <= 2010
-# define JSONCPP_OVERRIDE
-#else
+// In c++11 the override keyword allows you to explicity define that a function
+// is intended to override the base-class version.  This makes the code more
+// managable and fixes a set of common hard-to-find bugs.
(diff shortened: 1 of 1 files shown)


jcfr:
Cc: @thewtex @aylward

thewtex:
+1

jcfr:
Cc: @cdunn2001

cdunn2001:
I'm fine with this in principle, but if there are complaints from VisualStudio users, we might have to revert this. I seem to recall inconsistent definitions for `_cpluplus` in older VS versions.

jcfr:
Thanks for reviewing and integrating. If there are any issues, I will make sure to follow up.

Otherwise, if you are willing to change the minimum required CMake version from CMake `2.8.5`  to at least CMake `3.2`, we could use the module [WriteCompilerDetectionHeader](https://cmake.org/cmake/help/v3.2/module/WriteCompilerDetectionHeader.html#module:WriteCompilerDetectionHeader).

Let me know if that sound reasonable.

cdunn2001:
That might have to wait for 2.0, unfortunately.

jcfr:
Cc: @vovythevov @fedorov

PR #468, opened 2016-05-02
Update json_reader.cpp

vraychev:
Fix a race condition in the JSON parser when called from multiple threads.

Tested with gcc 5.3.1 (fails without the fix, succeeds with the fix).

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -44,8 +44,8 @@
 #pragma warning(disable : 4996)
 #endif
 
-static int const stackLimit_g = 1000;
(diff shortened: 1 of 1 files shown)


cdunn2001:
Methods of the readers are not thread-safe for a given instance. Instances used on separate threads *are* thread-safe, except for the deprecated Reader as you've noticed. Please use `CharReaderBuilder`. Note that the implementation `OurReader` does not use global variables.

Even if the old `Reader` were not deprecated, we could not so easily add a dependency on a threading library, which is why the tests fail. Sorry we cannot incorporate this, but thanks for your contribution nonetheless.

PR #477, opened 2016-05-25
Add string conversion to other types

gelldur:
A lot of API's sends data in such style:

`"round_time":"2160"`

Noob PHP devs don't like them :)

All old tests and new one pass

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -708,6 +708,8 @@ Value::Int Value::asInt() const {
     return 0;
   case booleanValue:
     return value_.bool_ ? 1 : 0;
+  case stringValue:
+    return std::stoi(asCString());
   default:
     break;
   }
@@ -730,6 +732,8 @@ Value::UInt Value::asUInt() const {
     return 0;
   case booleanValue:
     return value_.bool_ ? 1 : 0;
+  case stringValue:
+    return (unsigned)std::stoul(asCString());
   default:
     break;
   }
@@ -753,6 +757,8 @@ Value::Int64 Value::asInt64() const {
     return 0;
   case booleanValue:
     return value_.bool_ ? 1 : 0;
+  case stringValue:
+     return std::stoll(asCString());
   default:
     break;
   }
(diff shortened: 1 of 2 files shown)


gelldur:
Duplicate: https://github.com/open-source-parsers/jsoncpp/pull/270 but C++11 :) here

cdunn2001:
I don't want to be the one to take the decision on this. If others agree, then maybe I'm crazy.

My instinct is to reject this. First, `isConvertibleTo()` is clearly documented. It tells whether the conversion is safe. A proper implementation using the new code would be to attempt the conversion and check whether an exception was thrown. That's quite expensive. And what about a value that is not fully converted, e.g. `"  57xyz"`? Is that considered "convertible"?

But my main point is that a wrapper is trivial. Why does everything need to be in a library? Or even if it makes sense to include the code, why not use external functions? E.g.
```cpp
Json::Value foo = root["x"]["y"];
Int64 x = Json::asInt64(foo);
```
The wrapper might test data-types first. Is that really too awkward?

gelldur:
Yeah shure i can make wrapper for this. But it would be nice to have this wrapper in lib :)
In Java (gson) when get such response form serwer:

```
{
	"magicInt":"223"
}
```

I can simply call:

```
JsonElement jsonElement = jsonObject.get(...)
int valueInt = jsonElement.getAsInt() // 223
String valueString = jsonElement.getAsInt() //"223"
```

But in jsoncpp i get error :(

I understand your point of view. Of course parsing `"54ezx"` as to int should throw exception

But in the end such "util" should be in library this is to common :)

Dark-Passenger:
Would love to have this in the library, already if I try to get a type string asInt it throws so if conversion fails and it throws ... Seems okay for me.

BillyDonahue:
I really disagree with this idea of lexical casting in the `asT()` accessors.
It hides bugs. If a user received a string that happens to be `"123"`, `asInt()` should not succeed. It will fail when a different string value is received. The receiver had a type error in its code that was masked by the success of `asInt()` there, and such success can't be relied on in the future.

If the receiver is deliberately accepting a `string` value and extracting an `int` from it, let him say so with proper conversion functions at the call sites.

cdunn2001:
Is there a way to provide these as wrappers, in a separate header? I agree with @BillyDonahue.

Issue #481, opened 2016-05-26
When can below code segfault ?


ya1gaurav:
I am meeting segfault in below dereference in json_value.cpp
I am using 0.10.4 version.

```
    if (other.comments_) {
      comments_ = new CommentInfo[numberOfCommentPlacement];
      for (int comment = 0; comment < numberOfCommentPlacement; ++comment) {
        const CommentInfo& otherComment = other.comments_[comment];
        if (otherComment.comment_)
          comments_[comment].setComment(
              otherComment.comment_, strlen(otherComment.comment_));
      }
   }
```

Segfault happend at:
       if (otherComment.comment_)
Similar code in latest branch is at:
https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_value.cpp#L456
When can such scenario occur?
Can anyone create test case for this ?

ya1gaurav:
I do not have sample code. But it crashed is some scenario like:
Json::Value abc = def;
"def" when assigned to abc, it causes json lib to crash when comments are encountered.
How can such situation happen ?

cdunn2001:
@ya1gaurav, please try to produce a failing test-case. I'm not sure whether your fix is on the right track or not.

baylesj:
Closing due to inactivity. If this issue is still relevant, please feel free to comment/reopen/open a new issue as appropriate.

PR #482, opened 2016-05-26
Fix Null/Invalid Pointer Dereference

ya1gaurav:
Can this fix below issue ?
https://github.com/open-source-parsers/jsoncpp/issues/481


ya1gaurav:
Ohh I made mistake in this fix as Object of CommentInfo I can't check to Null as its not pointer type.
But what if it is refer to invalid memory location ?

PR #492, opened 2016-06-28
general compile error and warning fixes

zeromus:
. fix a bunch of missing string and stdio headers (really weird that these crept into the code...?)
. typo amalgated -> amalgamated
. add #if JSON_USE_EXCEPTION guard around exception-throwing code
. change format string "%%.%dg" to "%%.%ug" for an unsigned int

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -18,6 +18,8 @@
 #include <memory>
 #include <set>
 #include <limits>
+#include <stdio.h>
+#include <string.h>
 
 #if defined(_MSC_VER)
 #if !defined(WINCE) && defined(__STDC_SECURE_LIB__) && _MSC_VER >= 1500 // VC++ 9.0 and above 
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -18,6 +18,7 @@
 #endif
 #include <cstddef> // size_t
 #include <algorithm> // min()
+#include <string.h>
 
 #define JSON_ASSERT_UNREACHABLE assert(false)
 
@@ -190,6 +191,7 @@ static inline void releaseStringValue(char* value, unsigned) {
 
 namespace Json {
 
+#if JSON_USE_EXCEPTION
 Exception::Exception(JSONCPP_STRING const& msg)
   : msg_(msg)
 {}
@@ -213,6 +215,7 @@ JSONCPP_NORETURN void throwLogicError(JSONCPP_STRING const& msg)
 {
   throw LogicError(msg);
 }
+#endif
 
 // //////////////////////////////////////////////////////////////////
 // //////////////////////////////////////////////////////////////////
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -15,6 +15,8 @@
 #include <cassert>
 #include <cstring>
 #include <cstdio>
+#include <stdio.h>
+#include <string.h>
 
 #if defined(_MSC_VER) && _MSC_VER >= 1200 && _MSC_VER < 1800 // Between VC++ 6.0 and VC++ 11.0
 #include <float.h>
@@ -42,7 +44,9 @@
 #else
 #include <cmath>
 #if !(defined(__QNXNTO__)) // QNX already defines isfinite
+#ifndef isfinite
 #define isfinite std::isfinite
+#endif
 #endif
 #endif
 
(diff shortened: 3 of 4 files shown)


src/lib_json/json_reader.cpp
```diff
@@ -18,6 +18,8 @@
 #include <memory>
 #include <set>
 #include <limits>
+#include <stdio.h>
```
cdunn2001: Minimally, wouldn't these have to be `cstring` and `cstdio` for strict compatibility with modern compilers?

Adding headers is fine. But could you tell us where this was failing to compile?

src/lib_json/json_writer.cpp
```diff
@@ -42,9 +44,11 @@
 #else
 #include <cmath>
 #if !(defined(__QNXNTO__)) // QNX already defines isfinite
+#ifndef isfinite
```
cdunn2001: Good!

src/lib_json/json_writer.cpp
```diff
@@ -143,7 +147,7 @@ JSONCPP_STRING valueToString(double value, bool useSpecialFloats, unsigned int p
   int len = -1;
 
   char formatString[6];
-  sprintf(formatString, "%%.%dg", precision);
+  sprintf(formatString, "%%.%ug", precision);
```
cdunn2001: Good.

src/lib_json/json_value.cpp
```diff
@@ -190,6 +191,7 @@ static inline void releaseStringValue(char* value, unsigned) {
 
 namespace Json {
 
+#if JSON_USE_EXCEPTION
```
cdunn2001: I don't understand the need for this guard. As long as we don't *throw* an exception, who cares if exceptions are define? Also, I think it's still declared, right?

zeromus: That code contains 'throw'. When exceptions are disabled in the compiler, throw is unavailable.

src/lib_json/json_reader.cpp
zeromus: Beats me. I dont use modern compilers or know anything about them. This was failing on visual c++ 2010.

cdunn2001: I'm not sure we can continue to support VS2010 for the `master` branch. `0.y.z` will work fine.

But could you try `#include <cstdio>` and `#include <cstring>` in VS2010, and if it works, push that to this branch to update this pull-request?

src/lib_json/json_value.cpp
cdunn2001: What if we simply guard the `throw`? We could also `abort()` on the next line. **JsonCpp** will not call these `throw*` functions when `JSON_USE_EXCEPTIONS` is false anyway.

src/lib_json/json_reader.cpp
zeromus: Ahh sorry I was mistaken. It wasn't VS2010, it was a ghetto version of ARMCC I have to use. Each of the files I added stdio.h or string.h to already had `<cstring>` and` <cstdio>` in it, and it wasn't sufficient--so your test is already done.

I noticed this snippet in one place (and snippets like it in many other places)

> #if defined(`__QNXNTO__`)
> #define sscanf std::sscanf
> #endif`

This is representative of another solution to the actual problem I had (symbols like sscanf being undeclared, being in the std:: namespace instead). So, it seems you've dealt with this problem before. 

Since I don't know anything about compilers from the present or the future (it's all the same to a guy stuck in the past) all I can do is naively google up things like http://en.cppreference.com/w/cpp/io/c/sscanf which seems to say those functions belong in std:: namespace indeed and assume your great modern compilers which support them without the std:: qualifier are leading you astray

I'll let you think on this a while before I do any reverting. I can't competently suggest anything, but can incompetently suggest you sprinkle the code with std:: liberally if you want to be standards-compliant with c++29 or whatever you're targeting. For instance, I fixed json_writer.cpp in a different way by adding six std::

cdunn2001: > your great modern compilers which support them without the std:: qualifier are leading you astray

Some older compilers have problems with `std::` sometimes, and `sscanf` in particular can be available in weird ways. We'll be standards compliant in `2.0`, but for now it's helpful to some users if we avoid being too strict.

> I'll let you think on this a while before I do any reverting.

Don't revert. But can you wrap the new include lines in and `ifdef` for your compiler? IIRC, some modern compilers actually do not include the old headers, or consider them a compiler warning (which becomes an error with `-Werror`).

> For instance, I fixed json_writer.cpp in a different way by adding six std::

Could you show us where? We can't add that on `0.y.z`, but on `master` I think that would be fine.

cdunn2001:
Not sure about this anymore. Rebase and reopen if important.

Issue #279, opened 2015-05-19
pkg-config file contains incorrect path when INCLUDE_INSTALL_DIR is overriden


mgorny:
Long story short:

```
$ cmake .. -DJSONCPP_WITH_PKGCONFIG_SUPPORT=1 
-- JsonCpp Version: 1.6.2
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/jsoncpp-1.6.2/_build
$ grep includedir= pkg-config/jsoncpp.pc 
includedir=${prefix}/include
```

Correct so far. However, if I change `INCLUDE_INSTALL_DIR` to another name:

```
$ cmake .. -DJSONCPP_WITH_PKGCONFIG_SUPPORT=1 -DINCLUDE_INSTALL_DIR=test
-- JsonCpp Version: 1.6.2
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/jsoncpp-1.6.2/_build
$ grep includedir= pkg-config/jsoncpp.pc 
includedir=${prefix}//tmp/jsoncpp-1.6.2/_build/test
```

An absolute path with build directory is put there, after the prefix. And if I pass an absolute path:

```
$ cmake .. -DJSONCPP_WITH_PKGCONFIG_SUPPORT=1 -DINCLUDE_INSTALL_DIR=/usr/include/jsoncpp
-- JsonCpp Version: 1.6.2
-- Configuring done
-- Generating done
-- Build files have been written to: /tmp/jsoncpp-1.6.2/_build
$ grep includedir= pkg-config/jsoncpp.pc 
includedir=${prefix}//usr/include/jsoncpp
```

Then the prefix is duplicated.

dreifachstein:
I cannot reproduce your problem with cmake 3.4, and it seems to be a cmake bug in your case.
And your fix does not interact well with CMAKE_STAGING_PREFIX.
Suppose one is compiling with CMAKE_INSTALL_PREFIX=/usr/local and CMAKE_STAGING_PREFIX=/tmp/jsoncpp. The include dir is now installed into /tmp/jsoncpp/usr/local/include instead of /tmp/jsoncpp/include, which is inconsistent with the cmake documentation. Also a windows build is broken if both CMAKE_INSTALL_PREFIX and CMAKE_STAGING_PREFIX are set to some absolute paths.

cdunn2001:
[...]

PR #493, opened 2016-07-07
Fix compilation errors for downstream projects caused by incorrect pkconfig paths

zorun:
Recent commit 911e2b0fea ("By default use \<prefix\> relative paths when
installing") introduced relative install paths in CMake.  But this
interacts badly with commit e6f1cffdd3 from a year ago: now, the paths in
`pkgconfig/jsoncpp.pc` are relative, which is incorrect.

Before 911e2b0fea (1.7.2 on Archlinux), this was correct:

    $ head -4 /usr/lib/pkgconfig/jsoncpp.pc
    prefix=/usr
    exec_prefix=${prefix}
    libdir=/usr/lib
    includedir=/usr/include

After 911e2b0fea (1.7.3 on Archlinux), this is now incorrect:

    $ head -4 /usr/lib/pkgconfig/jsoncpp.pc
    prefix=/usr
    exec_prefix=${prefix}
    libdir=lib
    includedir=include

This change causes hard-to-debug compilation errors for projects that
depend on jsoncpp, for instance:

    CXXLD    libring.la
    /tmp/ring-daemon/src/ring-daemon/src/../libtool: line 7486: cd: lib: No such file or directory
    libtool:   error: cannot determine absolute directory name of 'lib'
    make[3]: *** [Makefile:679: libring.la] Error 1

This is because jsoncpp contributes `-Llib -ljsoncpp` to the LDFLAGS, via
the pkg-config machinery.  Notice the relative path in `-Llib`.

To fix this, simply revert commit e6f1cffdd3 ("Fix custom includedir &
libdir substitution in pkg-config").  The change in 911e2b0fea should have
the same effect.

See #279, #470 for references.

### DIFF
--- a/pkg-config/jsoncpp.pc.in
+++ b/pkg-config/jsoncpp.pc.in
@@ -1,7 +1,7 @@
 prefix=@CMAKE_INSTALL_PREFIX@
 exec_prefix=${prefix}
-libdir=@LIBRARY_INSTALL_DIR@
-includedir=@INCLUDE_INSTALL_DIR@
+libdir=${exec_prefix}/@LIBRARY_INSTALL_DIR@
+includedir=${prefix}/@INCLUDE_INSTALL_DIR@
 
 Name: jsoncpp
 Description: A C++ library for interacting with JSON


cdunn2001:
Thanks!

zorun:
Great, thanks for the merge!

Would you mind spinning up a new release with this fix?

cdunn2001:
Done:
* https://github.com/open-source-parsers/jsoncpp/releases/tag/1.7.4

A "pre-release" for now. If anybody reports a problem, we might need your help to address it.

cdunn2001:
Could you comment on #497? We might need to revert.

PR #508, opened 2016-07-27
Change Value::append() to take its arg by value, then std::move it to its final destination.

kosak:
As things are, the comment ```Equivalent to jsonvalue[jsonvalue.size()] = value;``` is not true because ```append()``` will sometimes make an extra copy.

Example: ```v.append("hello")``` is more expensive than ```v[v.size()] = "hello"```

By taking its argument by value and then ```std::move```ing that arg to the right place, we can actually make the two forms equivalent.

Tested: passes all existing tests.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -434,7 +434,7 @@ Json::Value obj_value(Json::objectValue); // {}
   /// \brief Append value to array at the end.
   ///
   /// Equivalent to jsonvalue[jsonvalue.size()] = value;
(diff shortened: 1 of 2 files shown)


cdunn2001:
True, not quite "equivalent". Could you fix the comment?

We cannot accept this pull-request because it would not maintain binary-compatibility. When we get around to another major-version bump, hopefully we will remember this change. If you want, we could start a `2.y.z` branch.

PR #521, opened 2016-08-22
Con 87 01 branch

DaveSmithPT:
Some fixes for problems found by the Coverity static analysis tool. These changes are based on the 0.10.6 tag, and have been tagged 0.10.7.


cdunn2001:
This needs to be based on either the `0.y.z` branch, or at least on the commit you started with. Instead, this is based on `master`.

I think GitHub recently added support for switching the basis branch, but if you can't work that out, then close and submit a new PR.

cdunn2001:
Oh, I think I'm wrong. GitHub now lets you change the source branch, but not the basis branch. I think you need to submit a new PR. You can use the drop-down to select a different base branch.

cdunn2001:
@nicmcd, you can file a separate issue for `auto_ptr`, so we don't forget. Or submit a PR. That can be changed easily.

@DaveSmithPT, I'm chary of changes to `0.y.z` since we *must* maintain back-compatibility there. I think it's better to live with Coverity warnings for that old branch.

PR #523, opened 2016-08-26
Workaround for missing lconv::decimal_point on android

gidapataki:
On android, `struct lconv` is defined as an empty struct, so accessing `decimal_point` is a compiler error.
This PR adds a compile-time check, and `lc->decimal_point` is only compiled if we have a reason to believe it's there.


cdunn2001:
Sorry for delay. Very helpful change. Thank you.

Issue #527, opened 2016-09-05
localeconv() workaround doesn't help on Android


DrMetallius:
The workaround for missing locale support introduced in 894e78bff13b7e9ba69051904c6b37c197a6cb66 doesn't seem to be working. I compile with NDK r12b, with clang as the compiler, `c++_shared` as my library, and `android-9` as the platform, and it tells me that it can't build it because of the "undefined reference to 'localeconv'". This comes from `json_tool.h`.

`sizeof(struct lconv)` is 56. So it clearly is larger than a simple stub for whatever reason, that's why everything fails. Honestly, I think that a preprocessor macro would be much simpler and more reliable than this template magic.

cdunn2001:
Agreed. If you submit a macro-based fix, we'll merge it right away.

cdunn2001:
See discussion on 52cfe5ae889adf0c4099b16ddeb4b529b1f09007

Gachapen:
CMake can check if a type exists and if a struct has a member. This should be sufficient. I can create a PR and then @DrMetallius can test it.

See https://cmake.org/cmake/help/v3.0/module/CheckTypeSize.html and https://cmake.org/cmake/help/v3.0/module/CheckStructHasMember.html

DrMetallius:
There is [a problem with CMake](https://code.google.com/p/android/issues/detail?id=221907) in CI environments for Android developers. Besides, I'm pretty sure that in our case the structure wasn't empty since it had the size more than zero, but it didn't work regardless. Checking anything about the structure just doesn't seem reliable to me.

Gachapen:
So you're approach to CMake on Android is quite different than ours. We use the system CMake with the toolchain file provided by the NDK. So in our docker build image CMake is installed as an apt package, while the NDK and SDK are downloaded. If you're interested, see https://gitlab.com/suttungdigital/nox-engine/blob/master/tools/docker/build-ubuntu-latest-all/Dockerfile.

From what I can see from your issue description, your problem was that `localeconv` was undefined. What if we, in addition to the aforementioned checks, also check if the `localeconv` symbol exists using https://cmake.org/cmake/help/v3.0/module/CheckSymbolExists.html?

If we check if `lconv` exists, then if `localeconv` is defined, and finally if `lconv` has the member `decimal_point`, we should cover everything used.

DrMetallius:
[...]

PR #528, opened 2016-09-06
Used macros to disable localeconv() calls

DrMetallius:
Fixes issue #527. Some NDK platforms don't have locale support, so there should be a reliable way to disable the support for locales. Since it isn't possible to determine the current target platform, the best way to do this is using macros.

### DIFF
--- a/src/lib_json/json_tool.h
+++ b/src/lib_json/json_tool.h
@@ -5,7 +5,10 @@
 
 #ifndef LIB_JSONCPP_JSON_TOOL_H_INCLUDED
 #define LIB_JSONCPP_JSON_TOOL_H_INCLUDED
+
+#ifndef NO_LOCALE_SUPPORT
 #include <clocale>
+#endif
 
 /* This header provides common string manipulation support, such as UTF-8,
  * portable conversion from/to string...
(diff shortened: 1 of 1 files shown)


cdunn2001:
I'll tell you what drives me crazy: I don't even think that a JSON library should deal with this sort of thing. None of the well-known JSON parsers do this the way I consider the right way, which is to let the user handle conversion from sub-string to value. That could be via template arguments, plugins, callbacks, wrappers, or something else. It's obviously better to me, but users have never clamored for it.

Anyway, this seems like a good change to me. Hopefully this will solve the remaining problems. Thanks!

DrMetallius:
I also found it weird that the JSON library handles issues like these. If you want to write out a number as a string instead of an actual number, I think that trying to forecast the way it should appear is not something a library should do anyway.

It is not my call to make any changes to the logic though, so I did the minimal changes required to make sure the library compiles for older native Android platforms.

cdunn2001:
That's not what I'm talking about. I don't know why anyone would convert from a number to a string, but that also doesn't hurt anything. That's only for convenience.

I'm talking about serialization, which is the production of the JSON string. The JSON spec does not say anything about how numbers are represented. For example, JSON does not limit the number of digits of an integer or the size of a mantissa of a floating point number. That's why it's silly for the library to take these decisions. The parser should only parse for numeric boundaries; something else should convert to an internal representation. Writing should be even simpler.

DrMetallius:
I actually submitted the bug about compiling for Android exactly because I needed type coercion which appeared in a later library version than was used in the project. Our server guys changed the data type of a JSON value (maybe accidentally because they use Ruby), and when an integer was parsed as a string, the whole application crashed. So I decided to update to the latest library version which had coercion.

Regarding the numbers, the decimal point is definitely specified as the '.' symbol. The number of digits is indeed not specified, which means that it can be very large. In the light of that I think that it is even more important to have the ability to read the numbers as strings. Say, I have a number which can't fit even in a 64-bit variable. In that case I would read it as a string and then use a big number class which can construct itself from a string. If I couldn't read this number from JSON as a string, that would mean that it's impossible to read it at all.

cdunn2001:
I think I misunderstood your comment before. Yes, I agree.

cdunn2001:
> In the light of that I think that it is even more important to have the ability to read the numbers as strings.

You might like the **gason** library, which I admire for its simplicity. It produced linked lists instead of dictionaries, but it is very efficient without the low-level optimizations that you can find in rapidjson, e.g.

DrMetallius:
It is pretty remarkable, that's true. I wouldn't use it in our project though, type coercion isn't supported there, and while being fast is nice, correctness is much more important for us.

PR #530, opened 2016-09-08
Added support for omitting the last newline in the StyledWriter

gleerman:
I came across a use case where it is not desirable to have a newline at the end when using the StyledWriter. Similarly to the FastWriter, I added the option to omit that newline. This is disabled by default, hence it should not break any existing usage of the StyledWriter.

### DIFF
--- a/include/json/writer.h
+++ b/include/json/writer.h
@@ -219,6 +219,8 @@ public: // overridden from Writer
    */
   JSONCPP_STRING write(const Value& root) JSONCPP_OVERRIDE;
 
+  void omitEndingLineFeed();
+
 private:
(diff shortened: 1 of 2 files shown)


gleerman:
Closing the pull request because in hindsight I could have used the FastWriter instead.

cdunn2001:
FastWriter and StyledWriter are deprecated. They exist only for backwards compatibility.

Please try the newer `StreamWriterBuilder`, which is more configurable.

cdunn2001:
Also, note that the change as submitted would alter the `sizeof` of public struct, which is not binary-compatible. That's the main reason for our switch to Builders.

PR #543, opened 2016-10-14
Rename variable empty to emptyString

chfast:
Rename variable empty to emptyString in Value constructor to avoid shadowing of Value::empty().

GCC 4.8 produces the warning about this:
lib_json/json_value.cpp: In constructor ‘Json::Value::Value(Json::ValueType)’:
lib_json/json_value.cpp:346:27: warning: declaration of ‘empty’ shadows a member of 'this' [-Wshadow]


cdunn2001:
Thanks!

PR #549, opened 2016-11-02
Added stack overflow test

jia3ep:
Added a test which tries to provoke stack overflow using the syntactically correct JSON file.


cdunn2001:
I think the global stack-limit applies only to the old, deprecated `Reader`, not to the new `CharReaderBuilder`. But I guess it doesn't hurt to prove that the limit/depth is working, if only you can get TravisCI to pass.

jia3ep:
Yes, global variable is only used in Reader. Still I'll figure out why clang fails to compile in TravisCI.

PR #551, opened 2016-11-07
Check for locale support in CMake

Gachapen:
This should improve #527 by checking for locale support at configure time.

Example output from our Android build with this change:
```
-- Looking for C++ include clocale
-- Looking for C++ include clocale - found
-- Check size of lconv
-- Check size of lconv - done
-- Performing Test HAVE_DECIMAL_POINT
-- Performing Test HAVE_DECIMAL_POINT - Failed
-- Looking for localeconv
-- Looking for localeconv - not found
CMake Warning at third_party/jsoncpp/src/lib_json/CMakeLists.txt:25 (message):
  Locale functionality is not supported
```

I would also like to suggest changing `NO_LOCALE_SUPPORT` to `JSONCPP_NO_LOCALE_SUPPORT` as to not clutter the global namespace. But this would break builds relying on `NO_LOCALE_SUPPORT`, so I didn't implement it.

### DIFF
--- a/src/lib_json/CMakeLists.txt
+++ b/src/lib_json/CMakeLists.txt
@@ -9,6 +9,34 @@ if( CMAKE_COMPILER_IS_GNUCXX )
   endif()
 endif( CMAKE_COMPILER_IS_GNUCXX )
 
+include(CheckIncludeFileCXX)
+include(CheckTypeSize)
+include(CheckStructHasMember)
+include(CheckCXXSymbolExists)
+
+check_include_file_cxx(clocale HAVE_CLOCALE)
+check_cxx_symbol_exists(localeconv clocale HAVE_LOCALECONV)
+
+if(CMAKE_VERSION VERSION_LESS 3.0.0)
+	# The "LANGUAGE CXX" parameter is not supported in CMake versions below 3,
+	# so the C compiler and header has to be used.
+	check_include_file(locale.h HAVE_LOCALE_H)
+	set(CMAKE_EXTRA_INCLUDE_FILES locale.h)
+	check_type_size("struct lconv" LCONV_SIZE)
+	unset(CMAKE_EXTRA_INCLUDE_FILES)
+	check_struct_has_member("struct lconv" decimal_point locale.h HAVE_DECIMAL_POINT)
+else()
+	set(CMAKE_EXTRA_INCLUDE_FILES clocale)
+	check_type_size(lconv LCONV_SIZE LANGUAGE CXX)
+	unset(CMAKE_EXTRA_INCLUDE_FILES)
+	check_struct_has_member(lconv decimal_point clocale HAVE_DECIMAL_POINT LANGUAGE CXX)
+endif()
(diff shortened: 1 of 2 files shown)


src/lib_json/CMakeLists.txt
```diff
@@ -9,6 +9,23 @@ if( CMAKE_COMPILER_IS_GNUCXX )
...
+include(CheckTypeSize)
+include(CheckStructHasMember)
+include(CheckCXXSymbolExists)
+
+check_include_file_cxx(clocale HAVE_CLOCALE)
+set(CMAKE_EXTRA_INCLUDE_FILES clocale)
+check_type_size(lconv LCONV_SIZE LANGUAGE CXX)
+set(CMAKE_EXTRA_INCLUDE_FILES "")
+check_struct_has_member(lconv decimal_point clocale HAVE_DECIMAL_POINT LANGUAGE CXX)
+check_cxx_symbol_exists(localeconv clocale HAVE_LOCALECONV)
```
Gachapen: Frustrating how inconsistent the CMake API is with regards to setting language and adding headers.

Gachapen:
I see both gcc and clang in travis fail `HAVE_DECIMAL_POINT`, I should look into this. It did not fail for me with clang 3.9. Travis is using clang 3.4.

Gachapen:
I nailed it down to a change in CMake 3.0 where they added the `LANGUAGE` option to `check_struct_has_member`. Before this change, it always compiled with the C compiler, not C++.

I see some options, but not sure which is best:
- Upgrade the Travis CMake to >=3.0 and drop support for older versions.
- Change the check to use the `locale.h` header instead of `clocale` so that it can compile it with the C compiler. (lconv in the C header and C++ header might not be the same)
- Add the updated `check_struct_has_member` module to the jsoncpp project to fix the check. (updates to the system cmake won't affect the `check_struct_has_member` anymore)

Gachapen:
I implemented the second option as it seems to be the least problematic. Now, I misread the build log. It wasn't only `HAVE_DECIMAL_POINT` that failed, but also the `lconv` size check. Possibly because of the same reason. I'll investigate more....

Gachapen:
I amended my previous commit with a fix for the problem. It's not pretty, but it's the best I could do if we want to support older CMake versions.

cdunn2001:
> I would also like to suggest changing NO_LOCALE_SUPPORT to JSONCPP_NO_LOCALE_SUPPORT as to not clutter the global namespace. But this would break builds relying on NO_LOCALE_SUPPORT, so I didn't implement it.

Could you include that amendment in this PR? Don't worry too much about build-breakage. But if you fallback to `NO_LOCALE_SUPPORT` when `JSONCPP_NO_LOCALE_SUPPORT` is undefined, that might be good for now.

Gachapen:
I squashed my two first commits, and added a commit for `JSONCPP_NO_LOCALE_SUPPORT`. I think it makes more sense to separate them.

PR #552, opened 2016-11-07
change throw() to noexcept to conform to c++11

omki2005:
throw() is deprecated in c++11. Should probably move to using noexcept instead

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3051.html

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -83,10 +83,16 @@
 // managable and fixes a set of common hard-to-find bugs.
 #if __cplusplus >= 201103L
 # define JSONCPP_OVERRIDE override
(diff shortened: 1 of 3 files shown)


cdunn2001:
Can we do anything about the AppVeyor build?

omki2005:
Done but I don't have MSVCC. Does anyone know what _MSC_VER will be for
Visual Studio 2015? Because that's when noexcept starts to be supported.

https://msdn.microsoft.com/en-us/library/hh567368.aspx

I guess it's 1900 from the wiki page

PR #556, opened 2016-11-30
Removed a static variable used to contain the current recursion depth in json_reader.cpp

nnkur:
Removed a static variable used to contain the current recursion depth of Reader::readValue().  The number of elements in an internal container Reader::nodes_  is used instead.  It is correct because any recursive call of Reader::readValue() is executed with adjacent nodes_.push()  and nodes_.pop() calls.  
Added the option to change the allowed recursion depth at compile time by defining a macro JSONCPP_DEPRECATED_STACK_LIMIT as the required integer value.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -45,12 +45,12 @@
 #pragma warning(disable : 4996)
 #endif
 
-// Define JSONCPP_STACK_LIMIT as an appropriate integer at compile time to change the stack limit
-#if !defined(JSONCPP_STACK_LIMIT)
-#define JSONCPP_STACK_LIMIT 1000
+// Define JSONCPP_DEPRECATED_STACK_LIMIT as an appropriate integer at compile time to change the stack limit
+#if !defined(JSONCPP_DEPRECATED_STACK_LIMIT)
+#define JSONCPP_DEPRECATED_STACK_LIMIT 1000
 #endif
 
-static size_t const stackLimit_g = JSONCPP_STACK_LIMIT; // see readValue()
+static size_t const stackLimit_g = JSONCPP_DEPRECATED_STACK_LIMIT; // see readValue()
 
 namespace Json {
 
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1,4 +1,5 @@
 // Copyright 2007-2011 Baptiste Lepilleur
+// Copyright (C) 2016 InfoTeCS JSC. All rights reserved.
 // Distributed under MIT license, or public domain if desired and
 // recognized in your jurisdiction.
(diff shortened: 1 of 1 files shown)


src/lib_json/json_reader.cpp
```diff
@@ -1096,8 +1090,8 @@ bool OurReader::parse(const char* beginDoc,
 }
 
 bool OurReader::readValue() {
-  if (stackDepth_ >= features_.stackLimit_) throwRuntimeError("Exceeded stackLimit in readValue().");
-  ++stackDepth_;
+  //  To preserve the old behaviour we cast size_t to int.
+  if (static_cast<int>(nodes_.size()) > features_.stackLimit_) throwRuntimeError("Exceeded stackLimit in readValue().");
```
cdunn2001: Isn't `stack<>.size()` an O(n) operation?

I don't understand what was wrong with `OurReader`. Could you explain?

src/lib_json/json_reader.cpp
```diff
@@ -44,13 +45,13 @@
 #pragma warning(disable : 4996)
 #endif
 
-static int const stackLimit_g = 1000;
-#if __cplusplus >= 201103L
-thread_local static int stackDepth_g = 0;  // see readValue()
-#else
-static int       stackDepth_g = 0;  // see readValue()
+// Define JSONCPP_STACK_LIMIT as an appropriate integer at compile time to change the stack limit
+#if !defined(JSONCPP_STACK_LIMIT)
```
cdunn2001: I think this should be `JSONCPP_DEPRECATED_STACK_LIMIT`, since this should apply only to the old deprecated `Reader`, which nobody should be using anyway.

nnkur: About JSONCPP_DEPRECATED_STACK_LIMIT: Ok. I will make the reqiured changes.
Actually there are some product, which used jsoncpp in the old manner. Replacing a dll is relatively simple, changing all modules may be considerable harder.

src/lib_json/json_reader.cpp
nnkur: stack<>.size() complexity is constant: stack is deque (http://www.cplusplus.com/reference/stack/stack/) and deque<>.size complexity - http://www.cplusplus.com/reference/deque/deque/size/

About OurReader. I have change it because of the following comment to the class:
// exact copy of Reader, renamed to OurReader
and 
// complete copy of Read impl, for OurReader

If you allow some small differences I would rollback my changes to OurReader.

cdunn2001: Yes, it is ok for OurReader to differ.

jia3ep: @cdunn2001 it have no sense to keep stackDepth_ in OurReader. New solution is use less additional memory (it's one member but it could be important on embedded devices) and still has the same computational complexity. And the tests are OK.

Issue #560, opened 2016-12-09
Use GNUInstallDirs instead of defining own variables


SoapGentoo:
Dear JsonCpp devs,
in Gentoo we like to standardise on certain features. Autoconf+Automake can make things a lot easier for us, as automake provides a number of standard paths that can be changed. While you have a similiar scheme for your CMake, none of that is standardised. Would you consider replacing your custom variables with the `GNUInstallDirs` (https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html), provided with CMake? This provides more targets and is more idiomatic and makes it easier for downstream users to use the same conventions across packages, instead of packages reinventing this (convention over configuration). This would simplify a lot of things for us.

If you would be willing to accept such a change, I will gladly provide a Pull Request.

cdunn2001:
Yes, we welcome contributions!

SoapGentoo:
@cdunn2001 Thanks for the cooperative atmosphere! :smile:

PR #562, opened 2016-12-10
Replace current install variables with GNUInstallDirs

SoapGentoo:
* The GNUInstallDirs module is more idiomatic and supported by
  Kitware upstream, whereas the current directories are not
  standardised across CMake-using packages. Using CMake native
  mechanisms is better than reinventing the wheel, as it makes
  using the build system more uniform across the ecosystem
* Use CMAKE_CXX_STANDARD to force C++11
* Require CMake 3.1.0 at a minimum
* Fixed lower/UPPERcase format for function/macro calls
* Fixed indents by replacing tabs with 4 spaces

This PR is in response to https://github.com/open-source-parsers/jsoncpp/issues/560


CMakeLists.txt
```diff
@@ -86,65 +76,71 @@ CONFIGURE_FILE( "${PROJECT_SOURCE_DIR}/version.in"
...
-if(CCACHE_FOUND)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE ccache)
-    set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK ccache)
-endif(CCACHE_FOUND)
+    SET(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /W4 ")
+ENDIF()
+
+# Require C++11 support, prefer ISO C++ over GNU variants,
+# as relying solely on ISO C++ is more portable.
+SET(CMAKE_CXX_STANDARD 11)
```
strahlc: this feature requires at least cmake 3.1:

You should change line 3 to:
cmake_minimum_required(VERSION 3.1)

SoapGentoo: Done!

cdunn2001:
Can't merge if Travis fails. Maybe there is a way to load the right cmake?

SoapGentoo:
@cdunn2001 I've fixed the CMake error by switching to a newer CMake version, using https://github.com/travis-ci/travis-ci/issues/4631#issuecomment-191153634
Would you consider merging now? :smile:

cdunn2001:
This looks like a lot of work. Thanks!

cdunn2001:
I have this error:
```
$ cmake -DCMAKE_BUILD_TYPE=debug -DBUILD_STATIC_LIBS=ON -DBUILD_SHARED_LIBS=OFF -DARCHIVE_INSTALL_DIR=. -G "Unix Makefiles" ../..
CMake Error at CMakeLists.txt:4 (PROJECT):
  The CMAKE_C_COMPILER:
    /Library/Developer/CommandLineTools/usr/bin/cc

  is not a full path to an existing compiler tool.
```
I'll try setting `CC` in my environment.  I guess we need to update the README. Would you do that?

cdunn2001:
Oh, it worked after removing the cmake cache. (Actually, I removed the whole build directory and started over.) No problems!

cdunn2001:
https://github.com/open-source-parsers/jsoncpp/tree/1.8.0

PR #569, opened 2017-01-04
Update config.h

Saxin:
missing ".h" 
should be #include <string.h>


cdunn2001:
Sorry for long delay. Why `<string.h>` instead of `<string>`? What symbol were we missing.

cdunn2001:
Can we close this?

okodron:
I think, yes.

PR #571, opened 2017-01-12
README: Give some love

ibc:
It was so hard for to find the documentation link in the previous README. This PR makes it easier and improves a bit the whole documentation syntax.


ibc:
No even a comment?

cdunn2001:
I much prefer wiki to README, since there can be a lot of churn. But this seem like good changes. Thank you!

Issue #578, opened 2017-02-11
Reader::parse copies string data with capacity() instead of size()


baustinanki:
Our valgrind tests, some of which use jsoncpp, detected an uninitialized value being used when creating a `std::string` of a certain size and passing it into `Reader::parse`. The `std::string`, which we create with a string literal, has a `size()` of 54 but a `capacity()` of 63. After being passed into `Reader::parse`, this code runs:

```c++
bool
Reader::parse(const std::string& document, Value& root, bool collectComments) {
  JSONCPP_STRING documentCopy(document.data(), document.data() + document.capacity());
  std::swap(documentCopy, document_);
```

This results in `document_.size()` being 63; the `Reader` is now operating on a string that is different than the one we passed in. Our valgrind errors suggest that the extra bytes here are not zero-initialized and are eventually read down the line in `Reader::skipSpaces()`.

I believe the call to `capacity()` in that code above should instead be a call to `size()`.

cinemast:
@cdunn2001 can we fix this? It introduces memory troubles for all reverse-dependencies. Can you fix that?

cdunn2001:
@cinemast, ok. I'll address the comments in #580 and push this weekend.

PR #580, opened 2017-02-15
Fix uninitialized value detected by valgrind

ya1gaurav:
Fix issue reported in https://github.com/open-source-parsers/jsoncpp/issues/578
For std::string variable, length() is more readable than size().

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -103,7 +103,7 @@ Reader::Reader(const Features& features)
 
 bool
 Reader::parse(const std::string& document, Value& root, bool collectComments) {
-  JSONCPP_STRING documentCopy(document.data(), document.data() + document.capacity());
+  JSONCPP_STRING documentCopy(document.data(), document.data() + document.length());
   std::swap(documentCopy, document_);
   const char* begin = document_.c_str();
   const char* end = begin + document_.length();


BillyDonahue:
I disagree about the "readability" of `length` vs `size`, as `size` is part of the container requirements and is more widely used throughout C++ containerdom.  I believe the synonym `length` is a holdover from the pre-STL string class.

src/lib_json/json_reader.cpp
```diff
@@ -103,7 +103,7 @@ Reader::Reader(const Features& features)
 
 bool
 Reader::parse(const std::string& document, Value& root, bool collectComments) {
-  JSONCPP_STRING documentCopy(document.data(), document.data() + document.capacity());
+  JSONCPP_STRING documentCopy(document.data(), document.data() + document.length());
```
BillyDonahue: `JSONCPP_STRING` is going to be some kind of `std::basic_string`. The allocator changes with different build configurations. So it has the full `std::string` constructor API.

    JSONCPP_STRING documentCopy(document.begin(), document.end());

But really we don't need `documentCopy` at all. The swap doesn't seem to be doing anything we can't do better with an assignment.

    document_.assign(document.begin(), document.end());

cdunn2001: I agree with @BillyDonahue.

PR #582, opened 2017-02-20
Use full CMake paths in pkg-config template

SoapGentoo:
Using full paths is more versatile. The current solution
breaks when specifying an absolute path for `CMAKE_INSTALL_INCLUDEDIR`
which is an otherwise supported option by CMake's `GNUInstallDirs`.


SoapGentoo:
@cdunn2001 ping?

cdunn2001:
I thought there was a reason why someone used these prefix paths. Well, I'll merge this, and if it causes someone else a problem, they can debate it.

SoapGentoo:
the `${prefix}` stuff was my an attempt to appear mroe autotools-y, but it only makes sense if the build system actually uses `${prefix}`

Issue #586, opened 2017-03-04
cmake throws warning treated as error message when executed inside TravisCI


fire-d:
I'm trying to use jsoncpp as part of my project. The end goal is to get the following files:

- dist/jsoncpp.cpp (need to run python amalgamate.py)
- dist/json/json.h (need to run python amalgamate.py)
- dist/json/json-forwards.h (need to run python amalgamate.py)
- src/lib_json/libjsoncpp.a (need to run cmake)

So, I have a .travis.yml file inside my repo (which will always run whenever I push to github) that looks like this:

`---`
`language: java`
`sudo: true`
`jdk: oraclejdk8`
`addons:`
`  apt:`
`    sources:`
`      - ubuntu-toolchain-r-test`
`      - llvm-toolchain-precise-3.5`
`      - george-edison55-precise-backports # cmake 3.2.3 / doxygen 1.8.3`
`    packages:`
`      - cmake`
`      - cmake-data`
`      - gcc-4.9`
`      - g++-4.9`
`      - clang-3.5`
`      - valgrind`
`before_install:`
`  - sudo apt-get -qq update`
`  - git clone https://github.com/open-source-parsers/jsoncpp.git`
`  - cd jsoncpp`
`  - python amalgamate.py`
`  - mkdir build`
`  - cd build`
`  - cmake ..`
`  - make`
`script:`
`  - cp src/lib_json/libjsoncpp.a /usr/local/lib/`

The yaml runs perfectly up until the make statement (so I'm able to retrieve the first three files I need), but then it logs the following message:

`$ make`
`Scanning dependencies of target jsoncpp_lib_static`
`[ 16%] Building CXX object src/lib_json/CMakeFiles/jsoncpp_lib_static.dir/json_reader.cpp.o`
`[ 33%] Building CXX object src/lib_json/CMakeFiles/jsoncpp_lib_static.dir/json_value.cpp.o`
`/home/travis/build/fire-d/testDemo/jsoncpp/src/lib_json/json_value.cpp: In copy constructor Json::Value::CZString::CZString(const Json::Value::CZString&)’:`
`/home/travis/build/fire-d/testDemo/jsoncpp/src/lib_json/json_value.cpp:273:78: error: conversion to unsigned char:2’ from ‘unsigned int’ may alter its value [-Werror=conversion]`
`cc1plus: some warnings being treated as errors`
[...]

PR #589, opened 2017-03-10
Fix warning issue with gcc flags.

ya1gaurav:
PR for - https://github.com/open-source-parsers/jsoncpp/issues/586
Separating the default options for compiler flags.


cdunn2001:
I don't understand the code existing *before* this PR. `AND NOT JSONCPP_WITH_WARNING_AS ERROR`? Seems like that was backward. I'll merge, but we'll revert if anybody complains.

(I've been traveling again, and out sick. Sorry for long delay.)

PR #591, opened 2017-03-16
C++11 : Use using instead of typedef

ya1gaurav:
C++11 type alias.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -55,7 +55,7 @@ static size_t const stackLimit_g = JSONCPP_DEPRECATED_STACK_LIMIT; // see readVa
 namespace Json {
 
 #if __cplusplus >= 201103L || (defined(_CPPLIB_VER) && _CPPLIB_VER >= 520)
-using CharReaderPtr =  std::unique_ptr<CharReader>;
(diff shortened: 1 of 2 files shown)


src/lib_json/json_reader.cpp
```diff
@@ -55,9 +55,9 @@ static size_t const stackLimit_g = JSONCPP_DEPRECATED_STACK_LIMIT; // see readVa
 namespace Json {
 
 #if __cplusplus >= 201103L || (defined(_CPPLIB_VER) && _CPPLIB_VER >= 520)
-typedef std::unique_ptr<CharReader> CharReaderPtr;
+using CharReaderPtr =  std::unique_ptr<CharReader>;
```
yiqiju: I am sorry, the extra space before "std..." hurt my eyes...

ya1gaurav:
@yiqiju Thanks, fixed.
@All members, please review. If it seems fine, I will extend changes to header files also.

cdunn2001:
I have no strong objection, but could you provide some justification for this change?

ya1gaurav:
It is just coding style according to new c++ standards.
http://en.cppreference.com/w/cpp/language/type_alias

hjmjohnson:
these style changes have already been incorporated upstream.

Issue #595, opened 2017-03-30
Reader Parser is not able handle escape sequences in input string


hirenpitroda:
**Test Case-1**
```
String Input :- "abc\b"
Expected Output  from reader parser :- "abc\\b"
Observed behavior on JSONCPP :- "abc^H"
OUTPUT After Above Code changes :- "abc\\b"
```
**Test Case-2**
```
String Input :- "abc\bd"
Expected Output  from reader parser :- "abc\\bd"
Observed behavior on JSONCPP :- "abd"
```
As per reader parser API, it should handle escape sequence as it is coming. but reader parser is not behaving same as required above. Please share your thoughts.

cdunn2001:
You closed #596 a day after posting this, yes? Did that not solve your problem? Did it contradict the JSON spec? May we close this?

hirenpitroda:
No, i am working on PATCH for this issue, i will submit a PATCH for this. wait for 2 days.

hirenpitroda:
@cdunn2001 ..
Added PATCH, with some test cases. Have a look.
Provide Comments.

cdunn2001:
I am not convinced. I agree that it would be nice if a JSON library gave the user the ability to determine the internal representation, but that's just not how jsoncpp works. This library converts the escaped characters to 8-bit characters internally. That's just how it works. `\b` is converted to the ASCII "Bell", code 7. In theory, the writer will convert that back to `\b`. It's only the round-trip that matters. JSON.org does not specify the *parsed* representation.

But I'm confused by your examples. Why would JSON `"\b"` become `"\\b"` internally? Are you looking at Python? I'm very confused.

hirenpitroda:
give me some time i will provide my exact requirement. i identified that my changes will lead to some other things to breakdown. so i am working to provider a API for my problem statement.

hirenpitroda:
After Recent work with JSON CPP ,after reading JSON spec and doing some internal UT, found that. its client's(browser) responsibility to handle input Payload, i.e., if user is giving input payload as {"password":"abc\b"} than its browser or client tool responsibility to handle proper escaping and provide same input below layer APIs to work.
Closing this issue.

PR #598, opened 2017-04-06
Issue #595:- Reader Parser to handle escape chars.

hirenpitroda:
re: #595 

Referred from, JSON.org: "A string is a sequence of zero or more Unicode characters, wrapped in double quotes, using backslash escapes. A character is represented as a single character string. A string is very much like a C or Java string."
```
Test Case-1
String Input :- "abc\b"
Expected Output  from reader parser :- "abc\\b"
Observed behavior on JSONCPP :- "abc^H"
OUTPUT After Above Code changes :- "abc\\b"

Test Case-2
String Input :- "abc\bd"
Expected Output  from reader parser :- "abc\\bd"
Observed behavior on JSONCPP :- "abd"
OUTPUT After Above Code changes :- "abc\\bd"
```

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1653,6 +1653,7 @@ bool OurReader::decodeString(Token& token, JSONCPP_STRING& decoded) {
     else if (c == '\\') {
       if (current == end)
         return addError("Empty escape sequence in string", token, current);
+      decoded += '\\';
       Char escape = *current++;
       switch (escape) {
       case '"':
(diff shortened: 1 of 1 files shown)


cdunn2001:
I think your interpretation is wrong. I've just re-read the doc at json.org. `\b` should become the "backslash" character, which you've printed as `^H`. That's correct already.

However, you 2nd example could be a big. You say `"abc\bd"` leads to `"abd"`? Are you sure? If so, please re-open and add a failing test-case, and a suggestion to fix that (but not Test-case 1).

PR #604, opened 2017-04-21
Use std::unique_ptr correctly across compiler and language versions

AbigailBuccaneer:
Previously, the code tried to avoid a `-Wdeprecated-declarations` warning by avoiding `std::auto_ptr` (since #413). However, the implementation has two issues:

1. It uses `std::scoped_ptr`, which as far as I can tell is a nonexistent type. `std::unique_ptr` is presumably what was meant.
2. It avoids `auto_ptr` whenver we're using GCC6. This means we don't dodge warnings for GCC5 or clang (which also support C++11). It also means that using GCC6 in C++03 mode fails, because `std::unique_ptr` doesn't exist there.

This fixes both of these issues, using `std::unique_ptr` in C++11 (and MSVC2010) and above.

Tested on GCC 5.4.1, GCC 6.2.0 and Clang 5.0.0-svn300863 in both C++03 and C++11 modes.

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -51,6 +51,16 @@
 #define JSON_API
 #endif
 
+#if !defined(JSON_HAS_UNIQUE_PTR)
+#if __cplusplus >= 201103L
+#define JSON_HAS_UNIQUE_PTR (1)
+#elif _MSC_VER >= 1600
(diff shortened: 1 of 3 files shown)


cdunn2001:
If anybody objects, we might have to revert. But it looks good to me.

PR #607, opened 2017-04-24
Refactor authorship information for more technical accuracy.

ssbr:
Google advises its employees to add Google Inc. as an author, but that hasn't
been done yet and would be super inconvenient. So instead I've refactored the
file to refer to "The JsonCpp Authors", which are listed in the AUTHORS file.

The AUTHORS file itself is generated via:

    git log --pretty="%an <%ae>%n%cn <%ce>" | sort | uniq

Plus the addition of "Google Inc." as a copyright author. (Google owns the work
of anyone contributing from an @google.com address, for example.)

The list contains some probable duplicates where people have used more than one
email address. I didn't deduplicate because -- well, who's to say they're
duplicates, anyway? :)

PR #610, opened 2017-05-02
Add a simple fuzz test for jsoncpp.

ssbr:
The idea is to then add it to the [google oss-fuzz project](https://github.com/google/oss-fuzz/) and get jsoncpp regularly tested for bugs.

I've followed the instructions at [Ideal Integration](https://github.com/google/oss-fuzz/blob/master/docs/ideal_integration.md) 

Tested using a simple oss-fuzz setup and with the modified jsoncpp unit tests (`cmake . && make`).


ssbr:
BTW, the only bit of information I need to actually propose it for oss-fuzz is a google-associated email address. See https://github.com/google/oss-fuzz#accepting-new-projects

cdunn2001:
Would my @gmail address suffice? Then you could set this up? If you'd rather use your own email address for convenience, that's fine too. Or use both if possible.

I can't say I completely understand what's going on here, but I don't object.

ssbr:
Yeah, a gmail address works fine. The one you commit with / in AUTHORS, right?

> I can't say I completely understand what's going on here, but I don't object.

The idea is supposed to be that we let oss-fuzz pass in (potentially garbage) data into jsoncpp and see if it triggers accesses to uninitialized memory or other problems that might indicate a security issue. So this test should just accept that data and parse it.

The dumb trick with the hashing to get flags is just to increase the possible code coverage.

src/test_lib_json/main.cpp
```diff
@@ -1,9 +1,10 @@
-// Copyright 2007-2010 Baptiste Lepilleur
```
ssbr: Sorry, missed this because I apparently am having trouble getting updates from github (they go to my personal email instead of work email.)

The previous thing I did was replace with The JsonCpp authors. I actually missed a bunch of files by accident. Want I should redo it for "Baptiste Lepilleur and the JsonCpp Authors?" globally in a separate PR?

https://github.com/open-source-parsers/jsoncpp/pull/607

cdunn2001: Sorry, but I wouldn't feel right approving a commit that removes Baptiste's name from any file.

ssbr: This has been fixed. (idk how github code review works, I guess I have to respond to mark this as done, since it still has the orange blob saying work needs to be done?)

ssbr:
I think I'm confused. I thought this was merged, but I guess it was only "approved". Is there anything that needs to be done on my end? (Sorry, new to git/github...)

cdunn2001:
I don't remember why it wasn't merged back then. Maybe tests were not passing?

Recently we started having problems with TravisCI, which makes a merge even harder. And I want to be sure that the runtime has not gotten worse. But most important, we now use Meson instead of Cmake.

ssbr:
I'm not sure about meson. It looks like the cmake files are still here, and It merged cleanly. Is the plan to get rid of cmake eventually?

I can't set up meson / test any changes there because my version of Ubuntu is too old. (Only has Ninja 1.3, Meson requires Ninja 1.6?)  so I'd prefer to keep using cmake if that's possible, and let you update the meson config.

cdunn2001:
Ok, as long as you're confident in your changes. This is only available to cmake users, and we no longer test cmake in TravisCI. (TravisCI is kinda broken right now anyway, because of python3.)

cdunn2001:
Unmerged. This was failing in TravisCI, and it was merged when TravisCI itself was having problems (a transition to new O/S). This needs to be rebased, and then the tests need to pass.

Dor1s:
Any progress on OSS-Fuzz integration? Re: https://github.com/google/oss-fuzz/pull/575

//cc @kcc

src/test_lib_json/fuzz.cpp
```diff
@@ -0,0 +1,57 @@
...
+}  // namespace
+
+namespace Json {
+class Exception;
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
+  Json::CharReaderBuilder builder;
+
+  uint32_t hash_settings = JenkinsOneAtATimeHash(data, size);
```
zetafunction: Rather than hashing the input, it's simpler to just use the first byte or two for configuring options. We can early return 0 if size < REQUIRED_BYTES_FOR_CONFIG, and the fuzzer will figure out not to follow this edge pretty quickly.

ssbr:
No progress. I've been swamped. (And I also took like a month off, which didn't help.)

PR #616, opened 2017-05-26
Backport 'reject duplicate keys' option to Reader.

jdelfino:
I have a use of jsoncpp that requires access to the numeric spans provided by structured error messages, but also needs the ability to reject duplicate keys. Because StructuredError is defined as an inner class of the private OurCharReader, it is much simpler to backport this option rather than refactor to expose structured errors through CharReader.

### DIFF
--- a/include/json/features.h
+++ b/include/json/features.h
@@ -52,6 +52,9 @@ public:
 
   /// \c true if numeric object key are allowed. Default: \c false.
   bool allowNumericKeys_;
+
+  /// \c true if duplicate keys within an object should be rejected. Default: \c false.
+  bool rejectDupKeys_;
 };
 
 } // namespace Json
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -65,7 +65,8 @@ typedef std::auto_ptr<CharReader>   CharReaderPtr;
 
 Features::Features()
     : allowComments_(true), strictRoot_(false),
-      allowDroppedNullPlaceholders_(false), allowNumericKeys_(false) {}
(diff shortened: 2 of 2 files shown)


include/json/features.h
```diff
@@ -52,6 +52,9 @@ class JSON_API Features {
 
   /// \c true if numeric object key are allowed. Default: \c false.
   bool allowNumericKeys_;
+
+  /// \c true if duplicate keys within an object should be rejected. Default: \c false.
+  bool rejectDupKeys_;
```
cdunn2001: Sorry. We cannot add fields to public classes. That could break binary compatibility.

This is, in fact, the main reason why we create `CharReader` and deprecated `Reader`. With `CharReader`, you can configure these settings via an actual JSON string:
* http://open-source-parsers.github.io/jsoncpp-docs/doxygen/class_json_1_1_char_reader_builder.html
* http://open-source-parsers.github.io/jsoncpp-docs/doxygen/index.html ("Advanced Usage")

jdelfino: Ok, thanks for the explanation.

I need both rejectDupKeys and error offsets, so neither class currently satisfies my needs. Do you expect any upcoming release to be allowed to break binary compatibility? I'm happy to expose structured error info through the CharReader interface (e.g. perhaps as a Json::Value), but don't see how to do that without breaking binary compatibility either.

cdunn2001: Oh, you like the exposed error-offsets. I didn't think that was used for anything except jsoncpp exceptions.

Well, you might be better off with your own fork, unfortunately.

PR #622, opened 2017-06-09
Allocate the proper memory for formatString. Fix a warning with gcc 7.1

sylvestre:
Warning:
```
/root/firefox-gcc-last/toolkit/components/jsoncpp/src/lib_json/json_writer.cpp:139:16: note: using the range [-2147483648, 2147483647] for directive argument
/root/firefox-gcc-last/toolkit/components/jsoncpp/src/lib_json/json_writer.cpp:146:10: note: 'sprintf' output between 5 and 15 bytes into a destination of size 6
   sprintf(formatString, "%%.%dg", precision);
   ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -142,7 +142,7 @@ JSONCPP_STRING valueToString(double value, bool useSpecialFloats, unsigned int p
   char buffer[36];
   int len = -1;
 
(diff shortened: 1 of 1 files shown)


cdunn2001:
Thanks!

Issue #567, opened 2016-12-23
Bug: formatting real values for locales with ',' decimal separator


siddie:
JsonWriter (89ab7ec  on 27 Oct), line 155:
    
`// try to ensure we preserve the fact that this was given to us as a double on input
    if (!strstr(buffer, ".") && !strstr(buffer, "e")) {
      strcat(buffer, ".0");
    }
`

Let
`a = 0.1;`
     this piece of code gives me
`buffer == "0,1.0"`
     then after fixNumericLocale i get 
`buffer == "0.1.0"`

and this value is actually written to json.

The straigttforward way of fixing it is obvious, but I think you should generalize your way of working with locales. 

Anyway, thank you for this good lib.

cdunn2001:
Interesting. Can you submit a PR to fix this? Simple fix first is fine; more complicated fix later.

siddie:
The crucial thing I`ve found in my case is this happening at the beginning of an application I work with:
`std::locale::global(std::locale(""));`
Which of course can cause all sorts of undesired behaviours.

Without this call reading and writing JSON works as expected. Not sure whether this call sets a C locale also. If it does not, probably the bug is caused by mixing of C++ and C formatting stuff in the lib.

bernhardHartleb:
The issue is that certain locales (like german) use a comma as decimal point, which snprintf is happy to produce. To resolve this, fixNumericLocale should be called directly after snprintf, otherwise the next search for a decimal point will fail and another ".0" is appended.

PR #623, opened 2017-06-22
Fix issue #567 in writing real values in different locales

bernhardHartleb:
The output of snprintf might produce ',' separators for decimal places if
certain locales are set. This commit moves the converversion from ',' to '.'
at the correct place. Otherwise an additional ".0" might be appended.


bigcock45:
Commit

PR #630, opened 2017-07-01
Fix shared/static lib build conflict

### LINKED ISSUE #631: fails to build with shared libs enabled on msvc
see:
https://github.com/open-source-parsers/jsoncpp/pull/630

jschueller:
When both static and shared lib build are enabled, there is a conflict between the static and import lib ; both are named jsoncpp.lib and the test was incorrectly linked to the static lib but built with JSON_DLL flag and so failed to import symbols with the dllimport keyword. We propose to fix this by appending a "_static" suffix to the static lib when the shared lib is also enabled, so as by default (shared=0, static=1) the behavior is still the same.
This fixes #631.

Also, the appveyor script was messy and mixed msvc versions and 32/64 bits configurations.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,7 @@ environment:
 build_script:
   - cmake --version
   - cd c:\projects\jsoncpp
-  - cmake -G "%CMAKE_GENERATOR%" -DCMAKE_INSTALL_PREFIX=%CD:\=/%/install -DCMAKE_BUILD_TYPE=Release .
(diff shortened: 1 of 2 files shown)


cdunn2001:
appveyor build failed.

jschueller:
yes, see #631

jschueller:
@cdunn2001 there is a conflict between the static and import lib both named jsoncpp.lib, 24114ff fixes this by appending a _static suffix to the static lib if the shared lib is also enabled.

cdunn2001:
Thanks for the appveyor changes!

Issue #621, opened 2017-06-08
Any plan to add move assignment operator to Json::Value


gitpaladin:
In the master branch, `Json::Value` has move constructor but move assignment operator is missing. So it's very inconvenient to move a existing `Value` object to a parent node. E.g.

```C++
Json::Value child("abc");
Json::Value parent;

// Use swap to void deep copy node
Json::Value &parentSubNodeReference = parent["subNode"];
parentSubNodeReference.swap(child);

// If there is a move assignment operator, it will become very easy
parent["subNode"] = std::move(child);
```

Additionally, add move assignment operator can solve function return value deep copy problem. E.g.
```C++
Json::Value SomeAwesomeFoo()
{
   Json::Value someHugeNode;
   // process some huge node.....omitted
   return someHugeNode;
}

void main()
{
   Json::Value parent;

  // Use swap to prevent deep copy node.
  Json::Value child = SomeAwesomeFoo();  // RVO will prevent deep copy child object
  parent["subNode"].swap(child);

  // This cannot be compiled because swap cannot hold a r-value reference
  // parent["subNode"].swap(SomeAwesomeFoo()); 

  // If there is a move assignment operator, it will become very easy
  // The function return a r-value, the assignment operator will use move assignment 
  // operator to move the temporary object to the subNode
  parent["subNode"] = SomeAwesomeFoo();
}
```

On the other hand, `append` method should also have a overloaded version, which accepts r-value reference of `Json::Value` . So move a temporary object to a array value will come very easy.

```C++
Json::Value child("abc");
Json::Value parent(Json::arrayValue);

// Use swap to prevent deep copy node.
Json::Value &parentSubNodeReference = parent.append(Json::objectValue);
parentSubNodeReference.swap(child);

// Use r-value reference version
parent.append(std::move(child));
```

cdunn2001:
@BillyDonahue, Any opinion on this? Do we need to do a version bump to add move-assignment?

BillyDonahue:
These are all good points. It's surprising to have move construct without move assign. They arise mutually for value types.

Aside, in the above comment this line was offered as a problem:

    parent["subNode"].swap(SomeAwesomeFoo());

But in classical C++ this should be written with a member call on the rvalue, as:

    SomeAwesomeFoo().swap(parent["subNode"]);

[...]

PR #635, opened 2017-07-08
Add move assignment operator for Json::Value class and overload append member function for RValue references

Dark-Passenger:
TODO: add move assignment for CZString class also.

One problem is that I had to change the copy assignment from pass by value to pass by const reference because json_reader was giving errors of ambiguous overloads.
Though to make all tests pass I had to do a const_cast inside the assignment operator since I couldn't do a copy but rather a swap.

I don't like that and would like to fix this. But why is the swap necessary ? shouldn't it be a copy ?

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -398,7 +398,7 @@ Value::Value(double value) {
 
 Value::Value(const char* value) {
   initBasic(stringValue, true);
-  JSON_ASSERT_MESSAGE(value != NULL, "Null Value Passed to Value Constructor");	
+  JSON_ASSERT_MESSAGE(value != NULL, "Null Value Passed to Value Constructor");
   value_.string_ = duplicateAndPrefixStringValue(value, static_cast<unsigned>(strlen(value)));
 }
 
@@ -508,10 +508,18 @@ Value::~Value() {
   value_.uint_ = 0;
 }
 
-Value& Value::operator=(Value other) {
+Value& Value::operator=(const Value& other) {
+  swap(const_cast<Value&>(other));
+  return *this;
+}
+
+#if JSON_HAS_RVALUE_REFERENCES
+Value& Value::operator=(Value&& other) {
+  initBasic(nullValue);
   swap(other);
   return *this;
 }
+#endif
 
 void Value::swapPayload(Value& other) {
   ValueType temp = type_;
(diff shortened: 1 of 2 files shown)


Dark-Passenger:
should fix all things for #621

Dark-Passenger:
Well TODO completed too. Added move assignment for CZString also.

cdunn2001:
Thanks. I'll bump the minor version before the next release.

cfyzium:
> I don't like that and would like to fix this. But why is the swap necessary ? shouldn't it be a copy ?

I believe it should be a copy. Moving/swapping objects on assignment is counterintuitive, especially so since argument of assignment operator is declared to be a const reference. It also breaks the code that relies on value semantics, e. g. working with containers. AFAIK it was one of the main reasons to remove auto_ptr.

cdunn2001:
@cfyzium, I'm mainly interested with maintaining backward-compatibility with *this* library. If we've altered behavior, please help us to fix it.

cfyzium:
@cdunn2001 yes, it clearly changes behavior. I've made pull request #640 to fix that.

src/lib_json/json_value.cpp
```diff
@@ -292,11 +292,21 @@ void Value::CZString::swap(CZString& other) {
...
+  cstr_ = other.cstr_;
+  index_ = other.index_;
   return *this;
 }
 
+#if JSON_HAS_RVALUE_REFERENCES
+Value::CZString& Value::CZString::operator=(CZString&& other) {
+  cstr_ = other.cstr_;
+  index_ = other.index_;
+  other.cstr_ = nullptr;
```
evandrocoan, line 305: Accordingly to https://docs.microsoft.com/en-us/cpp/cpp/move-constructors-and-move-assignment-operators-cpp?view=vs-2019

1. You should zero the original objects after moving the value.
1. Check if you are not moving *this into *this.

```c++
   // Copy the data pointer and its length from the
   // source object.
   _data = other._data;
   _length = other._length;

   // Release the data pointer from the source object so that
   // the destructor does not free the memory multiple times.
   other._data = nullptr;
   other._length = 0;

...
   if (this != &other)
```

BillyDonahue: Thanks for the link. I am a little skeptical of some sloppiness in the article, but it's fine for beginners I guess.

You don't have to zero everything. Only do enough work to make ~T and T::operator= valid and enforce your postconditions regarding the moved-from state of your class, which is up to you. Moved-from object's destructor and reassignment need to be correct, and maybe nothing else.

You don't have to do check against self-move and do nothing. You only have to anticipate that self-move-assign is possible and prove that it leaves the object in a safe moved-from state. Adding these checks without thinking about it can slow down the operation, just to support an outlier case.

All that said!  CZString is not safe against self-move-assign. It will leak its allocation.
That's a for-real bug.

PR #637, opened 2017-07-21
Restore BL's authorship attribution, and add "The Jsoncpp Authors" where it was missing

ssbr:
Requested/noticed in https://github.com/open-source-parsers/jsoncpp/pull/610, and a followup to https://github.com/open-source-parsers/jsoncpp/pull/607 .


cdunn2001:
Thanks!

PR #641, opened 2017-08-03
Fixing warnings. Added JSONCPP_DEPRECATED definition for clang. Also …

maksdamir:
…updating .gitignore to ignore .DS_Store files (Mac OS Finder generated)

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -81,7 +81,7 @@ Features Features::strictMode() {
 // Implementation of class Reader
 // ////////////////////////////////
 
-static bool containsNewLine(Reader::Location begin, Reader::Location end) {
+bool Reader::containsNewLine(Reader::Location begin, Reader::Location end) {
   for (; begin < end; ++begin)
     if (*begin == '\n' || *begin == '\r')
       return true;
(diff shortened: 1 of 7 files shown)


src/lib_json/json_value.cpp
```diff
@@ -1453,8 +1453,13 @@ ptrdiff_t Value::getOffsetStart() const { return start_; }
 ptrdiff_t Value::getOffsetLimit() const { return limit_; }
 
 JSONCPP_STRING Value::toStyledString() const {
-  StyledWriter writer;
-  return writer.write(*this);
+  StreamWriterBuilder builder;
+
+  JSONCPP_STRING out = this->hasComment(commentBefore) ? "\n" : "";
```
cdunn2001: This seems slow. And this might alter the behavior of `toStyledString()`.

src/lib_json/json_reader.cpp
```diff
@@ -1345,6 +1355,25 @@ bool OurReader::readComment() {
   return true;
 }
 
+JSONCPP_STRING OurReader::normalizeEOL(OurReader::Location begin, OurReader::Location end) {
```
cdunn2001: I'd rather not duplicate so much code. Is there a simpler way?

cdunn2001:
I'm not sure this was a helpful change. Deprecation in documentation is fine, but as a compiler warning, it affects our CI builds. I don't like to see any warnings when I compile.

I'm thinking of reverting this entire change, but maybe it would be enough to revert only the new deprecation warnings for the classes.

maksdamir:
Fixing the compiler warnings was actually the exact reason for making that change - when compiling with clang I'm seeing the warnings about methods marked as deprecated in documentation, but not with corresponding attributes.

What warnings do you see currently?

cdunn2001:
Well, when I look at the AppVeyor build, I see tons of warnings now. I'd rather not see *any* warnings. I could live with one or two.

PR #648, opened 2017-08-18
modify double precision 17 to 15

SanJue:
```
#include <iostream>
#include <cstdio>
#include <limits>
using namespace std;

int main()
{
double a = 99.99;
char c[32]={0};
cout<<"i want get "<<a<<endl;
snprintf(c,sizeof(c),"%.17g",a);
cout <<"but use 17, i get "<<c <<endl;
c[32]={0};
snprintf(c,sizeof(c),"%.15g",a);
cout <<"use 15 is right: "<<c <<endl;
cout<<"because std::numeric_limits::digits10 ="<<std::numeric_limits<double>::digits10<<endl;
return 0;
}
```

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -62,7 +62,7 @@
 #endif
 #endif
 
-#if defined(__BORLANDC__)  
+#if defined(__BORLANDC__)
 #include <float.h>
 #define isfinite _finite
 #define snprintf _snprintf
@@ -172,7 +172,7 @@ JSONCPP_STRING valueToString(double value, bool useSpecialFloats, unsigned int p
 }
 }
 
-JSONCPP_STRING valueToString(double value) { return valueToString(value, false, 17); }
+JSONCPP_STRING valueToString(double value) { return valueToString(value, false, 15); }
 
 JSONCPP_STRING valueToString(bool value) { return value ? "true" : "false"; }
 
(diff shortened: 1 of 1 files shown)


SanJue:
run the above code, i got:

i want get 99.99
but use 17, i get 99.989999999999995
use 15 is right: 99.99
because std::numeric_limits::digits10 =15

bernhardHartleb:
No, 17 is the correct value according to
std::numeric_limits<double>::max_digits10.

Serializing a double to a JSON string and back should result in the same
double value by default.

Your example just demonstrates that 99.99 can not be represented exactly in
floating point. But it prints the same result when rounded to 15 digits.

Please change to precision to whatever you need using
Json::StreamWriterBuilder.

Am 18.08.2017 3:28 nachm. schrieb "Li Dongwei" <notifications@github.com>:

run the above code, i got:

i want get 99.99
but use 17, i get 99.989999999999995
use 15 is right: 99.99
because std::numeric_limits::digits10 =15

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/open-source-parsers/jsoncpp/pull/648#issuecomment-323353403>,
or mute the thread

SanJue:
thanks for your reply.

but in case flowing:
        Json::Value paper;
        paper["price"]=99.99;
        paper["weight"]=67.89;
        Json::FastWriter fw;
        cerr<<fw.write(paper).c_str()<<endl;

i get:
       {"price":99.989999999999995,"weight":67.890000000000001}

usually, we use json format to deliver message betweens APIs
but now ,we cant't use jsoncpp to deliver a simple double variable only if we convert it to a string ,just like  paper["price"]=std::to_string(99.99), it tasts bad;

we can change precision of Json::StreamWriterBuilder to solve this problem, but i think most of users
don't know this method;

so our focus is that: std::numeric_limits<double>::max_digits10 and std::numeric_limits<double>::max_digits10, which one we should pick?

refer to http://www.cplusplus.com/reference/limits/numeric_limits/
        digits10:  Number of digits (in decimal base) that can be represented without change.Equivalent 
        max_digits10:Number of digits (in decimal base) required to ensure that values that differ are 
        always differentiated.

i think  digits10 is more suitable.

cdunn2001:
We've had a lot of discussion on this. I don't personally care one way or the other, but a lot of people want a round-trip to match (for representable numbers). And definitely, the tests would need to pass.

melonaerial:
Hello.
Why don't you want to set precision by parameter for JsonWriter class ? I think it is the best solution that can be done.

cdunn2001:
@melonaerial, a precision-setting would be fine. Please submit a PR. If the tests pass, we'll merge it.

Issue #659, opened 2017-09-02
Compile issue


SloCompTech:
Hi,
I installed jsoncpp via meson and ninja, all worked fine, until I tried to comple my program, then I got
```
In file included from /usr/local/include/json/autolink.h:9:0,
                 from /usr/local/include/json/json.h:9,
                 from Test.cpp:1:
/usr/local/include/json/config.h:146:32: error: #if with no expression
/usr/local/include/json/config.h:172:32: error: #if with no expression
In file included from /usr/local/include/json/json.h:10:0,
                 from Test.cpp:1:
/usr/local/include/json/value.h:354:32: error: #if with no expression
```
because `#if` insead of `#ifdef` in some header file.
I will prepare fix, and pull request.

SloCompTech:
Here is meson output
```
The Meson build system
Version: 0.42.0
Source dir: /home/martin/Workspace/Projects/Libraries/jsoncpp
Build dir: /home/martin/Workspace/Projects/Libraries/jsoncpp/build
Build type: native build
Project name: jsoncpp
Native C++ compiler: c++ (gcc 5.4.0)
Build machine cpu family: x86_64
Build machine cpu: x86_64
Configuring version.h using configuration
WARNING: The variable(s) 'JSONCPP_USE_SECURE_MEMORY' in the input file 'src/lib_json/version.h.in' are not present in the given configuration data
Build targets in project: 3
```

datakid:
Can you roll 1.8.4 to include this patch please?

cdunn2001:
Ok. I'd like to get another fix in there too, but we have to wait for TravisCI to start working again. Next week, hopefully.

cdunn2001:
https://github.com/open-source-parsers/jsoncpp/releases/tag/1.8.4

PR #660, opened 2017-09-02
Hotfix for compile issue #659

SloCompTech:
Fix for compile issue when compile there were error in `config.h` and `value.h` in pre-processing code, details in issue #659.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,14 @@ First, install both meson (which requires Python3) and ninja.
 Then,
 
     cd jsoncpp/
-    BUILD_TYPE=shared
-    #BUILD_TYPE=static
-    LIB_TYPE=debug
-    #LIB_TYPE=release
+    BUILD_TYPE=debug
+    #BUILD_TYPE=release
+    LIB_TYPE=shared
+    #LIB_TYPE=static
     meson --buildtype ${BUILD_TYPE} --default-library ${LIB_TYPE} . build-${LIB_TYPE}
     ninja -v -C build-${LIB_TYPE} test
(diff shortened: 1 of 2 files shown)


SloCompTech:
I also fixed instructions for install where parameters were switched

cdunn2001:
* The README fix is good.
* The C++11 setting is fine. I'm not sure it's needed, but we don't want to support non-C++11 anyway.
* I'm not sure about the `JSONCPP_USING_SECURE_MEMORY` change. Someone might be relying on `0` as false. Note that `version.h` sets that to `0` if it was not defined, which is the proper logic.

Maybe there is a different problem with the Meson build, where version.h is somehow dropped?

SloCompTech:
Fixed readme, temporary fixed compile bug with adding `jsoncpp_cdata.set('JSONCPP_USE_SECURE_MEMORY',0)` in meson build config, that means that use of secure memory is disabled, but if you compile any program with `-ljsoncpp` it works, but if you enable USE_SECURE_MEMORY, then you get error on `asString()` method 
```
ConfigModule.cpp:252:54: error: conversion from ‘std::__cxx11::basic_string<char, std::char_traits<char>, Json::SecureAllocator<char> >’ to non-scalar type ‘std::__cxx11::string {aka std::__cxx11::basic_string<char>}’ requested
           std::string text = entry[OPT_TYPE].asString();
```

cdunn2001:
For our future reference:

* http://mesonbuild.com/Configuration.html
* http://mesonbuild.com/Reference-manual.html

PR #666, opened 2017-09-08
Update json_reader.cpp

wolframroesler:
Logging parser errors to stderr is an unexpected side effect, and it messes up the application's error output. Not every bad input to the JSON parser is an error I want logged on stderr. No other JsonCpp function writes anything to stderr before throwing. Also, it uses fprintf which it not really C++-like (which makes me hope it's just some forgotten debug output).


cdunn2001:
Thanks. I don't even know how that got there. I would merge this right away, but our Travis build is suddenly failing. (We need pip3 in order to install Meson. It's *so* difficult to get a simple, normal, up-to-date Linux environment in TravisCI.)

PR #668, opened 2017-09-11
Avoid redundant depreciation warnings in Visual Studio 2012

lanzkron:
See the following [stackoverflow question](https://stackoverflow.com/questions/46151531/how-works-deprecated-warnings-and-how-to-remove-them-when-using-jsoncpp), deprecation warnings will show up even if nothing is used.

**Note:** I'm not able to clone the repository to my computer at the moment so this pull request was created on the web, please make sure it's OK.

PR #669, opened 2017-09-11
Update writer.h

cdunn2001:
Replace #668, which was not properly rebased.

### DIFF
--- a/include/json/writer.h
+++ b/include/json/writer.h
@@ -159,7 +159,6 @@ public:
 #pragma warning(push)
 #pragma warning(disable:4996) // Deriving from deprecated class
 class JSONCPP_DEPRECATED("Use StreamWriterBuilder instead") JSON_API FastWriter : public Writer {
-#pragma warning(pop)
 public:
   FastWriter();
   ~FastWriter() JSONCPP_OVERRIDE {}
@@ -186,6 +185,7 @@ private:
   bool dropNullPlaceholders_;
   bool omitEndingLineFeed_;
 };
+#pragma warning(pop)
 
 /** \brief Writes a Value in <a HREF="http://www.json.org">JSON</a> format in a
  *human friendly way.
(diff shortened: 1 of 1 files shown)


lanzkron:
Thanks for picking this up @cdunn2001 !
Please see the comment on my [stackoverflow answer](https://stackoverflow.com/a/46156833/3848) regarding improvements to this pull request. As I said, I did it online on github,com so it needs a lot of scrutiny.

cdunn2001:
From SO comment:

> Thanks. Note that you forgot a third class StyledStreamWriter. Also, #pragma warning(pop)has to be done at the end of class declaration..else, the warning remains (at least that's what I see with VS 2015)

Ok. I'll try that. But first I'm looking into new TravisCI build failures that have nothing to do with this PR...

include/json/writer.h
```diff
@@ -270,6 +275,8 @@ class JSONCPP_DEPRECATED("Use StreamWriterBuilder instead") JSON_API StyledWrite
  * \sa Reader, Value, Value::setComment()
  * \deprecated Use StreamWriterBuilder.
  */
+#pragma warning(push)
+#pragma warning(disable:4996) // Deriving from deprecated class
```
lanzkron: I don't actually think that disabling the deprecation warnings on this class are necessary. They were added to `StyledWriter` and `FastWriter` since both those files derived from `Writer` which is deprecated (and thus considered by VS to be using `Writer`). `StyledStreamWriter` is itself deprecated (same as `Writer`) and doesn't  use a deprecated class.

I know I referred to the stackoverflow comment but I didn't examine the class it referred to. Now I think it's in error. (Although I don't think it's harmful either, I can't really tell since I'm not at my dev machine).

cdunn2001: Ok. I'll remove the push/pop for StyledStreamWriter. Thanks for following up.

PR #679, opened 2017-09-16
Fix unknown pragma warnings with clang

hughbe:


cdunn2001:
Thanks for fixing this.

PR #688, opened 2017-10-05
Optimize removing negligible values

pkierski:
Sometimes I don't care about removed value, so it will be easier just forget it (and not need to add local variable for it).

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -534,6 +534,7 @@ Json::Value obj_value(Json::objectValue); // {}
   /** \brief Remove the named map member.
 
       Update 'removed' iff removed.
+      'removed' can be NULL if removed value is negligible.
       \param key may contain embedded nulls.
       \return true iff removed (no exceptions)
   */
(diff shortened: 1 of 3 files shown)


BillyDonahue:
Another option that I'd prefer would be to add a single-argument overload of removeValue that doesn't have the pointer at all. Then you have more expressive call sites, we don't talk about `NULL` in the docs, and we skip a conditional.

include/json/value.h
```diff
@@ -534,6 +534,7 @@ Json::Value obj_value(Json::objectValue); // {}
   /** \brief Remove the named map member.
 
       Update 'removed' iff removed.
+      'removed' can be NULL if removed value is negligible.
```
BillyDonahue: The word "negligible" is misused here as it implies a small size.
The removed value might be ignorable, but not because of its size.

wolframroesler:
Seems to be identical to a PR I suggested some weeks ago: #689

It was rejected, the following solution was accepted instead: #693

cdunn2001:
If @wolframroesler is right, can we close this PR?

hjmjohnson:
Alternate solution accepted in solution was accepted instead: #693.

PR #689, opened 2017-10-05
Allow non-deprecated removeMember/removeIndex with null return pointer

wolframroesler:
Sometimes we just want to remove something we don't need anymore. Having to supply a return buffer for the removeMember/removeIndex function to return something we don't care about is a nuisance. There are removeMember/removeIndex functions that don't need a return pointer but they are deprecated. This commit allows the non-deprecated versions to be called with a null pointer as the last argument to simplify things for callers who don't need to get something back.

For example:

```
Json::Value j;

j.removeMember("name"); // nice but deprecated

Json::Value r; j.removeMember("name",&r); // possible but annoying

j.removeMember("name",nullptr); // possible with this commit
```

Note: Implementation only (albeit very trivial), no test coverage.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -1175,7 +1175,7 @@ bool Value::removeMember(const char* key, const char* cend, Value* removed)
   ObjectValues::iterator it = value_.map_->find(actualKey);
   if (it == value_.map_->end())
     return false;
-  *removed = it->second;
+  if (removed) *removed = it->second;
   value_.map_->erase(it);
   return true;
 }
@@ -1212,7 +1212,7 @@ bool Value::removeIndex(ArrayIndex index, Value* removed) {
   if (it == value_.map_->end()) {
     return false;
   }
-  *removed = it->second;
+  if (removed) *removed = it->second;
   ArrayIndex oldSize = size();
   // shift left all items left, into the place of the "removed"
   for (ArrayIndex i = index; i < (oldSize - 1); ++i){


BillyDonahue:
I feel it would be better to just un-deprecate the unary member functions if they're useful.
Optionally-null pointer parameters produce more a complex contract and leave cold branches.

wolframroesler:
@BillyDonahue I agree :) but I'm afraid they were deprecated for a reason. OTOH, the functions that receive the `removed` pointer should handle nullptr gracefully regardless of the unary functions, because why would they take a pointer instead of a reference in the first place?

BillyDonahue:
Well, there are lots of invalid pointers in the memory space, and we can't handle them all gracefully, so I don't like going out of my way to Nerf just the invalid nullptr argument.

As for why it's a pointer instead of a reference, I suspect it was just a style choice to make the mutability more obvious at the call sites (matching, e.g. Google style).

I wonder about the deprecation. I can't think of a reason for it so maybe it's reversible. That's a better API change than adding a new function IMO because it's not really an API change at all!

cdunn2001:
I'm not positive, but I think those were deprecated because of their misleading return values. They returned `nullValue` if missing, but that is a valid value.

I guess we could undeprecate, but only to return nothing. I gues returning `bool` would be ok too, but since the old version returned a real thing, I think `void` is better. If you actually want the removed value, then pass a pointer. (A reference would obscure the side-effect.)

wolframroesler:
Can we simply change the functions to return void instead of a value? This could break client code. If it's ok by you, I'll submit it as a new pull request.

cdunn2001:
If I were you, I would simply re-use this branch and this pull-request, after rebasing. `git push -f origin master`

wolframroesler:
I'm a bit reluctant to force push a branch that was pushed already. Also, should have used a branch in the first place, instead of editing on master. Submitted a new pull request: https://github.com/open-source-parsers/jsoncpp/pull/693

cdunn2001:
@wolframroesler, yes, if you use feature-branches, then you'll have less worry about a force-push. Anyway, I'll merge the new PR after the minor bug is fixed and the tests pass. Thanks.

wolframroesler:
Too good I decided not to follow your advice about the force push ... :-)
http://www.monkeyuser.com/2017/it-hell/

PR #693, opened 2017-10-16
Un-deprecate removeMember overloads, return void

wolframroesler:
Sometimes we just want to remove something we don't need anymore. Having
to supply a return buffer for the removeMember function to return something
we don't care about is a nuisance. There are removeMember overloads that
don't need a return buffer but they are deprecated. This commit un-deprecates
these overloads and modifies them to return nothing (void) instead of the
object that was removed.

Further discussion: https://github.com/open-source-parsers/jsoncpp/pull/689

WARNING: Changes the return type of the formerly deprecated removeMember
overloads from Value to void. May break existing client code.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -1194,12 +1194,12 @@ void Value::removeMember(const char* key)
   if (type_ == nullValue)
     return;
 
-  CZString actualKey(key, strlen(key), CZString::noDuplication);
+  CZString actualKey(key, unsigned(strlen(key)), CZString::noDuplication);
   value_.map_->erase(actualKey);
 }
 void Value::removeMember(const JSONCPP_STRING& key)
 {
-  return removeMember(key.c_str());
+  removeMember(key.c_str());
 }
 
 bool Value::removeIndex(ArrayIndex index, Value* removed) {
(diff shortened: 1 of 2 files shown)


src/lib_json/json_value.cpp
```diff
@@ -1187,18 +1187,17 @@ bool Value::removeMember(JSONCPP_STRING const& key, Value* removed)
...
-  Value removed;  // null
-  removeMember(key, key + strlen(key), &removed);
-  return removed; // still null if removeMember() did nothing
+  CZString actualKey(key, strlen(key), CZString::noDuplication);
+  value_.map_->erase(actualKey);
 }
-Value Value::removeMember(const JSONCPP_STRING& key)
+void Value::removeMember(const JSONCPP_STRING& key)
 {
   return removeMember(key.c_str());
```
cdunn2001: I think we need to drop the `return` here.

cdunn2001:
```cpp
../src/lib_json/json_value.cpp:1205:26: error: return-statement with a value, in function returning 'void' [-fpermissive]

     return nullSingleton();

                          ^

../src/lib_json/json_value.cpp:1209:10: error: return-statement with a value, in function returning 'void' [-fpermissive]

   return removed; // still null if removeMember() did nothing
          ^
```

src/lib_json/json_value.cpp
wolframroesler: Returning void from a void function works fine, it's just a matter of style. I left the return in to reduce the number of changed lines. Will submit a new PR but don't think that's the reason for the errors we got from the CI builds.

PR #694, opened 2017-10-18
Add comment to README with instructions on installing to a specific path

astrobit:
By default, ninja / meson install to /usr/local or the system equivalent. Some users (such as myself) work in complex environments where it is impossible or undesirable to install to such a location. While it didn't take much work for me to find out how to do this, it is beneficial to explicitly state it in the installation instructions.


cdunn2001:
Thanks! Feel free to update our wiki too.

PR #699, opened 2017-11-10
MSVC warning fixes in tests

MarcelRaad:
- only use "#pragma GCC" on GCC-compatible compilers
- suppress deprecation warnings also on MSVC


cdunn2001:
Thanks!

PR #703, opened 2017-11-27
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/amalgamate.py
+++ b/amalgamate.py
@@ -56,7 +56,7 @@ def amalgamate_source(source_top_dir=None,
            target_source_path: output .cpp path
            header_include_path: generated header path relative to target_source_path.
     """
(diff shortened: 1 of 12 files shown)


cdunn2001:
I'm afraid we might need to bump the soname because of the change to `yamlCompatiblityEnabled_`. Not sure. Thanks for the changes anyway.

jsoref:
That's a private member, I would hope it wouldn't matter.

cdunn2001:
Debian has a tool that checks for backward incompatible changes. It's a bit over-restrictive, and I don't remember exactly what it looks for. But don't worry about it. Better to get all the misspellings fixed at once.

PR #707, opened 2017-12-05
Fix sign mismatch in `valueToString`

remyjette:
`valueToString` takes an argument `unsigned int precision`, but it is used with `%d` rather than `%u` in the `snprintf` format string. Make the format string look for an unsigned value instead.


cdunn2001:
Thanks, @remyjette!

PR #712, opened 2017-12-18
fix a bug in utf8ToCodepoint

yuwenyong:
replace code point in range(0xD800, 0xDFFF) to replacement mark


cdunn2001:
Looks good. We'll have to wait for TravisCI to start passing again. I don't think it's a jsoncpp bug.

PR #726, opened 2018-01-12
Value::copy() creates a deep copy now

### LINKED ISSUE #704: Crash after use copy
Simple example:
```cpp
int main(int argc, char* argv[])
{
	Json::Value some_array;

	some_array.append( "text1" );
	some_array.append( "text2" );
	some_array.append( "text3" );

	{
		Json::Value al1, al2;
		al1.copy( some_array );
		al2.copy( some_array );
	}
	// Crash here
}
```
In older version (from sourceforge) works fine. If copy replace with "=", some_array become nullValue

okodron:
This fixes #704.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -606,6 +606,7 @@ Json::Value obj_value(Json::objectValue); // {}
 
 private:
   void initBasic(ValueType type, bool allocated = false);
+  void releasePayload();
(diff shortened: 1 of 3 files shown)


cdunn2001:
@okodron, nice work. Thank you!

PR #729, opened 2018-01-12
Fix C4996 warning with VS2017 by postponing the deprecation of Reader

### LINKED ISSUE #728: Fix warning C4996 with VS2017
Class Json::Reader is marked as deprecated. Unfortunately, VS2017 issues a bunch of warnings even when client code is not using Json::Reader at all. The warnings refer to usage of Json::Reader *within* Json::Reader itself:
```
...\jsoncpp\reader.h(38): warning C4996: 'Json::Reader::Char': Use CharReader and CharReaderBuilder instead
...\jsoncpp\reader.h(37): note: see declaration of 'Json::Reader::Char'
...\jsoncpp\reader.h(181): warning C4996: 'Json::Reader::Location': Use CharReader and CharReaderBuilder instead
...\jsoncpp\reader.h(38): note: see declaration of 'Json::Reader::Location'
...\jsoncpp\reader.h(182): warning C4996: 'Json::Reader::Location': Use CharReader and CharReaderBuilder instead
[...]
```

This minimal example also results in warnings: 
```
class Foo
{
    typedef char Char; // C4996 WARNING HERE
};

void main()
{
  // not using Foo
}
```

What works for VS2017 is the following:

```
// class declaration without declaring as "deprecated" yet
class Foo
{
    typedef char Char; // C4996 WARNING HERE
};
// now make it deprecated
class __declspec(deprecated("FOOBAR")) Foo;
```

I have *not* tested it whether this also works on other platforms with the JSON_DEPRECATED macro.

oxygene:
Should fix #728.
This has been only tested locally running VS2017.

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -32,7 +32,7 @@ namespace Json {
  *
  * \deprecated Use CharReader and CharReaderBuilder.
  */
-class JSONCPP_DEPRECATED("Use CharReader and CharReaderBuilder instead") JSON_API Reader {
+class JSON_API Reader {
 public:
   typedef char Char;
   typedef const Char* Location;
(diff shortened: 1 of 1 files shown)


oxygene:
Thanks to the ci build, I've already realized, that gcc doesn't really like this workaround:

```
../include/json/reader.h:250:83: warning: type attributes ignored after type is already defined [-Wattributes]
```

So apparently a new idea is needed to fix this...

oxygene:
New idea: only deprecate the constructors, not the whole class. Should hopefully catch all deprecated uses of the Reader class, without spamming the compiler output with false positive warnings.

cdunn2001:
@oxygene, good idea. Thank you!

PR #736, opened 2018-01-29
Move the existing value to "removed" argument; removed is optional (could be nullptr)

maxim-ky:
Hi guys,

There're two commits here. 
The first one could be considered to be a minor bug fix (_std::move_).
The second one is disputable. It slightly changes _Value::removeMember_ interface, so it could take _nullptr_ as the "removed" argument. It seems to be a pretty common use case since that extra returned value is not needed for everyone. Also, changing tons of code to support the new _removeMember_ interface could be much easier with this nullptr-feature.

Regards


BillyDonahue:
looks good to me

PR #751, opened 2018-03-06
Remove std::swap<Json::Value> in favor of ADL

BillyDonahue:
Comply with the [Swappable](http://en.cppreference.com/w/cpp/concept/Swappable) concept without opening namespace std.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -880,14 +880,9 @@ public:
   pointer operator->() const { return &deref(); }
 };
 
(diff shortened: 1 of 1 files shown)


ya1gaurav:
Can we move the definition to source file? As i already mentioned some structure analysis tool report it as call from include file to source file.

BillyDonahue:
@ya1gaurav , what tool is this? Why could such a call be a problem? I don't see what issue it's trying to prevent.

PR #752, opened 2018-03-10
Add setting precision for json writers and also add decimal places precision type.

melonaerial:
In thread of issue https://github.com/open-source-parsers/jsoncpp/pull/648 was an idea to make precision settable. So, this PR do that and also add decimal places precision setting.

### DIFF
--- a/include/json/writer.h
+++ b/include/json/writer.h
@@ -106,6 +106,10 @@ public:
       - If true, outputs non-finite floating point values in the following way:
         NaN values as "NaN", positive infinity as "Infinity", and negative infinity
         as "-Infinity".
+    - "precision": int
+      - Number of precision digits for formatting of real values.
+    - "precisionType": "significant"(default) or "decimal"
+      - Type of precision for formatting of real values.
 
     You can examine 'settings_` yourself
     to see the defaults. You can also write and read them just like any
@@ -177,9 +181,6 @@ public:
 
   void omitEndingLineFeed();
 
-  void setPrecision(unsigned int precision);
-  void setPrecisionType(PrecisionType precisionType);
-
 public: // overridden from Writer
   JSONCPP_STRING write(const Value& root) JSONCPP_OVERRIDE;
 
@@ -190,8 +191,6 @@ private:
   bool yamlCompatibilityEnabled_;
   bool dropNullPlaceholders_;
   bool omitEndingLineFeed_;
-  UInt precision_;
-  PrecisionType precisionType_;
 };
 #if defined(_MSC_VER)
 #pragma warning(pop)
@@ -237,9 +236,6 @@ public: // overridden from Writer
    */
   JSONCPP_STRING write(const Value& root) JSONCPP_OVERRIDE;
 
-  void setPrecision(unsigned int precision);
-  void setPrecisionType(PrecisionType precisionType);
-
 private:
   void writeValue(const Value& value);
   void writeArrayValue(const Value& value);
@@ -262,8 +258,6 @@ private:
   unsigned int rightMargin_;
   unsigned int indentSize_;
   bool addChildValues_;
-  UInt precision_;
-  PrecisionType precisionType_;
 };
 #if defined(_MSC_VER)
 #pragma warning(pop)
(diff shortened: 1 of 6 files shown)


include/json/writer.h
```diff
@@ -187,6 +190,8 @@ class JSONCPP_DEPRECATED("Use StreamWriterBuilder instead") JSON_API FastWriter
   bool yamlCompatibilityEnabled_;
   bool dropNullPlaceholders_;
   bool omitEndingLineFeed_;
+  UInt precision_;
```
cdunn2001: We cannot change the size of a class in a public header.

Let's concentrate on the newer CharReader and StreamWriter versions. (Everybody hates them because users do not like to type a few extra characters, but those classes be altered as often as we like without impacting binary-compatibility.)

cdunn2001:
If you drop the changes to the deprecated Reader and Writer classes, then I like this change, though it's big enough to deserve a close review.

include/json/writer.h
melonaerial: Fixed.

melonaerial:
These changes only for Writer. For this issue we don't need any changes for Reader.

src/lib_json/json_writer.cpp
```diff
@@ -118,21 +118,29 @@ JSONCPP_STRING valueToString(UInt value) {
...
+JSONCPP_STRING valueToString(double value, bool useSpecialFloats, unsigned int precision, PrecisionType precisionType) {
   // Allocate a buffer that is more than large enough to store the 16 digits of
   // precision requested below.
   char buffer[36];
   int len = -1;
 
   char formatString[15];
-  snprintf(formatString, sizeof(formatString), "%%.%ug", precision);
+  if (precisionType == PrecisionType::significantDigits) {
+    snprintf(formatString, sizeof(formatString), "%%.%ug", precision);
```
cdunn2001: Shouldn't this be `e` instead of `g`?
* http://en.cppreference.com/w/cpp/io/c/fprintf

src/lib_json/json_writer.cpp
```diff
@@ -1157,6 +1172,14 @@ StreamWriter* StreamWriterBuilder::newStreamWriter() const
   } else {
     throwRuntimeError("commentStyle must be 'All' or 'None'");
   }
+  PrecisionType precisionType(significantDigits);
+  if (pt_str == "significant") {
```
cdunn2001: These names seem long and ambiguous. Maybe `exp` and `noexp` are better names. Or something else?

src/test_lib_json/main.cpp
```diff
@@ -1757,6 +1757,27 @@ JSONTEST_FIXTURE(ValueTest, precision) {
     expected = "0.25634569487374054";
     result = Json::writeString(b, v);
     JSONTEST_ASSERT_STRING_EQUAL(expected, result);
+
+    b.settings_["precision"] = 5;
+    b.settings_["precisionType"] = "decimal";
```
cdunn2001: Maybe `floatPrecision` and `floatFormat`?

src/lib_json/json_writer.cpp
melonaerial: I was thinking about saving old behaviour with `g` when used the shortest variant for float, because maybe some guys expect old behaviour from code. Maybe it is better to add third type for `e` ?

src/lib_json/json_writer.cpp
melonaerial: I used `significant` word for old-style precision formatting. see https://github.com/open-source-parsers/jsoncpp/pull/752#discussion_r173970681. Answer it please. If we delete old-style formatting - then of course it is better change names.

src/test_lib_json/main.cpp
cdunn2001: * http://open-source-parsers.github.io/jsoncpp-docs/doxygen/class_json_1_1_stream_writer_builder.html

Oh, you're right. I forgot about that.

Yes, we need to keep the word "precision". I guess "precisionType" is fine.

src/lib_json/json_writer.cpp
cdunn2001: You're right. Preserving the old way by default is a good idea.

I don't have an opinion on a third type.

Let's drop this latest commit. Then we can merge.

melonaerial: Ok, i've dropped last commit. U can merge when test will be don.

cdunn2001:
If there are any complaints, we may have to revert to look more closely. But it looks good to me. Thanks!

PR #770, opened 2018-05-08
refactoring cross compiler macro

fo40225:
refactoring cross compiler macro

If compiler is c++11 compliant, use std version.


cdunn2001:
I'll merge this for now, but I wouldn't be surprised if we end up having to revert this. There are just too many platforms and compiler versions to test them all.

BillyDonahue:
I think it might be a little simpler to just not use snprintf and sscanf. They are being used in the simplest ways. We could also rename them to something Json-specific instead of doing a #define to squat on a standard library function, which might interfere with macro expansions in these files that might happen to use snprintf or sscanf.

Good luck!

PR #775, opened 2018-05-20
Apply clang format

BillyDonahue:
We have a .clang-format file in the root of the project.
This change just applies it.

Clang-format also reorders #include directives, exposing a missing #include in json_tool.h, so fix that too.


BillyDonahue:
I've noticed that places where our block comments are already a little weird, clang-format will make them worse. We could identify and fix those by hand, I guess.

PR #776, opened 2018-05-20
Reapply clang format

BillyDonahue:
We have a .clang-format file in the root of the project.
This change just applies it.

Clang-format also reorders #include directives, exposing a missing #include in json_tool.h, so fix that too.

redo PR #775


cdunn2001:
Thanks, @BillyDonahue.

PR #788, opened 2018-06-05
Fix #782

### LINKED ISSUE #782: python amalgamate.py - error #include <json/config.h>
fatal error C1083: Cannot open include file: 'json/config.h': No such file or directory

Fix
#if !defined(JSON_IS_AMALGAMATION)
#include <json/config.h>
#endif

?

pavel-pimenov:


cdunn2001:
I've been away for a while. Thanks for the fix.

PR #789, opened 2018-06-05
Add JSON_IS_AMALGAMATION guards around config.h include

mcbridejc:
This include was added a couple of weeks ago, and it looks like it broke the amalgamated build: 

json/jsoncpp.cpp:95:25: fatal error: json/config.h: No such file or directory


cdunn2001:
Oh, sorry. Somebody else might have done this already, and I think we merged the other version. But if you rebase, we can probably accept yours.

mcbridejc:
Cool. As long as it's fixed that's what matters! Thanks for letting me know.

PR #792, opened 2018-06-12
Add position independent code feature to CMakeList.txt

kamelc:
Enable Position Independent Code for shared lib


cdunn2001:
Thanks!

PR #797, opened 2018-06-16
BORLAND compatible

phy0292:
BORLAND compatible


cdunn2001:
There is no guarantee that Borland will *continue* to work, but I will merge this for now.

PR #804, opened 2018-07-03
allow nullptr when not care the removed array value

yantaozhao:


cdunn2001:
Sorry for delay. Seems like a good change. Thanks!

Issue #801, opened 2018-07-02
Access violation during getFormattedErrorMessages, if Parser read JSON from stream


lberserq:
Dear jsoncpp dev team,
I have faced with this bug, which is really annoying

Prerequisites:
jsoncpp-1.7.2
Affected OS: Windows/Linux 
Test sources: https://pastebin.com/YsNpLq86

Steps to reproduce:
wget https://pastebin.com/raw/YsNpLq86 -O test.cpp
g++ -std=c++11 -I/usr/include/jsoncpp test.cpp -ljsoncpp -o test

valgrind ./test

Actual result
```
==1427== Memcheck, a memory error detector
==1427== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==1427== Using Valgrind-3.11.0 and LibVEX; rerun with -h for copyright info
==1427== Command: ./test
==1427==
==1427== Invalid read of size 1
==1427==    at 0x4E4755B: Json::Reader::getLocationLineAndColumn(char const*, int&, int&) const (in /usr/lib/x86_64-linux-gnu/libjsoncpp.so.1.7.2)
==1427==    by 0x4E475CF: Json::Reader::getLocationLineAndColumn[abi:cxx11](char const*) const (in /usr/lib/x86_64-linux-gnu/libjsoncpp.so.1.7.2)
==1427==    by 0x4E488BA: Json::Reader::getFormattedErrorMessages[abi:cxx11]() const (in /usr/lib/x86_64-linux-gnu/libjsoncpp.so.1.7.2)
==1427==    by 0x4011CD: main (in /home/lberserq/test/test)
==1427==  Address 0x5ce8580 is 0 bytes inside a block of size 31 free'd
==1427==    at 0x4C2F24B: operator delete(void*) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==1427==    by 0x4E4D75B: Json::Reader::parse(std::istream&, Json::Value&, bool) (in /usr/lib/x86_64-linux-gnu/libjsoncpp.so.1.7.2)
==1427==    by 0x4011AD: main (in /home/lberserq/test/test)
==1427==  Block was alloc'd at
==1427==    at 0x4C2E0EF: operator new(unsigned long) (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
[...]

PR #811, opened 2018-07-22
(-) Fix JSON-CPP #801

lberserq:
Read stream to readers document_ instead of stack-located dangling
variable

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -120,9 +120,10 @@ bool Reader::parse(std::istream& is, Value& root, bool collectComments) {
 
   // Since JSONCPP_STRING is reference-counted, this at least does not
   // create an extra copy.
-  std::getline(is, document_, (char)EOF);
+  JSONCPP_STRING doc;
+  std::getline(is, doc, (char)EOF);
 
-  return parse(document_.data(), document_.data() + document_.size(), root, collectComments);
+  return parse(doc, root, collectComments);
 }
 
 bool Reader::parse(const char* beginDoc,
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -120,9 +120,9 @@ bool Reader::parse(std::istream& is, Value& root, bool collectComments) {
 
   // Since JSONCPP_STRING is reference-counted, this at least does not
   // create an extra copy.
-  JSONCPP_STRING doc;
-  std::getline(is, doc, (char)EOF);
-  return parse(doc.data(), doc.data() + doc.size(), root, collectComments);
+  std::getline(is, document_, (char)EOF);
+
+  return parse(document_.data(), document_.data() + document_.size(), root, collectComments);
 }
 
 bool Reader::parse(const char* beginDoc,
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -2499,6 +2499,17 @@ JSONTEST_FIXTURE(RValueTest, moveConstruction) {
 #endif
 }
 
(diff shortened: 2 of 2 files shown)


src/lib_json/json_reader.cpp
```diff
@@ -120,9 +120,9 @@ bool Reader::parse(std::istream& is, Value& root, bool collectComments) {
 
   // Since JSONCPP_STRING is reference-counted, this at least does not
   // create an extra copy.
-  JSONCPP_STRING doc;
-  std::getline(is, doc, (char)EOF);
-  return parse(doc.data(), doc.data() + doc.size(), root, collectComments);
+  std::getline(is, document_, (char)EOF);
+
+  return parse(document_.data(), document_.data() + document_.size(), root, collectComments);
```
cdunn2001: Are you sure this is needed?
```cpp
 106 bool Reader::parse(const std::string& document,
 107                    Value& root,
 108                    bool collectComments) {
 109   document_.assign(document.begin(), document.end());
 110   const char* begin = document_.c_str();
 111   const char* end = begin + document_.length();
 112   return parse(begin, end, root, collectComments);
 113 }
```
Doesn't that do the same thing? Are you just trying to avoid a copy? I don't see anything actually different here. The error-message points into `document_` either way, doesn't it?

* http://www.cplusplus.com/reference/string/string/assign/
* https://stackoverflow.com/questions/194634/string-c-str-vs-data

src/test_lib_json/main.cpp
```diff
@@ -2499,6 +2499,17 @@ JSONTEST_FIXTURE(RValueTest, moveConstruction) {
...
 
+struct ErrorTest : JsonTest::TestCase {};
+JSONTEST_FIXTURE(ErrorTest, getErrorWorksFineWithStreams)
+{
+  const char TEST_VALUE[] = R"__({"X":3, "Y":oops})__";
+  JSONCPP_ISTRINGSTREAM ss(TEST_VALUE);
+  Json::Reader reader;
+  Json::Value result;
+  JSONTEST_ASSERT(!reader.parse(ss, result));
+  JSONTEST_ASSERT(reader.getFormattedErrorMessages().size());
```
cdunn2001: Should we also test the *content* of the formatted error messages?

Anyway, the real issue is whether pointers are correct. I'm not sure how to test that, since it's internal to the class.

lberserq: Yep, You are right, unfortunately, the only way is to look up through the internals of the class(downcasting via layout) and check it with /proc/self/maps or use _CrtIsValidHeapPointer or using VirtualQuery but all of such methods are messy.

src/lib_json/json_reader.cpp
lberserq: Yep, you are right I just forgot to look through the methods and avoid changing the exact behavior of the method. No, I am not intended to avoid copy in this strange manner

src/test_lib_json/main.cpp
lberserq: Yes, Determining the type of pointer in C++ requires a lot of mess(parsing /proc/self/maps, checking  regions via VirtualQuery, or using MSVC specific _CrtIsValidHeapPointer)

hjmjohnson:
@lberserq Can you please address comments by @cdunn2001 .  "Resolve converstation" after you have addressed the comments/concerns raised.

hjmjohnson:
@lberserq Should this issue be closed?  If this issue is still important:

- [ ] Rebase on the current master
- [ ] Address comments by @cdunn2001 
- [ ] force-push this branch to update the PR.

hjmjohnson:
Closing due to inactivity. Please re-open if this issue is still relevant.

PR #827, opened 2018-10-03
CMake: Unique lib target name 

jschueller:
An issue when using the imported target from the cmake config file is that the target name differs whether the lib is compiled statically of dynamically so the client code cannot know easily whether to use the dynamic name (jsoncpp_lib) or the static name (jsoncpp_lib_static).

This PR sets a unique name for the lib so the client code can always use:
```
find_package (jsoncpp NO_MODULE REQUIRED)
target_link_libraries (client_target jsoncpp_lib)
```

Also BUILD_STATIC_LIBS is unused and not standard.

And also use CMAKE_CROSSCOMPILING_EMULATOR to run tests when cross-compiling.


jschueller:
ping @open-source-parsers/jsoncpp

jschueller:
ping @cdunn2001

cdunn2001:
We strongly prefer **meson** to **cmake**, but if you are confident in this change, go ahead and merge. You might help other **cmake** users.

jschueller:
@cdunn2001 I dont have rights to merge

PR #833, opened 2018-10-23
Fix debug libs

watsontandrew:
Fix debug lib names


cdunn2001:
We prefer **meson** to **cmake** now, but if you want a change, please resubmit. Sorry for delays.

Issue #836, opened 2018-11-14
removeMember causes exception


stefanofiorentino:
the deprecated version of the removeMember works like a charm, the "new" (untested to me) version of the function, which allow for `removed' parameters causes SIGSEGV in case nullptr is passed as Value* (pointer not checked for nullness?). 

== Lib version ==
Package: libjsoncpp-dev
Version: 1.7.4-3
Priority: extra
Section: libdevel
Source: libjsoncpp
Origin: Ubuntu (18.04)

== Reproducible code == 
   ```cpp
 GIVEN("a Json::Value instance with key1 member")
    {
        Json::Value jsonValue = Json::objectValue;
        jsonValue["key1"] = Json::objectValue;
        WHEN("removeMember is called with const char*")
        {
            std::string key = "key1";
            jsonValue.removeMember(key.c_str(), nullptr);
            THEN("SIGSEGV is rose")
            {
                std::cout << "never be here" << std::endl;
            }
        }
    }
```

PR #837, opened 2018-11-14
issue_836: check if `removed' is a valid pointer before copy data to it

stefanofiorentino:
functions involved:
  - bool Value::removeMember(const char* key, const char* cend, Value* removed)

Signed-off-by: Stefano Fiorentino <stefano.fiore84@gmail.com>


src/lib_json/json_value.cpp
```diff
@@ -1207,11 +1207,13 @@ bool Value::removeMember(const char* begin, const char* end, Value* removed) {
   if (it == value_.map_->end())
     return false;
   if (removed)
+  {
```
BillyDonahue: Join with previous line to match surrounding brace style.

But other than adding helpful braces, this change doesn't appear to have changed the code's behavior.
The #if/#else code sections were single statements. In both the #if and the #else sections, adding braces around the section doesn't change its meaning.

Maybe I'm missing something?

BillyDonahue:
Attached to Issue #836

src/lib_json/json_value.cpp
stefanofiorentino: You right, I was at tag 1.8.3 and the fix you mention has been pushed later.
The PR is still valid for the test part that is as important as the check-for-null-pointer stuff.

stefanofiorentino:
@BillyDonahue any news on this? in my opinion test are useful to detect future regression introductions on pointer de-referentiation.

PR #847, opened 2018-12-12
Use cxx11 headers

hjmjohnson:


hjmjohnson:
The windows build incantation is now correct for shared libraries.

PR #849, opened 2018-12-12
COMP: Use nullptr instead of 0 or NULL

hjmjohnson:
The check converts the usage of null pointer constants (eg. NULL, 0) to
use the new C++11 nullptr keyword.

SRCDIR=/Users/johnsonhj/src/jsoncpp #My local SRC
BLDDIR=/Users/johnsonhj/src/jsoncpp/cmake-build-debug/ #My local BLD

cd /Users/johnsonhj/src/jsoncpp/cmake-build-debug/
run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,modernize-use-nullptr  -header-filter=.* -fix

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -221,7 +221,7 @@ private:
                                    Location end,
                                    unsigned int& unicode);
   bool
-  addError(const JSONCPP_STRING& message, Token& token, Location extra = 0);
+  addError(const JSONCPP_STRING& message, Token& token, Location extra = nullptr);
   bool recoverFromError(TokenType skipUntilToken);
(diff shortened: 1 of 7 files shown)


hjmjohnson:
@ibc @bernhardHartleb @mnaberez @bmcdorman I am interested in using jsoncpp, but I am extremely concerned that there is not a clear process for reviewing pull requests in a timely manner.

Is there  a separate mailing list/blog site/communication channel that should be used to have a conversation about code contributions?  

I am willing to maintain and update the CMake build system (it is required for incorporating into many of the National Alliance for Medical Image Computing software tools).

Best regards,
Hans

PR #850, opened 2018-12-12
PERF: Replace explicit return calls of constructor

hjmjohnson:
Replaces explicit calls to the constructor in a return with a braced
initializer list. This way the return type is not needlessly duplicated in the
function definition and the return statement.

SRCDIR=/Users/johnsonhj/src/jsoncpp #My local SRC
BLDDIR=/Users/johnsonhj/src/jsoncpp/cmake-build-debug/ #My local BLD

cd /Users/johnsonhj/src/jsoncpp/cmake-build-debug/
run-clang-tidy.py -extra-arg=-D__clang__ -checks=-*,modernize-return-braced-init-list  -header-filter=.* -fix

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -69,7 +69,7 @@ Features::Features()
     : allowComments_(true), strictRoot_(false),
       allowDroppedNullPlaceholders_(false), allowNumericKeys_(false) {}
 
-Features Features::all() { return Features(); }
+Features Features::all() { return {}; }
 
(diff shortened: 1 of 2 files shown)


hjmjohnson:
@cdunn2001 @mnaberez What is the status of jsoncpp?  Is there a different mechanism to have code reviewed and incorporated?  I've not heard any comments regarding the submitted PR's and was wondering if there is a better mechanism for making these suggested changes?

hjmjohnson:
@cdunn2001 Thank you.

PR #853, opened 2018-12-29
Add support for building in Visual Studio 2017

clarkezone:
Add a solution and vcxproj file for building in Visual Studio 2017.


makefiles/msvc2017/jsoncpp.sln
```diff
@@ -0,0 +1,64 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.102
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "lib_json", "lib_json.vcxproj", "{B84F7231-16CE-41D8-8C08-7B523FF4225B}"
```
cdunn2001: I you sure we want these specific strings in the repository? I don't mind merging, but I want to be sure that you are confident.

Also, the comment mentions VS15, while the PR title says VS17. Will both be supported?

clarkezone: I basically ran the older visual studio project through the import wizard and this is what came out.  I’m pretty sure VS uses guids to uniquely identify projects.   I will create an new / empty project as a test and see if the above is present, if no I’ll remove from the PR if yes I’ll ack back on this thread

hjmjohnson: @clarkezone Thanks.  Ping me when you are confident of the answer and I'll address right away.

PS:  Can you please rebase on the latest master branch as well?

clarkezone: I did file new project and the version is 15.x per above, and the guid's are also part of the deal as well so I think we're fine with the above comment.  I have rebased and force-pushed the update.

PR #856, opened 2019-01-11
reader: fix signed overflow when parsing negative value

bradking:
Clang's ubsan (-fsanitize=undefined) reports:

    runtime error: negation of -9223372036854775808 cannot be represented in
    type 'Json::Value::LargestInt' (aka 'long'); cast to an unsigned type to
    negate this value to itself

Follow its advice and update the code to remove the explicit negation.


hjmjohnson:
@bradking Can you see why the CI tests are failing?

bradking:
@hjmjohnson it looks like the Travis CI build did not even get to compiling the code:

> Error: HOMEBREW_LOGS was not exported!
> Please don't worry, you likely hit a bug auto-updating from an old version.

PR #858, opened 2019-01-11
ENH: Remove conditional checks for C++11 support

hjmjohnson:
if __cplusplus >= 201103L is always true as C++11 support is required
for this library now.


hjmjohnson:
WRONG:  #if __cplusplus >= 201103L  is not supported for MSVC 2012, so we can not depend on it's setting.

leesangj:
나 공유

LEESANGJOON

PR #867, opened 2019-01-17
Reapply clang-format.

BillyDonahue:
```
  $ clang-format -i -style=file \
        $(find  . | egrep '.*\.(h|cpp|inl)$')

```

Also did some manual commenting to keep test data lines from joining up,
since they rely on `__LINE__` to be meaningful.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -630,7 +630,7 @@ private:
 
     void setComment(const char* text, size_t len);
 
-    char* comment_{nullptr};
+    char* comment_{ nullptr };
   };
 
   // struct MemberNamesTransform
@@ -679,7 +679,7 @@ private:
   enum Kind { kindNone = 0, kindIndex, kindKey };
   JSONCPP_STRING key_;
   ArrayIndex index_{};
-  Kind kind_{kindNone};
+  Kind kind_{ kindNone };
 };
 
 /** \brief Experimental and untested: represents a "path" to access a node.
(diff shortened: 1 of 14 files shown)


include/json/value.h
```diff
@@ -630,7 +630,7 @@ Json::Value obj_value(Json::objectValue); // {}
 
     void setComment(const char* text, size_t len);
 
-    char* comment_{nullptr};
+    char* comment_{ nullptr };
```
BillyDonahue: I don't really like this behavior, but it's what our .clang-format specifies so we'd better either accept it or change our .clang-format file.

src/test_lib_json/main.cpp
```diff
@@ -2354,14 +2354,22 @@ JSONTEST_FIXTURE(CharReaderAllowSpecialFloatsTest, issue209) {
     JSONCPP_STRING in;
   };
   const TestData test_data[] = {
-    { __LINE__, true, "{\"a\":9}" },         { __LINE__, false, "{\"a\":0Infinity}" },
```
BillyDonahue: These obviously need to all be on their own row. The EOL comments tell clang-format not to join them.

include/json/value.h
hjmjohnson: I have no preference, except for consistency.  The status-quo is fine.

I'll also use clang-format in the future to maintain style consistency.

src/test_lib_json/main.cpp
hjmjohnson: +1

PR #868, opened 2019-01-18
Jsoncpp aliases 2

BillyDonahue:
There's no reason for those JSONCPP_STRING etc to be macros.
Let's use real typedefs so the compiler can see them for what they are, and they can have proper namespacing, etc.

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -6,8 +6,11 @@
 #ifndef JSON_CONFIG_H_INCLUDED
 #define JSON_CONFIG_H_INCLUDED
 #include <cstddef>
-#include <cstdint> //typedef int64_t, uint64_t
-#include <string>  //typedef String
+#include <cstdint>
+#include <istream>
+#include <ostream>
+#include <string>
+#include <type_traits>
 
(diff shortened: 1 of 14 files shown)


include/json/config.h
```diff
@@ -170,24 +170,28 @@ typedef Int64 LargestInt;
...
+using IStream = std::istream;
+using OStream = std::ostream;
+} // namespace Json
+
+// Legacy names (formerly macros).
+using JSONCPP_STRING = Json::String;
+using JSONCPP_ISTRINGSTREAM = Json::IStringStream;
+using JSONCPP_OSTRINGSTREAM = Json::OStringStream;
+using JSONCPP_ISTREAM = Json::IStream;
+using JSONCPP_OSTREAM = Json::OStream;
```
hjmjohnson: NICE!

BillyDonahue: Oh yes, I intend to squash it all down to one commit.
It's separate commits just so that it's reviewable steps.
7b1e53c is huge, but it's just a big rename.

hjmjohnson: 👍  Feel free to merge right after squashing.

PR #872, opened 2019-01-19
Do not reference SecureAllocator when not JSONCPP_USING_SECURE_MEMORY

pps83:
Amalgamated version of jsoncpp does not include allocator.h and as a result doesn't have definition of SecureAllocator and results in compilation error when referencing SecureAllocator in config.h

this issue is mentioned in #871

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -173,10 +173,13 @@ typedef UInt64 LargestUInt;
 #define JSON_HAS_INT64
 #endif // if defined(JSON_NO_INT64)
 
+#if JSONCPP_USING_SECURE_MEMORY
 template <typename T>
-using Allocator = typename std::conditional<JSONCPP_USING_SECURE_MEMORY,
-                                            SecureAllocator<T>,
-                                            std::allocator<T>>::type;
+using Allocator = SecureAllocator<T>;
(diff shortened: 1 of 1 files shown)


hjmjohnson:
LGTM, but I'm going to let other have a chance to weigh in.

pps83:
if you use amalgamated version then it would have this code:
```
using Allocator = typename std::conditional<JSONCPP_USING_SECURE_MEMORY,
                                            SecureAllocator<T>,
                                            std::allocator<T>>::type;
```
but there would be no `SecureAllocator` defined anywhere. In other words, it would totally rely on compiler ignoring that template parameter to [std::conditional](https://en.cppreference.com/w/cpp/types/conditional). Proper fix would perhaps be to add allocator.h to amalgamated version, but in any case this change is good imo and there was no point in the first place to go through std::conditional.

hjmjohnson:
@BillyDonahue Are you happy with this improvement?  Can it be merged, or should it be abandoned?

hjmjohnson:
Adding json/allocator.h in the amalgamated header seems to have resolved this issue in an alterantive way.

Closing this issue.  Please re-open if other issues exist.

PR #873, opened 2019-01-19
Add explicit move assignment operator

pps83:
this fixes linking error mentioned in #871 when building chrome/webrtc

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -497,10 +497,18 @@ Value::~Value() {
   value_.uint_ = 0;
 }
 
-Value& Value::operator=(Value other) {
+Value& Value::operator=(const Value& other) {
+  Value tmp(other);
+  swap(tmp);
+  return *this;
+}
+
+#if JSON_HAS_RVALUE_REFERENCES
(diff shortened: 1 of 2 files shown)


pps83:
Also, afaik when there are custom ctor/dtors compiler cannot add implicit move assignment operator which potentially may prevent some optimizations.

hjmjohnson:
correct.

src/lib_json/json_value.cpp
```diff
@@ -497,10 +497,18 @@ Value::~Value() {
   value_.uint_ = 0;
 }
 
-Value& Value::operator=(Value other) {
+Value& Value::operator=(const Value& other) {
+  Value tmp(other);
+  swap(tmp);
```
BillyDonahue: Swapping into a temporary is wasted work.
I think we should rather write "real" operators rather than relying on swap.

I'll work on that now and send a different PR proposal.

hjmjohnson:
@pps83 Could you please rebase to resolve conflicts so that we may merge this branch?

hjmjohnson:
@pps83 This work was supperceeded by work of @BillyDonahue in 0c1cc6e1a373d.

PR #875, opened 2019-01-20
Issue #872: add json/allocator.h in the amalgamated header.

BillyDonahue:
I don't know why we didn't include this before.
It seems to work fine.


cdunn2001:
> Surely we don't care about VS2010 in trunk anymore.

That's correct. We don't.

Issue #882, opened 2019-01-24
Problem when compiling with Visual Studio 2010


AlessandroDUffizi:
I have successfully compiled jsoncpp with Visual studio 2017 and Windows 7
Now I'm trying to do the same with Visual studio 2010 over my laptop with Windows 10.

I got Sintax Error in json_value.cpp each time in the code I found:

---->          return {}; 
Errore	8	error C2059: errore di sintassi: '{'	c:\users\alessandro\chaos_compile\jsoncpp-master\src\lib_json\json_value.cpp	1533	1	lib_json

Errore	9	error C2143: errore di sintassi: ';' mancante prima di '{'	c:\users\alessandro\chaos_compile\jsoncpp-master\src\lib_json\json_value.cpp	1533	1	lib_json

Errore	10	error C2059: errore di sintassi: '{'	c:\users\alessandro\chaos_compile\jsoncpp-master\src\lib_json\json_value.cpp	1546	1	lib_json

Errore	11	error C2143: errore di sintassi: ';' mancante prima di '{'	c:\users\alessandro\chaos_compile\jsoncpp-master\src\lib_json\json_value.cpp	1546	1	lib_json

It seems like the compiler don't recognize {} as a constant empty json value.

I used the msvc2010 jsoncpp.sln   already provided in the Master branch.
I did nothing except compiling.

AlessandroDUffizi:
Okay. I found an old issue. Master need c++11 and MSVC2010 doesn't have that feature. So i Downloaded 0.x.y and it compiled.

But why there's a msvc2010 solution in the master, if it not works on msvc2010 ?
Just for let people upgrade solution from that?
Thanks a lot. Sorry for bothering.

cdunn2001:
We very recently started to require C++11, so we need to delete any 2010 references and document.

@hjmjohnson , could you take care of that? Also, if you can figure out where 2010 stops working, maybe we could tag the commit before that as `last-msvc2010`.

hjmjohnson:
I'd go a bit further.  Both CMake and Meson are compatible with various visual studio environments.  In my experience, those tools (meson or cmake) provide a generic abstraction layer that makes supporting various Visual studio (or xcode, or clion, or unix makefile, or ninja) IDE's more consistent and reusble across difference environments (i.e. OS versions, and IDE patch updates).

I'd recommend removing the visual studio 2017 solution files as well, and support only meson and cmake.

AlessandroDUffizi:
[...]

PR #883, opened 2019-01-24
COMP: Remove build files for unsupported IDE's

hjmjohnson:
The msvc2010 and vs71 IDE's do not support sufficient
C++11 feature sets for jsoncpp.

Remove these build environments.

resolves: #882


pratikpc:
I would say it would be an excellent idea to only support CMake as a build system.
CMake can be used by any user to build the type of project of their choice.
CMake also has substantial cross platform compatibility and has no dependency on any other major tool like Python etc. which again is beneficial

cdunn2001:
We should try to keep msvc2010 support on the `0.10.z` versions, but not on `master` or `1.y.z`.

PR #898, opened 2019-03-18
Update README.md

canihavesomecoffee:
Update the link to the conan page, as https://conan.io/source/jsoncpp/1.8.0/theirix/ci returns a 404.


cdunn2001:
I kinda think this kind of thing should be in wiki, but since it's here now, thanks for the fix.

PR #901, opened 2019-03-23
Implement Value::demand()

res2k:
There's a declaration of a `Value::demand()` method, but an implementation is nowhere to be found.

It would have some utility, though: it's the only method to create a named key that takes `[begin, end)` arguments. This is useful if you want to use a key stored in a `string_view`.

Note that I also changed the return value of `demand()`:
* I think a non-const pointer is more useful as you can immediately modify the `Value`.
* Leaving it as a pointer keeps some symmetry with `find()`.
* This change doesn't affect compatibility much, as `demand()` wasn't implemented, thus there can't be any user of it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,12 +63,12 @@ if(NOT DEFINED CMAKE_BUILD_TYPE)
 endif()
 
 project(JSONCPP
-        VERSION 1.8.4 # <major>[.<minor>[.<patch>[.<tweak>]]]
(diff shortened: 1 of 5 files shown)


cdunn2001:
Seems fine. I don't remember why this was not implemented. Personally, I like `demand()`, so thanks for this.

But you need to bump the minor version number.

res2k:
> But you need to bump the minor version number.

I guess the SOVERSION as well?

cdunn2001:
Yeah, that needs to match.

PR #915, opened 2019-04-28
COMP: Improve const correctness for ValueIterators

### LINKED ISSUE #914: Issues compiling with uClibc++
I'm getting this error:

```
In file included from /home/mangix/devstuff/openwrt/build_dir/target-mips64_octeonplus_64_musl/jsoncpp-0.10.7/src/lib_json/json_value.cpp:153:0:
/home/mangix/devstuff/openwrt/build_dir/target-mips64_octeonplus_64_musl/jsoncpp-0.10.7/src/lib_json/json_valueiterator.inl: In member function 'Json::Value& Json::ValueIteratorBase::deref() const':
/home/mangix/devstuff/openwrt/build_dir/target-mips64_octeonplus_64_musl/jsoncpp-0.10.7/src/lib_json/json_valueiterator.inl:27:20: error: binding reference of type 'Json::Value&' to 'const Json::Value' discards qualifiers
   return current_->second;
          ~~~~~~~~~~^~~~~~
/home/mangix/devstuff/openwrt/build_dir/target-mips64_octeonplus_64_musl/jsoncpp-0.10.7/src/lib_json/json_value.cpp: In member function 'const Json::Value& Json::Value::operator[](Json::Value::ArrayIndex) const':
/home/mangix/devstuff/openwrt/build_dir/target-mips64_octeonplus_64_musl/jsoncpp-0.10.7/src/lib_json/json_value.cpp:937:16: warning: returning reference to temporary [-Wreturn-local-addr]
   return (*it).second;
                ^~~~~~
/home/mangix/devstuff/openwrt/build_dir/target-mips64_octeonplus_64_musl/jsoncpp-0.10.7/src/lib_json/json_value.cpp: In member function 'const Json::Value* Json::Value::find(const char*, const char*) const':
/home/mangix/devstuff/openwrt/build_dir/target-mips64_octeonplus_64_musl/jsoncpp-0.10.7/src/lib_json/json_value.cpp:1010:17: warning: taking address of temporary [-fpermissive]
   return &(*it).second;
                 ^~~~~~
```

hjmjohnson:
The protected deref method had inconsistent interface
of being a const function that returned a non-const
reference.  Resolves #914.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -772,7 +772,14 @@ public:
   char const* memberName(char const** end) const;
 
 protected:
-  Value& deref() const;
+  /*! Internal utility functions to assist with implementing
+  *   other iterator functions. The const and non-const versions
+  *   of the "deref" protected methods expose the protected
+  *   current_ member variable in a way that can often be
+  *   optimized away by the compiler.
+  */
+  const Value& deref() const;
+  Value& deref();
 
   void increment();
 
@@ -895,9 +902,13 @@ public:
     return *this;
   }
 
-  reference operator*() const { return deref(); }
-
-  pointer operator->() const { return &deref(); }
+  /*! The return value of non-const iterators can be
+   *  changed, so the these functions are not const
+   *  because the returned references/pointers can be used
+   *  to change state of the base class.
+   */
+  reference operator*() { return deref(); }
+  pointer operator->() { return &deref(); }
 };
 
 inline void swap(Value& a, Value& b) { a.swap(b); }
--- a/src/lib_json/json_valueiterator.inl
+++ b/src/lib_json/json_valueiterator.inl
@@ -21,7 +21,8 @@ ValueIteratorBase::ValueIteratorBase(
     const Value::ObjectValues::iterator& current)
     : current_(current), isNull_(false) {}
 
-Value& ValueIteratorBase::deref() const { return current_->second; }
+Value& ValueIteratorBase::deref() { return current_->second; }
+const Value& ValueIteratorBase::deref() const { return current_->second; }
 
 void ValueIteratorBase::increment() { ++current_; }
 


neheb:
Mostly resolves. Still getting:

```
src/lib_json/json_value.cpp:937:16: warning: returning reference to temporary [-Wreturn-local-addr]
    return (*it).second;
src/lib_json/json_value.cpp:1010:17: warning: taking address of temporary [-fpermissive]
   return &(*it).second;
```

This is on 0.10.7. 1.8.4 has the same issue. -fpermissive must be passed. Otherwise error.

include/json/value.h
```diff
@@ -895,9 +902,13 @@ class JSON_API ValueIterator : public ValueIteratorBase {
...
 
-  reference operator*() const { return deref(); }
-
-  pointer operator->() const { return &deref(); }
+  /*! The return value of non-const iterators can be
+   *  changed, so the these functions are not const
+   *  because the returned references/pointers can be used
+   *  to change state of the base class.
+   */
+  reference operator*() { return deref(); }
```
cdunn2001: This changes an existing symbol, which would require a major version bump if the symbol is publicly accessible.

Would it be possible to keep these 2 operators as `const`? I think it would be ok to *add* 2 non-const operators, if that compiles.

cdunn2001:
I have always hated this iterator. It might be better to keep this as is, write a new iterator, and use only the new one from our own code. (Does this library actually use the iterator? Maybe not.)

include/json/value.h
neheb: Compiles when const

neheb: I lied.

```
/home/mangix/devstuff/openwrt/build_dir/target-arc_arc700_uClibc/jsoncpp-1.8.4/include/json/value.h:869:45: error: binding reference of type 'Json::ValueIterator::reference' {aka 'Json::Value&'} to 'const Json::Value' discards qualifiers
   reference operator*() const { return deref(); }
```

BillyDonahue:
I think this is likely to be the wrong fix. `operator->` and `operator*` should always be `const`.

For an iterator to mutable, they should be exposing some OTHER object to modification, not the iterator itself.

Are we doing some unusual proxying iterator thing here that would require the unusual departure from iterator convention? I don't think we are.

hjmjohnson:
Abandoning.

PR #919, opened 2019-05-14
small changes to provide compatibility with the -Weffc++ g++ compiler flag

bsouthwe:
I work for a company that uses jsoncpp in our product. We use the -Weffc++ g++ compiler flag in our build system. This flag simply verifies that the code follows the guidelines outlined in Effective C++ by Scott Myers. Our build system actually fails when building with this compiler option and the failure was traced back to a -Weffc++ warning thrown on the jsoncpp package. This pull request includes a few minor changes that will fix these warnings being thrown with the compiler option. We have a local patch that fixes them for our build system, but it would be nice to avoid this for future releases.

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -64,13 +64,8 @@ public:
   JSONCPP_DEPRECATED("Use CharReader and CharReaderBuilder instead")
   Reader(const Features& features);
 
-<<<<<<< HEAD
   Reader(const Reader&) = default;
   Reader& operator=(const Reader&) = default;
-=======
-  Reader(const Reader&) = delete; /* disabled for Weffc++ */
-  Reader& operator=(const Reader&) = delete; /* disabled for Weffc++ */
->>>>>>> d4f7d51576a9522e68324b439a06daa1be0b31a1
 
   /** \brief Read a Value from a <a HREF="http://www.json.org">JSON</a>
    * document.
@@ -197,16 +192,9 @@ private:
 
   class ErrorInfo {
   public:
-<<<<<<< HEAD
     ErrorInfo() : token_(), message_(""), extra_("") {};
     ErrorInfo(const ErrorInfo&) = default;
     ErrorInfo& operator=(const ErrorInfo&) = default;
-=======
-    ErrorInfo() : token_(), message_(), extra_() {}
-    ErrorInfo(const ErrorInfo&) = default;
-    ErrorInfo& operator=(const ErrorInfo&) = delete;
-
->>>>>>> d4f7d51576a9522e68324b439a06daa1be0b31a1
     Token token_;
     String message_;
     Location extra_;
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -44,6 +44,9 @@ public:
    *
    */
   struct StructuredError {
+    StructuredError() : offset_start(0), offset_limit(0), message("") {};
+    StructuredError(const StructuredError&) = default;
+    StructuredError& operator=(const StructuredError&) = default;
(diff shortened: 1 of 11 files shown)


include/json/value.h
```diff
@@ -649,7 +649,7 @@ Json::Value obj_value(Json::objectValue); // {}
 
   class Comments {
   public:
-    Comments() = default;
+    Comments();
```
baylesj: Why remove the default constructor? Now you have a linker error due to Comments::Comments being undefined.

include/json/reader.h
```diff
@@ -186,6 +192,9 @@ class JSON_API Reader {
 
   class ErrorInfo {
   public:
+    ErrorInfo() : token_(), message_(""), extra_("") {};
```
baylesj: Any reason this class is not movable?

include/json/reader.h
```diff
@@ -44,6 +44,9 @@ class JSON_API Reader {
    *
    */
   struct StructuredError {
+    StructuredError() : offset_start(0), offset_limit(0), message("") {};
```
baylesj: Any reason this class is not movable?

include/json/writer.h
```diff
@@ -44,7 +44,12 @@ class JSON_API StreamWriter {
   OStream* sout_; // not owned; will not delete
 public:
   StreamWriter();
+
+  StreamWriter(const StreamWriter&) = default;
```
baylesj: Any reason this class is not movable?

include/json/writer.h
```diff
@@ -169,6 +174,8 @@ class JSONCPP_DEPRECATED("Use StreamWriterBuilder instead") JSON_API FastWriter
     : public Writer {
 public:
   FastWriter();
+  FastWriter(const FastWriter&) = default;
```
baylesj: Any reason this class is not movable?

include/json/writer.h
```diff
@@ -229,6 +236,8 @@ class JSONCPP_DEPRECATED("Use StreamWriterBuilder instead") JSON_API
     StyledWriter : public Writer {
 public:
   StyledWriter();
+  StyledWriter(const StyledWriter&) = default;
```
baylesj: Any reason this class is not movable?

include/json/writer.h
```diff
@@ -301,8 +310,12 @@ class JSONCPP_DEPRECATED("Use StreamWriterBuilder instead") JSON_API
    * \param indentation Each level will be indented by this amount extra.
    */
   StyledStreamWriter(String indentation = "\t");
+
   ~StyledStreamWriter() = default;
 
+  StyledStreamWriter(const StyledStreamWriter&) = default;
```
baylesj: Any reason this class is not movable?

include/json/reader.h
```diff
@@ -61,6 +64,9 @@ class JSON_API Reader {
   JSONCPP_DEPRECATED("Use CharReader and CharReaderBuilder instead")
   Reader(const Features& features);
 
+  Reader(const Reader&) = default;
```
baylesj: Any reason this class is not movable?

baylesj:
For now, we are electing not to merge. This patch as written, functionally just breaks compilation and deletes the implicitly created move constructors. For more information, see:

https://en.cppreference.com/w/cpp/language/move_constructor
```
If no user-defined move constructors are provided for a class type (struct, class, or union), and all of the following is true:

there are no user-declared copy constructors;
there are no user-declared copy assignment operators;
there are no user-declared move assignment operators;
there are no user-declared destructors;
```

PR #928, opened 2019-05-29
Create an example directory and add some code examples.

dota17:
Personally speaking, I think the original code examples for this project are not very easy to find, and the readability is not very high,so I create a example directory and add some code examples here.


CMakeLists.txt
```diff
@@ -181,6 +181,9 @@ endif()
 # Build the different applications
 add_subdirectory( src )
 
+# add example 
+add_subdirectory( example )
+
```
dota17: I know little about cmake, perhaps this caused the CI build failed.

example/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+set(EXAMPLES
+	readFromString
+	readFromStream
+	stringWrite
+	streamWrite
+   )
+
+add_definitions(-D_GLIBCXX_USE_CXX11_ABI)
+set_property(DIRECTORY PROPERTY COMPILE_OPTIONS $(EXTRA_CXX_FLAGS))
```
ferdnyc: The problem was actually here — those should be curly braces: `${EXTRA_CXX_FLAGS}`. `g++` parsed `$(EXTRA_CXX_FLAGS)` as a filename, and failed to load the file.

PR #929, opened 2019-05-31
create a code example directory and add some examples

dota17:
The original code example for this project are not very easy to find, and the readability is not very high, so I create a example directory under the  root directory, and add some examples here.


example/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+set(EXAMPLES
+	readFromString
+	readFromStream
+	stringWrite
+	streamWrite
+   )
+
+add_definitions(-D_GLIBCXX_USE_CXX11_ABI)
+set_property(DIRECTORY PROPERTY COMPILE_OPTIONS $(EXTRA_CXX_FLAGS))
```
ferdnyc: Oh, this actually still has the same bug — that should be `${EXTRA_CXX_FLAGS}`. Not sure what changed that let it get past the CI system.

dota17: Yep, you are right. I just found. thank you~

Issue #812, opened 2018-07-23
Reader::pushError() maybe has typeset error


lymslive:
Recently I'm reading and learning the source, and seam to find a error:

in file json_reader.cpp:861(Reader::pushError)
token.end_ = end_ + value.getOffsetLimit();

I'm confused, maybe it should be:
token.end_ = begin_ + value.getOffsetLimit();

And in json_reader.cpp:1876 (OurReader::pushError) seams to have the same error.

cdunn2001:
Not sure. (And I never liked the pushError stuff.) Try an example and see if anything is wrong.

lilinchao-git:
As lymslive mentioned above，there are obvious contradictions at line 843 and line 862 in function pushError().

baylesj:
Fixed in #930

PR #930, opened 2019-05-31
repair a typo error

dota17:
As the issue #812 mentioned before, I'm sure there are some typo error in the `Reader::pushError()` funtion,
there should be a consisency with other functions like :
```c++
620 bool Reader::decodeString(Token& token) { 
621   String decoded_string; 
622   if (!decodeString(token, decoded_string)) 
623     return false; 
624   Value decoded(decoded_string); 
625   currentValue().swapPayload(decoded); 
626   currentValue().setOffsetStart(token.start_ - begin_); 
627   currentValue().setOffsetLimit(token.end_ - begin_); 
628   return true; 
629 } 
```


baylesj:
LGTM. This definitely seems like the right fix, and it's tightly scoped. We just need someone with write access...

PR #932, opened 2019-06-03
cmake cleanup

oleurodecision:
Minor improvement proposals for CMake. 

Note that the version.h.in changes target the generation of version.h file like the one currently stored in include/json (apart from the broken lines.

If any change are needed, please let me know.


baylesj:
Is there any functional changes, or is it just cleanup? Looks mostly like using syntactically sweeter method calls and macros.

+1 from me.

oleurodecision:
unless mistakes, yes, should only be cleanup commit.

PR #933, opened 2019-06-03
version.h : wrong file was deployed, added required include path and

oleurodecision:
Delivered version.h was some "static" version.h included in the include/json folder.

One can argue about the location of the install command, please let me know if change is required.


include/CMakeLists.txt
```diff
@@ -1,2 +1,6 @@
 file(GLOB INCLUDE_FILES "json/*.h")
-install(FILES ${INCLUDE_FILES} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/json)
+install(FILES
```
baylesj: If we are no longer using include/json/version.h, which is generated by CMake, does that file need to be deleted now??

oleurodecision: Hi,
don't understand the question :
- using CMake for configuration/build/install makes the include/json/version.h file useless (and source of desynchronization), the file **is** deleted in the current PR
- using other tool than CMake (which one ?), question is open, but in any case the file *should* be removed (and an equivalent, auto-generated by the conf. tool should be created and delivered) as a source of desynch

have a nice (not too hot) day

PR #934, opened 2019-06-03
added cmake config version file for proper cmake delivery

oleurodecision:
For proper cmake configuration setup, one should also deliver some config-version file. This allows consumers to check for some minimal/required version when importing.

The version compatibility was arbitrarily set to SameMajorVersion, SameMinorVersion or ExactVersion should probably be some smarter choice.

Please let me know if any change is required.


baylesj:
We don't do much from our end to manage CMake (e.g. bots), but this looks right to me.

PR #935, opened 2019-06-03
Add missing classes to forwards.h

### LINKED ISSUE #904: forwards.h doesn't declare most useful classes
[`json/forwards.h`](https://github.com/open-source-parsers/jsoncpp/blob/master/include/json/forwards.h) forward declares the deprecated FastWriter and Reader, but not the non-deprecated replacements StreamWriter and CharReader.

abigailbunyan:
Fixes #904.

### DIFF
--- a/include/json/forwards.h
+++ b/include/json/forwards.h
@@ -13,11 +13,17 @@
 namespace Json {
 
 // writer.h
+class StreamWriter;
+class StreamWriterBuilder;
+class Writer;
 class FastWriter;
 class StyledWriter;
+class StyledStreamWriter;
 
 // reader.h
 class Reader;
+class CharReader;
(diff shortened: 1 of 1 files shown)


BillyDonahue:
@dota17 that's expected. None of `forwards.h` is strictly necessary. But it's there so that users have a way to get correct declarations of the public API entities without including the full header.

include/json/forwards.h
```diff
@@ -13,11 +13,17 @@
 namespace Json {
 
 // writer.h
+class StreamWriter;
```
baylesj: I don't like the order in the writer.h and reader.h files, it doesn't really make sense to me.

However, since these are ordered the same as the files are I am happy with this patch.

PR #937, opened 2019-06-05
Modify the inconsistency between Reader and OurReader. 

dota17:


src/lib_json/json_reader.cpp
```diff
@@ -894,6 +946,7 @@ OurFeatures OurFeatures::all() { return {}; }
 // ////////////////////////////////
 
 // exact copy of Reader, renamed to OurReader
+// used by CharReader CharReaderBuilder
```
baylesj: Nit: missing "and" or comma, e.g.
```
// used by CharReader and CharReaderBuilder
```

include/json/reader.h
```diff
@@ -200,7 +203,7 @@ class JSON_API Reader {
   bool readCStyleComment();
   bool readCppStyleComment();
   bool readString();
-  void readNumber();
+  bool readNumber(bool checkInf);
```
baylesj: Any tests for this code? How do you know that it works?

src/lib_json/json_reader.cpp
```diff
@@ -193,19 +193,43 @@ bool Reader::readValue() {
...
     currentValue().setOffsetLimit(token.end_ - begin_);
-  } break;
+  } 
+  break;
   case tokenNull: {
     Value v;
     currentValue().swapPayload(v);
     currentValue().setOffsetStart(token.start_ - begin_);
     currentValue().setOffsetLimit(token.end_ - begin_);
-  } break;
```
baylesj: Nit: You should run clang format on this patch. We prefer `break` on the same like as the closing bracket of a scoped switch statement, so this line should be reverted and the following lines updated. 

If you haven't used clang format before, I added a section to our CONTRIBUTING.md documentation, and you can use the following command, assuming you are using lib type shared:

```
ninja -C build-shared clang-format
```

include/json/reader.h
dota17: The code should have been tested in `class OurReader`, it has changed a lot after exacting copy of Reader, renamed to OurReader, I just want to keep them consistent. Or we can just remove the comment `exact copy of Reader, renamed to OurReader`, because it is no longer accurate, and can be misleading.

src/lib_json/json_reader.cpp
baylesj: If you run clang-format on this patch, we can work on getting it checked in.

include/json/reader.h
baylesj: +1 on removing the comment. We should also work on getting all the test coverage on OurReader. Long term I would like to delete the Reader and Writer classes (probably as part of 2.0?)

dota17: Noticed that you have removed that comment, so do we still need this PR? And yes, delete the deprecated methods can keep the code clean.

PR #940, opened 2019-06-06
Fix JSON_USE_EXCEPTION=0 use case

baylesj:
This patch fixes the JSON_USE_EXCEPTION flag. Currently, due to the
throwRuntimeError and throwLogicError methods implemented in json_value,
even if JSON_USE_EXCEPTION is set to 0 jsoncpp will still throw. This
breaks integration into projects with -fno-exceptions set, such as
Chromium.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -50,6 +50,7 @@
  */
 namespace Json {
 
+#if JSON_USE_EXCEPTION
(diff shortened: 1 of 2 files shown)


milabs:
OMG, `abort` instead of error code that's hilarious IRLOL

Issue #774, opened 2018-05-14
some bug when using jsonReader.parse with some special string


linjunmian:
Hi ,I find a bug when using parse with some special string.My version is 1.8.1。My code is:
```cpp
//////////////////////
std::string str("->'-->`--><!--#set var=");
Json::Value body;
Json::Reader jsonReader;
bool succ = jsonReader.parse(str,body);
std::cout<<"succ = "<<succ<<"  is object = "<<body.isObject()<<std::endl;
//////////////////////
```
the out is

    succ = 1  is object = 0;

I think there is inconsistency between the return and the body.isobject.
But when I use the string like `("aaaaaaaaaaaa")`
the out is

    succ = 0  is object = 0;

So I think this inconsistency is only sent in some special circumstances.

cdunn2001:
Try `CharReader`. (See docs.) The deprecated `Reader` is less well defined.

dota17:
@cdunn2001 , The same inconsistency appeared on `CharReader` in my test. this may be a real bug.

dota17:
At first, I supossed no matter `CharReader` or `Reader`, when a string is beginning with `-`, then it will be recognized as a number, and as I  made some tests, the result were beyond my expectation, I'm sure there are some problems in `readNumber()` method and `readToken` method .  

 some result are here ctreated by `CharReader`:
```c++
the input strRes : -I
* Line 1, Column 1
  Syntax error: value, object or array expected.
isobject : 0
isString : 0
isNumber : 0
isDouble : 0
isArray : 0
```
```c++
the input strRes : -i
parse OK with newCharReader
isobject : 0
isString : 0
isNumber : 1
isDouble : 1
isArray : 0
```
```c++
the input strRes : -E
* Line 1, Column 1
  '-E' is not a number.
isobject : 0
isString : 0
isNumber : 0
isDouble : 0
isArray : 0
```

```c++
the input strRes : -e
* Line 1, Column 1
  '-e' is not a number.
isobject : 0
isString : 0
isNumber : 0
isDouble : 0
isArray : 0
```

```c++
the input strRes : -a
parse OK with newCharReader
isobject : 0
isString : 0
isNumber : 1
isDouble : 1
isArray : 0
```

```c++
the input strRes : -/[]
parse OK with newCharReader
isobject : 0
isString : 0
isNumber : 1
isDouble : 1
isArray : 0
```

```c++
the input strRes : -;
parse OK with newCharReader
isobject : 0
[...]

Issue #939, opened 2019-06-06
Some problems when using CharReaderBuilder with strictMode mode.


dota17:
As issue #774 have discussed, I used `CharReader` and `strictMode` in my code, but the result still shows that there are still some strange exceptions
Here is my code snippet：
```c++
std::string strRes = "{ \"a\": -; }";
int nLen = strRes.length();
const char *pStart = strRes.c_str();
cout << "the input strRes : " << strRes << endl;

Json::CharReaderBuilder jsreader;
jsreader.strictMode(&jsreader.settings_);
jsreader["strictRoot"] = false;
std::unique_ptr<Json::CharReader>const reader(jsreader.newCharReader());

if (!reader->parse(pStart, pStart + nLen, &root, &err)){
	cout << err << endl;
}
else
	cout << "parse OK with newCharReader" << endl;
	
cout << "isobject : " << root.isObject() << endl;
cout << "isString : " << root.isString() << endl;
cout << "isNumber : " << root.isNumeric() << endl;
cout << "isDouble : " << root.isDouble() << endl;
cout << "isArray : " << root.isArray() << endl;

return 0;
```
When I changed the `strRes`  to other values, the result is the same just like #774. This problem has been bothering me for a while. I discussed this in 774. But maybe it was because the issue was a long time ago, no one responded, so I opened a new one. Hope someone would like to help.
@cdunn2001 @hjmjohnson

dota17:
Sorry, I made a mistake.
Now, I don't think it is a bug. The return and the root.isobject (isint, isarray, ...) are irrelevant, so there is no consistency between them.
Some newer related mesages is in #774, I made new discussions.
I will close this issue.

PR #941, opened 2019-06-11
fix the issue #774 and #939

dota17:
fix issue #774 and #939

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1382,18 +1382,15 @@ bool OurReader::readCppStyleComment() {
 bool OurReader::readNumber(bool checkNeg) {
   const char* p = current_;
   char c = '0'; // stopgap for already consumed character  
-  if(checkNeg){
-    if (checkNeg && p != end_ && *p == 'I') {
+  if (checkNeg) {
+    if (p != end_ && *p == 'I') {   // negative Infinity
       current_ = ++p;
       return false;
-    }
-    else
-    {
+    } else {                         // just negative number.
       current_ = ++p;                // skip '-'
-      OurReader::readNumber(false);  // read the back part of the '-'
+      OurReader::readNumber(false);  // jump to the part after the '-'
     }
-  }
-  else{
+  } else {   // handling the part after '-' or just a positive number.
     // integral part
     while (c >= '0' && c <= '9')
       c = (current_ = p) < end_ ? *p++ : '\0';
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1381,16 +1381,16 @@ bool OurReader::readCppStyleComment() {
 // if checkNeg is true, read negative number
 bool OurReader::readNumber(bool checkNeg) {
   const char* p = current_;
-  char c = '0'; // stopgap for already consumed character  
+  char c = '0'; // stopgap for already consumed character
   if (checkNeg) {
-    if (p != end_ && *p == 'I') {   // negative Infinity
+    if (p != end_ && *p == 'I') { // negative Infinity
       current_ = ++p;
       return false;
-    } else {                         // just negative number.
-      current_ = ++p;                // skip '-'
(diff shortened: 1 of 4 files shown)


src/lib_json/json_reader.cpp
```diff
@@ -1378,30 +1378,39 @@ bool OurReader::readCppStyleComment() {
...
-
-bool OurReader::readNumber(bool checkInf) {
+// if checkNeg is true, read negative number
+bool OurReader::readNumber(bool checkNeg) {
   const char* p = current_;
-  if (checkInf && p != end_ && *p == 'I') {
-    current_ = ++p;
-    return false;
+  char c = '0'; // stopgap for already consumed character  
+  if(checkNeg){
```
baylesj: Nit: space around brackets. Do we use clang format or anything? We definitely should.

src/lib_json/json_reader.cpp
```diff
@@ -1378,30 +1378,39 @@ bool OurReader::readCppStyleComment() {
...
+    if (checkNeg && p != end_ && *p == 'I') {
```
baylesj: This is for checking against Infinity, right?  Using a flag called "checkNeg" to permit checking against Infinity is confusing. Plus, Infinity is not in the JSON spec, so do we support it but only if its capitalized??

src/lib_json/json_reader.cpp
```diff
@@ -1378,30 +1378,39 @@ bool OurReader::readCppStyleComment() {
...
+      current_ = ++p;
+      return false;
+    }
+    else
+    {
+      current_ = ++p;                // skip '-'
+      OurReader::readNumber(false);  // read the back part of the '-'
+    }
   }
-  char c = '0'; // stopgap for already consumed character
```
baylesj: Why move the `char c` declaration?  We don't use it in the checkNeg block.

src/lib_json/json_reader.cpp
```diff
@@ -1378,30 +1378,39 @@ bool OurReader::readCppStyleComment() {
...
+  char c = '0'; // stopgap for already consumed character  
+  if(checkNeg){
+    if (checkNeg && p != end_ && *p == 'I') {
+      current_ = ++p;
+      return false;
+    }
+    else
+    {
+      current_ = ++p;                // skip '-'
+      OurReader::readNumber(false);  // read the back part of the '-'
```
baylesj: Nit: I'd prefer an early return here instead of wrapping all of the parsing code below in an else.

src/lib_json/json_reader.cpp
dota17: This is awkward, it’s repeatedly checked here. And because there's a feature to allow some special floats, so, in some cases, we need to support this.

src/lib_json/json_reader.cpp
dota17: I move it outside of the checkNeg block.

src/lib_json/json_reader.cpp
dota17: I think I should explain a little bit why I modified it this way.
The original method **readNumber(bool checkInf)** looks like it can read number(positive or negtive) and check if it's `negtive Infinity` or not.  And in fact it read number positive or negtive that depend on a flag `checkInf`, this seems strange, right? 
Moreover, when `checkInf` is **true**, that's because it's in  `-` case, then, when `readNumber(true)` return flase, we think it's `-Infinity`, all other cases it return true, and we  think it's a negtive number, and set token's type to `tokenNumber`, But this is where the problem is. when it's in `-` case, and it's actually like `-i` or `->-->--><!--#set var=`  or something else that begin with  `-`, we still take it as a number, and token's type to `tokenNumber`. 
So I modify the method to **readNumber(bool checkNeg)** it looks more clear and unambiguous. it can read positive number or negtive number(which include negtive Infinity) by the flag **checkNeg**, and more important, it can selove the problem as I mentioned above. And the relative testcase I add has passed in my PC, so I think it really works.

BillyDonahue:
Description should have links to Issue #774 and Issue #939

PR #942, opened 2019-06-11
add a testcase for issue 774 and 939

dota17:


src/test_lib_json/main.cpp
```diff
@@ -2045,7 +2045,19 @@ JSONTEST_FIXTURE(CharReaderTest, parseWithStackLimit) {
...
 }
-
+JSONTEST_FIXTURE(CharReaderTest, issue774) {
+  Json::CharReaderBuilder b;
+  Json::CharReader* reader(b.newCharReader());
+  Json::String errs;
+  Json::Value root;
+  char const doc[] = "-i";
+  bool ok = reader->parse(doc, doc + std::strlen(doc), &root, &errs);
+  JSONTEST_ASSERT(!ok);
```
dota17: I think the CI build failed just because the `ok` is `true`, but actually it shouldn't. So I submit a PR(#941).

PR #947, opened 2019-06-21
Use fpclassify to convert real number to bool

gitpaladin:
According to JavaScript language, `0.0` and `NaN` are treated as `false`, others are treated as `true`. See 
 [JavaScript Booleans](https://www.w3schools.com/js/js_booleans.asp).

C++11 has already provided a function `fpclassify` to test a real number is zero or NaN. See [std::fpclassify](https://en.cppreference.com/w/cpp/numeric/math/fpclassify), so the unsafe comparison from `0.0f` should change to using this function.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -849,7 +849,7 @@ bool Value::asBool() const {
     return value_.uint_ ? true : false;
   case realValue:
     // According to JavaScript language zero or NaN is regarded as false
-    return std::fpclassify(value_.real_) != FP_ZERO && std::fpclassify(value_.real_) != FP_NAN
+      return std::fpclassify(value_.real_) != FP_ZERO && std::fpclassify(value_.real_) != FP_NAN;
   default:
     break;
   }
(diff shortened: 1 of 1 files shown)


src/lib_json/json_value.cpp
```diff
@@ -848,8 +848,8 @@ bool Value::asBool() const {
   case uintValue:
     return value_.uint_ ? true : false;
   case realValue:
-    // This is kind of strange. Not recommended.
-    return (value_.real_ != 0.0) ? true : false;
+    // According to JavaScript language zero or NaN is regarded as false
+      return std::fpclassify(value_.real_) != FP_ZERO && std::fpclassify(value_.real_) != FP_NAN;
```
baylesj: Nit: the standard library is unclear on the complexity of the `std::fpclassify` macro.

For readability as well as avoiding duplicate calls (unclear whether the second call would be compiled out), I would recommend placing the fpclassify in a single call on a separate line, e.g.

```
case realValue: {
    const auto value_classification = std::fpclassify(value_.real_);
    return (value_classification != FP_ZERO) && (value_classification != FP_NAN);
}
```

Issue #691, opened 2017-10-06
Insert item in array at specific ArrayIndex


rickyviking:
I couldn't find a method to insert a new value in an array at specific ArrayIndex. 
Say the equivalent (and expensive) method of `removeIndex()`.
Is there a particular reason or that could be added?

cdunn2001:
Yes, that could be added. Just submit a PR, with tests.

dota17:
Fix : #949

PR #949, opened 2019-06-21
add a new method to insert a new value in an array at specific index.

dota17:
fix issue #691.


src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,18 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
   }
   // JSONTEST_ASSERT_EQUAL(5, root["array"].size());
 }
-
+JSONTEST_FIXTURE(ValueTest, arrayIssue691) {
+  Json::Value array2_;
```
baylesj: Naming: no trailing underscore for local variables.

src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,18 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
...
+  array2_.append(1);
+  array2_.append(2);
+  array2_.append(3);
+  array2_.append(5);
+  ////use insert method
+  array2_.insert(3,4);
+  JSONTEST_ASSERT_EQUAL(Json::Value(3),array2_[2]);
```
baylesj: I would replace these checks with the following:

```cpp
  array2.append(10); // index 0
  array2.append(20); // index 1
  array2.append(30); // index 2
  array2.append(50); // index 3

  JSONTEST_ASSERT_EQUAL(Json::Value(50),array2_[3]);

  array2.insert(3, 40); // index 3

  // After updating, index 3 should be changed from 50 to 40, and 50 moved to index 4.
  JSONTEST_ASSERT_EQUAL(Json::Value(40),array2_[3]);
  JSONTEST_ASSERT_EQUAL(Json::Value(50),array2_[4]);

```

include/json/value.h
```diff
@@ -463,8 +463,9 @@ Json::Value obj_value(Json::objectValue); // {}
   ///
   /// Equivalent to jsonvalue[jsonvalue.size()] = value;
   Value& append(const Value& value);
+  /// \brief Insert value in array at specific index
+  bool insert(ArrayIndex index, const Value& newValue);
```
gitpaladin: A movable `insert` method should also be provided, which may declare as

```
bool insert(ArrayIndex index, Value&& newValue);
```

src/test_lib_json/main.cpp
dota17: Thanks a lot for your valuable suggestion.

include/json/value.h
dota17: Thank you for your advice. I'm not sure if it's necessary to use move assignment operator here.

dota17: Thank you gitpaladin, I'm gonna to add this today.

src/test_lib_json/main.cpp
dota17: Hi @baylesj, the build still failed, but it's ok in my PC. I don't know why.

baylesj: Are you sure you just haven't updated the patch? The Mac, Windows, and Linux builds are all failing with the same error:

```
../src/lib_json/json_value.cpp:1183:20: error: too few arguments to function call, single argument 'index' was not specified
  if(!isValidIndex()){
      ~~~~~~~~~~~~ ^
../src/lib_json/json_value.cpp:1119:1: note: 'isValidIndex' declared here
bool Value::isValidIndex(ArrayIndex index) const { return index < size(); }
^
1 error generated.
```

src/lib_json/json_value.cpp
```diff
@@ -1170,11 +1170,48 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
...
 Value& Value::append(const Value& value) { return (*this)[size()] = value; }
-
-Value& Value::append(Value&& value) {
-  return (*this)[size()] = std::move(value);
+/// \brief Insert value in array at specific index
+bool Value::insert(ArrayIndex index, const Value& newValue){
+  if(type() != arrayValue){
+    return false;
+  }
+  if(!isValidIndex()){
```
baylesj: Here's the compile failure:

```
../src/lib_json/json_value.cpp:1183:20: error: too few arguments to function call, single argument 'index' was not specified
  if(!isValidIndex()){
      ~~~~~~~~~~~~ ^
../src/lib_json/json_value.cpp:1119:1: note: 'isValidIndex' declared here
bool Value::isValidIndex(ArrayIndex index) const { return index < size(); }
^
1 error generated.
```

src/lib_json/json_value.cpp
```diff
@@ -1170,11 +1170,48 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
 #endif
 
 Value& Value::append(const Value& value) { return (*this)[size()] = value; }
-
-Value& Value::append(Value&& value) {
-  return (*this)[size()] = std::move(value);
+/// \brief Insert value in array at specific index
+bool Value::insert(ArrayIndex index, const Value& newValue){
+  if(type() != arrayValue){
```
baylesj: Run clang-format on this patch as well. LMK if you have any problems while trying to do that.

dota17: Finally, I configured the meson/ninja build environment and install clang-format. The result looks pretty good.
```c++
root@dggphicprc07936:~/json/jsoncpp-master/build-shared# meson test --no-rebuild --print-errorlogs
1/2 unittest_jsoncpp_test                   OK       0.02 s
2/2 unittest_jsontestrunner                 OK       0.87 s

Ok:                    2
Expected Fail:         0
Fail:                  0
Unexpected Pass:       0
Skipped:               0
Timeout:               0

```

src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,21 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
   }
   // JSONTEST_ASSERT_EQUAL(5, root["array"].size());
 }
+JSONTEST_FIXTURE(ValueTest, arrayIssue691) {
+  Json::Value array2;
+  array2.append(10);  // index 0
```
BillyDonahue: Suggestion: Remove the comment and append a hint-like string value like "index0" instead of 10.

src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,21 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
...
+  array2.append(20);  // index 1
+  array2.append(30);  // index 2
+  array2.append(50);  // index 3
 
+  JSONTEST_ASSERT_EQUAL(Json::Value(50),array2[3]);
+
+  array2.insert(3,40); // index 3
+  // After updating, index 3 should be changed from 50 to 40, and 50 moved to
```
BillyDonahue: that's clear from the assertions themselves and doesn't need extra commenting.

src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,21 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
...
   // JSONTEST_ASSERT_EQUAL(5, root["array"].size());
 }
+JSONTEST_FIXTURE(ValueTest, arrayIssue691) {
+  Json::Value array2;
+  array2.append(10);  // index 0
+  array2.append(20);  // index 1
+  array2.append(30);  // index 2
+  array2.append(50);  // index 3
 
+  JSONTEST_ASSERT_EQUAL(Json::Value(50),array2[3]);
```
BillyDonahue: We aren't testing that append works. This should come out.

src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,21 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
   }
   // JSONTEST_ASSERT_EQUAL(5, root["array"].size());
 }
+JSONTEST_FIXTURE(ValueTest, arrayIssue691) {
+  Json::Value array2;
```
BillyDonahue: why "array2"?

src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,21 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
...
+  array2.append(10);  // index 0
+  array2.append(20);  // index 1
+  array2.append(30);  // index 2
+  array2.append(50);  // index 3
 
+  JSONTEST_ASSERT_EQUAL(Json::Value(50),array2[3]);
+
+  array2.insert(3,40); // index 3
```
BillyDonahue: Need to test the insert of rvalues and lvalues to make sure both overloads are implemented correctly.

src/test_lib_json/main.cpp
```diff
@@ -307,7 +307,21 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
...
+  // After updating, index 3 should be changed from 50 to 40, and 50 moved to 
+  // index 4.
+  JSONTEST_ASSERT_EQUAL(Json::Value(40),array2[3]);
+  JSONTEST_ASSERT_EQUAL(Json::Value(50),array2[4]);
```
BillyDonahue: Should also make sure the preceding elements haven't been corrupted by the operation.
It's easy to just check the whole container. Maybe copy the array out into a std::vector and check for == against an expected std::vector value?

src/lib_json/json_value.cpp
baylesj: Awesome, thanks for figuring that out!

src/lib_json/json_value.cpp
```diff
@@ -1170,11 +1170,48 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
...
+    return false;
+  }
+  ArrayIndex oldsize = size();
+  resize(oldsize+1);
+  ArrayIndex length = size();
+  if(length != oldsize +1) {
+    return false;
+  }
+  for(ArrayIndex i = length ; i> index ; i--) {
+    (*this)[i] = (*this)[i-1];
```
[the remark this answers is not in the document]
dota17: There have array iterators on Value, there is no match for `operator+`(operand types are Json::Value::iterator and Json::ArrayIndex) when it comes to `begin() + length`.  Should we write a new operator+ overload?

src/lib_json/json_value.cpp
```diff
@@ -1170,11 +1170,48 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
...
+  if(type() != arrayValue) {
+    return false;
+  }
+  if(!isValidIndex(index)) {
+    return false;
+  }
+  ArrayIndex oldsize = size();
+  resize(oldsize+1);
+  ArrayIndex length = size();
+  if(length != oldsize +1) {
```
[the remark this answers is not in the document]
dota17: Nop, it has no return value. I'll remove it.

src/test_lib_json/main.cpp
```diff
@@ -2535,6 +2549,7 @@ int main(int argc, const char* argv[]) {
   JSONTEST_REGISTER_FIXTURE(runner, ValueTest, objects);
   JSONTEST_REGISTER_FIXTURE(runner, ValueTest, arrays);
   JSONTEST_REGISTER_FIXTURE(runner, ValueTest, arrayIssue252);
+  JSONTEST_REGISTER_FIXTURE(runner, ValueTest, arrayIssue691);
```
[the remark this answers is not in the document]
dota17: Uh, how about `arrayInsertionAtRandomIndex`?

src/test_lib_json/main.cpp
dota17: Well, the `array1_` has been used above,  so I used `array2_`, but consdering the rule of no trailing underscore for local variables  I changed it to `array2`.

src/lib_json/json_value.cpp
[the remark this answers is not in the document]
dota17: Hi Billy, I suppose the array in there is still `map`,not a real array(or vector), so it doesn't support rotate or reverse method.

src/test_lib_json/main.cpp
dota17: Uh, I think this is unnecessary, the `operator[ ]` is base on the map, we get the element by specifying the key value, which is unique, not by the index position.

src/lib_json/json_value.cpp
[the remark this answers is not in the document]
dota17: If there have any new idea, welcome to figure it out, we need it .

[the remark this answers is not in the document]
BillyDonahue: any forward iterator can be used with std::rotate. I think we should do whatever it takes to make our Array iterator a forward iterator, even if its operations aren't traditional or particularly cheap. Think of all the algorithms that would unlock for users. We don't need operator+ per se, but we do need to be able to form an iterator representing the midpoint of the rotate somehow or another.

src/test_lib_json/main.cpp
BillyDonahue: My point was that there's no other local array, so the array2 makes me wonder where array1 went. Just simpler array or arr would be better, I think.

src/test_lib_json/main.cpp
BillyDonahue: I don't follow, but the insert was shuffling and overwriting elements and I would like to check that we didn't overwrite some stray elements during the insert.

src/lib_json/json_value.cpp
```diff
@@ -1172,43 +1172,43 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
...
-  (*this)[index]= newValue;
+  (*this)[index] = newValue;
   return true;
 }
 #if JSON_HAS_RVALUE_REFERENCES
-Value& Value::append(Value&& value) { return (*this)[size()] = std::move(value); }
-bool Value::insert(ArrayIndex index, const Value&& newValue) {
-  if(type() != arrayValue) {
+Value& Value::append(Value&& value) {
+  return (*this)[size()] = std::move(value);
```
BillyDonahue: This syntax has to make a new uninitialized value and move-assign onto it.
We can do better with emplace semantics, I think?

src/lib_json/json_value.cpp
```diff
@@ -1172,43 +1172,43 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
...
+}
+bool Value::insert(ArrayIndex index, Value&& newValue) {
```
BillyDonahue: we should reuse the logic of the const Value& overload instead of duplicating it.

src/lib_json/json_value.cpp
dota17: The emplace syntax seems incredible. I‘d like to try it.

src/test_lib_json/main.cpp
dota17: Yep, it has been chaned to `array` in the former update.

src/lib_json/json_value.cpp
dota17: Due to the `append` method is not related to this PR, so in recent commit, I did not use this syntax.

baylesj: I'll update in a separate patch then.

src/lib_json/json_value.cpp
```diff
@@ -1168,13 +1168,35 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
...
+                      "in Json::Value::insert(ArrayIndex index, const Value& "
+                      "newValue): requires arrayValue");
+  return insert(index, Value(newValue));
+}
+bool Value::insert(ArrayIndex index, Value&& newValue) {
+  JSON_ASSERT_MESSAGE(type() == nullValue || type() == arrayValue,
+                      "in Json::Value::insert(ArrayIndex index, Value&& "
+                      "newValue): requires arrayValue");
+  ArrayIndex length = size();
+  if (index > length) {
```
baylesj: What if size is 10, and index is 20? Do we want to add nine blank entries?

src/test_lib_json/main.cpp
```diff
@@ -308,6 +311,72 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
   // JSONTEST_ASSERT_EQUAL(5, root["array"].size());
 }
 
+JSONTEST_FIXTURE(ValueTest, arrayInsertAtRandomIndex) {
+  Json::Value array;
+  Json::Value str0("index2");
```
baylesj: const.

src/test_lib_json/main.cpp
```diff
@@ -308,6 +311,72 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
...
+  Json::Value str1("index3");
+  array.append("index0"); // append rvalue
+  array.append("index1");
+  array.append(str0); // append lvalue
+
+  std::vector<Json::Value*> vec; // storage value address for checking
+  for (int i = 0; i < 3; i++) {
```
baylesj: Why have a temporary array at all? It's not clear to me what taking the address of each array index and checking them against themselves does. Are you testing that the array accessor doesn't mutate the list?

src/test_lib_json/main.cpp
```diff
@@ -308,6 +311,72 @@ JSONTEST_FIXTURE(ValueTest, arrayIssue252) {
...
+  vec.push_back(&array[5]);
+  // beyond max array size, it should be allowed to insert into its tail
+  array.insert(10, "index10");
+  JSONTEST_ASSERT_EQUAL(Json::Value("index3"), array[0]);
+  JSONTEST_ASSERT_EQUAL(Json::Value("index0"), array[1]);
+  JSONTEST_ASSERT_EQUAL(Json::Value("index4"), array[2]);
+  JSONTEST_ASSERT_EQUAL(Json::Value("index1"), array[3]);
+  JSONTEST_ASSERT_EQUAL(Json::Value("index2"), array[4]);
+  JSONTEST_ASSERT_EQUAL(Json::Value("index5"), array[5]);
+  JSONTEST_ASSERT_EQUAL(Json::Value("index10"), array[6]);
```
baylesj: What does it mean for index 10 to exist in an array size 7?

src/lib_json/json_value.cpp
dota17: I wish it will be appended to the end of array to avoid blank entries.

src/test_lib_json/main.cpp
dota17: Index 10 is out of range, then append it to the tail.

dota17: if it should change to throw a exception or error, i will change the testcase.

src/test_lib_json/main.cpp
dota17: Yes.
It is used to test the array accessor doesn't mutate the list after inserting a value and used to make sure that the key-value pair in value.map_ had not be changed.

src/lib_json/json_value.cpp
dota17: It is a question.
If the index is bigger than size, is it better to throw a exception or error?
I am not sure so that I invoke the append method. Maybe it should be removed and throw a exception.

dota17: > What if size is 10, and index is 20? Do we want to add nine blank entries?
@baylesj 
in this case, should we throw an exception or just return false?
Appending to the tail may be not good.

dota17: change to return false only, like the method `removeMember`
```
  if (type() != objectValue) {
    return false;
  }
```

src/test_lib_json/main.cpp
dota17: chang to return false only

baylesj:
Thanks for fixing this up, didn't see the latest revision.

src/lib_json/json_value.cpp
```diff
@@ -1165,6 +1166,28 @@ Value& Value::append(Value&& value) {
   return this->value_.map_->emplace(size(), std::move(value)).first->second;
 }
 
+/// \brief Insert value in array at specific index
+bool Value::insert(ArrayIndex index, const Value& newValue) {
+  JSON_ASSERT_MESSAGE(type() == nullValue || type() == arrayValue,
+                      "in Json::Value::insert(ArrayIndex index, const Value& "
+                      "newValue): requires arrayValue");
+  return insert(index, Value(newValue));
```
baylesj: If we always copy, why do we have a const& function parameter?

You can replace both of these methods with 

`Value::insert(ArrayIndex index, Value newValue)` then people can copy or move based on how they call it.

baylesj:
If you can fix the merge conflict, we are good to go.

dota17:
update code, fix the merge conflict  and clang-format.

cdunn2001:
I guess you need to rebase this yourself, and maybe squash some commits?

PR #950, opened 2019-06-21
modify some type error.

dota17:


include/json/value.h
```diff
@@ -539,17 +539,17 @@ Json::Value obj_value(Json::objectValue); // {}
   bool removeMember(const char* key, Value* removed);
   /** \brief Remove the named map member.
 
-      Update 'removed' iff removed.
+      Update 'removed' if removed.
```
baylesj: I don't think this is a typo. "iff" typically is math speak for "if and only if".

dota17: Uh, I understand now. thanks a lot. I'll close this PR.

Issue #917, opened 2019-05-05
Project file meson version update


eirnym:
Port explicitly uses meson `0.41.1` while features from `0.50.0` are used. Also used features deprecated in meson `0.48.0`

new features: `install arg in configure_file`
deprecated feature: `python3 module`

PR #952, opened 2019-06-24
Update meson build requirement

baylesj:
Currently, we have a build type warning due to listing a requirement for
meson build version that doesn't implement features we use in our build
file. The minimum meson build version required is actually 0.50.0, so
this PR updates our meson.build file to depend on 0.50.0.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@ project(
     'cpp_std=c++11',
     'warning_level=1'],
   license : 'Public Domain',
-  meson_version : '>= 0.41.1')
+  meson_version : '>= 0.50.0')
 
 jsoncpp_ver_arr = meson.project_version().split('.')
 jsoncpp_major_version = jsoncpp_ver_arr[0]


baylesj:
Fixes #917

cdunn2001:
We still need directions on how to build with Meson. All the cmake support is user-provided, not official. We actually use meson in TravisCI.

I'll add a PR for Meson directions, plus Meson updates ...

cdunn2001:
Hmmm. I don't actually see a merge commit for the README changes. Well, that's what I meant to comment on...

cdunn2001:
Oh, I see. That comment was moved to CONTRIBUTING.md. Ok.

I just updated Meson in my Homebrew, which took some time. Apparently we no longer support older Meson at all, which is fine...

cdunn2001:
I was getting this:
```
    DEPRECATION: Project targetting '>= 0.41.1' but tried to use feature depreca
    Build targets in project: 3
    WARNING: Deprecated features used:
     * 0.48.0: {'python3 module'}
```
So I did this:
```diff
-python = import('python3').find_python()
+python = find_program('python3', 'python')
```
But I no longer see that deprecation warning either way. Strange. It does seem to be deprecated:

* https://mesonbuild.com/Python-3-module.html

> This module is deprecated and replaced by the python module.

Issue #920, opened 2019-05-16
error arm android


manang:
Hi,
I'm trying to compile jsoncpp for android-arm, with clang as compiler.
I have the following error:
`[ 75%] Building CXX object src/lib_json/CMakeFiles/jsoncpp_lib.dir/json_writer.cpp.o
/Users/angelomantellini/android_fdio/src/jsoncpp/src/lib_json/json_reader.cpp:1909:50: error: implicit
      conversion loses integer precision: 'Json::Value::UInt64' (aka 'unsigned long long') to 'size_t'
      (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
  features.stackLimit_ = settings_["stackLimit"].asUInt64();`

do you know how I can solve?
Thank you very much

dota17:
It shows that you assigned a `unsigned long long int ` value to a `unsigned int` value, the size of former is longer than the later, so the complier will drop integer precision.
A solution is :  
add an explicit cast to the size_t type variable :  `result = (unsigned long long) size_t`
Here are some references:
[Is size_t is always unsigned int [duplicate]](https://stackoverflow.com/questions/33585094/is-size-t-is-always-unsigned-int)
[ARM Compiler toolchain Compiler Reference](http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0491c/Babfcgfc.html)

baylesj:
dota17 is correct that it can be fixed with a cast, however that's not the best fix--we shouldn't be downgrading a 64-bit int without any typing checks. I think 32-bits is plenty wide for this feature anyway, so I have posted a PR here:

https://github.com/open-source-parsers/jsoncpp/pull/954

baylesj:
Closing as fixed.

PR #954, opened 2019-06-24
Issue 920: Fix android build with casting fix

baylesj:
This patch removes an unchecked conversion from a 64bit wide type
to a 32bit wide type, fixing a compile error on some platforms.

Issue #920


baylesj:
@manang Does this fix your issue?

manang:
Hi @baylesj 
Now I have a warning on the same line, so I think that the workaround in the master branch is in the cmake file.

Thanks for your patch.
Angelo

baylesj:
Angelo, feel free to submit a patch with the CMake fix and I can review it.

Cheers,

Jordan

Issue #909, opened 2019-04-17
armhf build failed


ghost:
### Seems the new github code broken the default build:

$ git clone https://github.com/open-source-parsers/jsoncpp.git jsoncpp
$ cd jsoncpp
$ mkdir -p build/release
$ cd build/release
ubuntu@tegra-ubuntu:/opt/app/jsoncpp/build/release$ cmake -DCMAKE_BUILD_TYPE=release -DBUILD_STATIC_LIBS=ON -DBUILD_SHARED_LIBS=OFF -DARCHIVE_INSTALL_DIR=. -G "Unix Makefiles" ../..
-- JsonCpp Version: 1.9.0
CMake Error at src/lib_json/CMakeLists.txt:89 (target_compile_features):
  target_compile_features specified unknown feature "cxx_std_11" for target
  "jsoncpp_lib".

-- Configuring incomplete, errors occurred!
See also "/opt/app/jsoncpp/build/release/CMakeFiles/CMakeOutput.log".
ubuntu@tegra-ubuntu:/opt/app/jsoncpp/build/release$ 

ubuntu@tegra-ubuntu:~$ uname -a
Linux tegra-ubuntu 3.10.96-tegra #1 SMP PREEMPT Wed Nov 9 19:42:57 PST 2016 aarch64 aarch64 aarch64 GNU/Linux
ubuntu@tegra-ubuntu:~$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04 LTS
Release:        16.04
Codename:       xenial

cdunn2001:
If you can find a specific commit which needs to be reverted, we could do that quickly.

Or are you saying that C++11 is not available on armhf?

The cmakefiles are entirely user-supported. We maintain only Meson/Ninja. Would Meson work for you?

JinmingLv:
I saw the cmake version is 3.5.1, maybe cmake is too old, I think. between C++11 is supported on my environment, because when I remove cxx_std_11 flag in CMakeLists.txt, the build can success.

$ cmake --version
cmake version 3.5.1

CMake suite maintained and supported by Kitware (kitware.com/cmake).

cdunn2001:
If it works without C++11, push a PR to suppress that. We'll cut a release before we turn it on again.

cdunn2001:
We can use at least c++0x, right?

JinmingLv:
[...]

PR #963, opened 2019-06-28
Update minimum CMake version requirement

baylesj:
CMake added support for `cxx_std_11` in version 3.8. This patch updates our minimum requirement to version 3.8.


baylesj:
Fixes #909

Issue #487, opened 2016-06-22
asEnum method


dan-ryan:
Can we get this helpful method added? 
`json[KEY].asEnum<EnumClass>();`

It's less verbose compared to:
`static_cast<EnumClass>(json[KEY].asInt())`

cdunn2001:
Seems reasonable, assuming it compiles everywhere. Submit a PR, and be aware that if any fairly recent compilers barf, we'll have to revert or hide (conditionally compile) that method.

On the other hand, why not just create a global function yourself? I think a repo of such helpers would be ... helpful.

    asEnum<EnumClass>(json[KEY])

zammbi:
I was thinking to put it in our JsonHelper class, but thought this would be cleaner and better supported in the library itself. 
I wouldn't know where to start for a PR. But I'm guessing the code be something like:
```
template<typename T, typename std::enable_if<std::is_enum<T>::value>::type* = nullptr>
	T toEnum() {
		return static_cast <T>(asInt());
	}
```

BillyDonahue:
Bjarne said that the static_cast and reinterpret_cast are deliberately ugly in C++.
So you have to think when you see them. I think that holds here too.
Yes you have to write an application-level static_cast to get an enum out of a json numeric, but really I think that's probably for the best. We should be conservative about adding members to Json::Value which has an ABI compatibility concern.

baylesj:
Agreed, thanks for taking a look. Implementing the patch has convinced me this is a bad idea.

Less verbosity != better, often times it is worse if it hides complexity that shouldn't be hidden.

PR #966, opened 2019-07-01
Add new "asEnum" template method

baylesj:
This patch is a solution to #487, as request to add an "asEnum" template
helper for integral type Json::Value instances. Includes tests.

Issue: #487

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -296,14 +296,14 @@ public:
     To create an empty array, pass arrayValue.
     To create an empty object, pass objectValue.
     Another Value can then be set to this one by assignment.
-This is useful since clear() and resize() will not alter types.
+    This is useful since clear() and resize() will not alter types.
 
     Examples:
-\code
-Json::Value null_value; // null
-Json::Value arr_value(Json::arrayValue); // []
-Json::Value obj_value(Json::objectValue); // {}
-\endcode
+        \code
+        Json::Value null_value; // null
+        Json::Value arr_value(Json::arrayValue); // []
+        Json::Value obj_value(Json::objectValue); // {}
+        \endcode
   */
   Value(ValueType type = nullValue);
   Value(Int value);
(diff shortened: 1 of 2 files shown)


baylesj:
@BillyDonahue, @hjmjohnson, I have mixed feelings on whether this patch adds value. Thoughts? I want to resolve #487 either with some variation of this patch or as "Won't Fix."

include/json/value.h
```diff
@@ -392,6 +392,19 @@ Json::Value obj_value(Json::objectValue); // {}
...
   bool asBool() const;
 
+  template <typename T,
+            typename std::enable_if<std::is_enum<T>::value>::type* = nullptr>
+  T asEnum() {
+    switch (type()) {
+    case intValue:
+      return static_cast<T>(asLargestInt());
+    case uintValue:
+      return static_cast<T>(asLargestUInt());
```
BillyDonahue: We don't really know that the LargestInt can be stored by enum T. We also don't know what what that would mean as an enum T. It could be invalid or have an unintended meaning. I think we should probably bow out and leave the conversion to enums to application-layer logic.

Now one way to do that would be to implement an ADL function protocol or some other kind of extension hook so that the enum maintainers can TELL jsoncpp what the appropriate value should be, but I don't think we've supported any other kinds of user-defined types this way, so I wouldn't want to start now.

include/json/value.h
```diff
@@ -392,6 +392,19 @@ Json::Value obj_value(Json::objectValue); // {}
   double asDouble() const;
   bool asBool() const;
 
+  template <typename T,
+            typename std::enable_if<std::is_enum<T>::value>::type* = nullptr>
+  T asEnum() {
```
BillyDonahue: json doesn't know what an enum is.
I feel this an application-level issue.

baylesj:
Agreed, Billy. I'll close this and the related issue, I'm glad we took a look at it...

PR #967, opened 2019-07-02
pop the root node after readValue()

dota17:
We should pop the root node after readValue, although there is no problem when we don't care it, because it's the node at the bottom of the stack,  but in order to maintain consistency, it's better to pop it out.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1049,6 +1049,7 @@ bool OurReader::parse(const char* beginDoc,
   nodes_.push(&root);
 
   bool successful = readValue();
+  nodes_.pop();
   Token token;
   skipCommentTokens(token);
   if (features_.failIfExtra_) {


src/lib_json/json_reader.cpp
```diff
@@ -1049,6 +1049,7 @@ bool OurReader::parse(const char* beginDoc,
   nodes_.push(&root);
 
   bool successful = readValue();
```
baylesj: Do we need to pop from nodes_ every time we call readValue? If readValue() reads from nodes_ as a side effect, why not pop inside of the readValue method??

dota17: The inside of the readValue method do `push-pop` operation when it meet `readObject` and `readArray` case, It's a recursive call, and when it goes back to the origin(root) node, we'd better to pop it(or release it), which can be consistent in logic. and that's my little opinoin.

BillyDonahue: I agree with this change.
After a call to reader.parse(Value& root, ....) is over, the reader shouldn't be holding on to the address of that root reference in its internal std::stack<Value*> _nodes member. It's an invalid pointer after the parse exits, as the caller has no obligation to keep it around. The next time we call parse() the nodes_ is cleared by a slow one-by-one pop loop, so luckily it does no harm, but still, why retain the sloppy pointer hygeine?

BillyDonahue: @baylesj, we have internal members readObject, readArray, and readValue. None take arguments or return a value other than bool. They operate by doing an in-place modification of whatever is on the top of the nodes_ stack, and they are all mutually recursive to accept arbitrary data structure shapes. It could have been different, but this is how it is. So the top-level caller has to push a nodes_.top() for them to operate on, and to be responsible, should pop them when the parse is finished.

We should probably go further and make sure parse() clears the nodes_ even if parse() throws.

baylesj: https://github.com/open-source-parsers/jsoncpp/issues/977

PR #968, opened 2019-07-02
Adjust some code's position.

dota17:


src/lib_json/json_reader.cpp
```diff
@@ -1455,18 +1455,18 @@ bool OurReader::readObject(Token& token) {
     } else {
       break;
     }
-
-    Token colon;
-    if (!readToken(colon) || colon.type_ != tokenMemberSeparator) {
-      return addErrorAndRecover("Missing ':' after object member name", colon,
-                                tokenObjectEnd);
-    }
     if (name.length() >= (1U << 30))
```
dota17: This part of code should follow tightly with `tokenName.type_ == tokenString)`

PR #969, opened 2019-07-02
Change to throw macro, fix JSON_USE_EXCEPTION=0

baylesj:
Currently, if exceptions are disabled we throw instead, even though that
probably isn't desirable. This PR switches to using an early return,
which will typically cause methods to return false. Callers are
responsible for any cleanup.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -24,19 +24,6 @@
 #include <cpptl/forwards.h>
 #endif
 
-// Conditional NORETURN attribute on the throw functions would:
-// a) suppress false positives from static code analysis
-// b) possibly improve optimization opportunities.
-#if !defined(JSONCPP_NORETURN)
-#if defined(_MSC_VER)
-#define JSONCPP_NORETURN __declspec(noreturn)
-#elif defined(__GNUC__)
-#define JSONCPP_NORETURN __attribute__((__noreturn__))
-#else
-#define JSONCPP_NORETURN
-#endif
-#endif
-
 // Disable warning C4251: <data member>: <type> needs to have dll-interface to
 // be used by...
 #if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING)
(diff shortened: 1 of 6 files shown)


baylesj:
@mfoltzgoogle

baylesj:
I'm not married to this PR, would love our thoughts on whether aborting is too aggressive.

sbc100:
Actually I think until we have an actual error handling path then `abort()` is probably best substitute we have for `throw`.   its probably better to abort that do something unexpected/unhandled, at least from a security POV.

include/json/value.h
```diff
@@ -89,9 +76,17 @@ class JSON_API LogicError : public Exception {
 #endif
 
 /// used internally
-JSONCPP_NORETURN void throwRuntimeError(String const& msg);
-/// used internally
-JSONCPP_NORETURN void throwLogicError(String const& msg);
+#if JSON_USE_EXCEPTION
```
mfoltzgoogle: Should you avoid double declarations by guarding with:

#if !defined(THROW_RUNTIME_ERROR) && !defined(THROW_LOGIC_ERROR)

baylesj:
Aborting this PR, we can take a larger look in the following bug:

#972

include/json/value.h
```diff
@@ -89,9 +76,17 @@ class JSON_API LogicError : public Exception {
...
+#define THROW_RUNTIME_ERROR(msg_ref)                                           \
+  { throw RuntimeError(msg_ref); }
+#define THROW_LOGIC_ERROR(msg_ref)                                             \
+  { throw LogicError(msg_ref); }
+#else
+#define THROW_RUNTIME_ERROR(msg_ref)                                           \
```
mfoltzgoogle: It seems like abort() should be called in these macros for equivalence with the existing throwRuntimeError/throwLogicError?

For example JSON_ASSERT above will no longer abort() with this change.

include/json/value.h
```diff
@@ -89,9 +76,17 @@ class JSON_API LogicError : public Exception {
...
+  { return {}; }
```
mfoltzgoogle: I guess there's no logging facility, so if !JSON_USE_EXCEPTION the message gets swallowed.  Too bad.

PR #974, opened 2019-07-08
Add Definition

2mhk:
Add three definition : CheckAndGetMember,CheckIfFalse,CheckEnd.

It can replase
            if (root.isMember("GeneralInfo"))
            {
                Json::Value GeneralInfo = root["GeneralInfo"];
                ...
            }
            else
            {
                m_GeneralInfo.Clear();
            }
to
            CheckAndGetMember(root, GeneralInfo)
            {//this brackets is for visual studio's Typesetting
                ...
            }
            CheckIfFalse
                m_GeneralInfo.Clear();
            CheckEnd

The following code is an example.When a project has many keys,it may be useful and readable

    if (reader.parse(m_jsonbuffer, root))
    {
        try
        {
            CheckAndGetMember(root, GeneralInfo)
            {
                CheckAndGetMember(GeneralInfo, Company)
                    m_GeneralInfo.company = Company.asString();
                CheckIfFalse
                    m_GeneralInfo.company = "";
                CheckEnd;

                CheckAndGetMember(GeneralInfo, Department)
                    m_GeneralInfo.department = Department.asString();
                CheckIfFalse
                    m_GeneralInfo.department = "";
                CheckEnd;
            }
            CheckIfFalse
                m_GeneralInfo.Clear();
            CheckEnd
        }
        catch (...)
        {
            return false;
        }
    }
    else
    {
        return false;
    }

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -25,6 +25,14 @@
 
 #pragma pack(push, 8)
 
+#define CheckAndGetMember(value, key)\
+        if (value.isMember(#key)){\
+            Json::Value key = value[#key];
+
+#define CheckIfFalse }else{\
+
+#define CheckEnd }
+
 namespace Json {
 
 /** \brief Unserialize a <a HREF="http://www.json.org">JSON</a> document into a


baylesj:
Generally we don't want to use the preprocessor for shortening logic in this fashion, it hinders readability and makes people new to the code have to learn a lot of different shorthands for the same thing. For example, if we used this pattern throughout the reader we could have dozens of different #defines that are all resolved to `}`. See https://foonathan.net/blog/2017/05/08/preprocessor.html for a good explanation.

I would be open to including more inline functions for ease-of-use, but generally I will strongly discourage patches using preprocessor directives unless absolutely necessary.

Issue #979, opened 2019-07-10
Json::parseFromStream expects std::string in include/json/reader.h


Dfenyi:
The definition of "Json::parseFromStream" expects a std::string* in file include/json/reader.h:378.
The definition of "Json::parseFromStream" expects a Json::String* in file src/lib_json/json_reader.cpp:1992.

baylesj:
Good catch, I'll push a fix.

PR #980, opened 2019-07-11
#979 Fix parseFromStream definition

baylesj:
This patch fixes issue #979, where the parseFromStream definition in
the header is different from the implementation.


baylesj:
Probably the smallest patch I will ever get to submit to this repository...

Issue #985, opened 2019-07-14
Poor comment formatting in the headers.


BillyDonahue:
The layout of in-header comments is in some trouble, particularly reader.h but a little in value.h as well. They seem to be indented and flowed incorrectly. Many lines are missing leading asterisks. There's no clang-format for comments, so we'll have to fix it by hand.

baylesj:
Closed fixed by #987

PR #986, opened 2019-07-14
Improve code comment formatting (Issue #985)

BillyDonahue:


src/lib_json/json_value.cpp
```diff
@@ -235,9 +235,7 @@ LogicError::LogicError(String const& msg) : Exception(msg) {}
 [[noreturn]] void throwRuntimeError(String const& msg) {
   throw RuntimeError(msg);
 }
-[[noreturn]] void throwLogicError(String const& msg) {
-  throw LogicError(msg);
-}
+[[noreturn]] void throwLogicError(String const& msg) { throw LogicError(msg); }
```
BillyDonahue: clang-format did this. We should keep up with clang-format.

baylesj: Thanks, you are right.

Issue #988, opened 2019-07-17
Cannot build with python nor MESON


rafak360:
**Windows 7 64-bit
Tested on Python 2.6 32-bit and 64-bit same error output**

```
C:\jsoncpp-master>python amalgamate.py
Amalgamating header...
Traceback (most recent call last):
  File "amalgamate.py", line 155, in <module>
    main()
  File "amalgamate.py", line 147, in main
    header_include_path=options.header_include_path)
  File "amalgamate.py", line 69, in amalgamate_source
    header.add_file("include/json/version.h")
  File "amalgamate.py", line 30, in add_file
    f = open(os.path.join(self.top_dir, relative_input_path), "rt")
IOError: [Errno 2] No such file or directory: 'C:\\jsoncpp-master\\include/json/
version.h'
```
I just followed these steps: [Amalgamated source](https://github.com/open-source-parsers/jsoncpp/wiki/Amalgamated) but it's not working for me, why?

**EDIT:**
I tried with MESON, but it didn't help:

```
Build started at 2019-07-17T18:38:44.657486
Main binary: c:\users\userexample\appdata\local\programs\python\python37\python.exe
Python system: Windows
The Meson build system
Version: 0.51.1
Source dir: C:\jsoncpp-master
Build dir: C:\jsoncpp-master\builddir
Build type: native build
Project name: jsoncpp
Project version: 1.9.0

meson.build:1:0: ERROR: Unknown compiler(s): ['icl', 'cl', 'c++', 'g++', 'clang++', 'clang-cl']
The follow exceptions were encountered:
Running "icl " gave "[WinError 2] The system cannot find the file specified"
Running "cl /?" gave "[WinError 2] The system cannot find the file specified"
Running "c++ --version" gave "[WinError 2] The system cannot find the file specified"
Running "g++ --version" gave "[WinError 2] The system cannot find the file specified"
Running "clang++ --version" gave "[WinError 2] The system cannot find the file specified"
Running "clang-cl /?" gave "[WinError 2] The system cannot find the file specified"
```

baylesj:
Looks like your meson isn't working because you don't have a compiler installed.

I have a fix for the amalgamate script, this has been an evolving issue recently. I will close this issue when that PR is checked in, thanks!

rafak360:
> Looks like your meson isn't working because you don't have a compiler installed.
> 
[...]

Issue #734, opened 2018-01-15
cmake/meson inconsistent soversion with 1.8.4


userwithuid:
[meson.build](https://github.com/open-source-parsers/jsoncpp/blob/ddabf50f72cf/meson.build#L56) uses 20
[CMakeLists.txt](https://github.com/open-source-parsers/jsoncpp/blob/ddabf50f72cf/CMakeLists.txt#L67) uses 19

cdunn2001:
This will probably happen again. Could you help us to include a single soversion into both?

baylesj:
Added some comments to hopefully avoid this in future releases.

dota17:
Closing. seems like this has been resolved.

PR #989, opened 2019-07-17
Cleanup versioning strategy

baylesj:
Currently, versioning is a mess. CMake and Meson have seperate build
version number storage locations, with no way of knowing you need to
have both. Plus, due to recent revisions the amalgamate script is broken
unless you build first, and may still be broken afterwards.

This PR fixes some issues with versioning, and adds comments clarifying
what has to be done when doing a release.

#988


baylesj:
Failures are due to us overusing resources. I guess I'll wait?

```
Submit files (using https)
   Using HTTP submit method
   Drop site:https://my.cdash.org/submit.php?project=jsoncpp
   Submission failed: Maximum number of builds reached for jsoncpp.  Contact webmaster@cdash.org for support.
   Submit failed, waiting 5 seconds...
   Retry submission: Attempt 1 of 3
```

baylesj:
#734

include/json/version.h
```diff
@@ -0,0 +1,28 @@
...
+#define JSONCPP_VERSION_STRING "1.9.2"
+#define JSONCPP_VERSION_MAJOR 1
+#define JSONCPP_VERSION_MINOR 9
+#define JSONCPP_VERSION_PATCH 2
+#define JSONCPP_VERSION_QUALIFIER
+#define JSONCPP_VERSION_HEXA                                                   \
+  ((JSONCPP_VERSION_MAJOR << 24) | (JSONCPP_VERSION_MINOR << 16) |             \
+   (JSONCPP_VERSION_PATCH << 8))
+
+#ifdef JSONCPP_USING_SECURE_MEMORY
```
BillyDonahue: What's this stuff about `JSONCPP_USING_SECURE_MEMORY` ?
It has nothing to do with version.
Actually JSONCPP_VERSION_HEXA and JSONCPP_VERSION_STRING are derivatives and should be elsewhere. I would like it if version.h were just the number tuple and that's about it.

CMakeLists.txt
```diff
@@ -63,11 +63,18 @@ if(NOT DEFINED CMAKE_BUILD_TYPE)
...
 project(JSONCPP
-        VERSION 1.9.0 # <major>[.<minor>[.<patch>[.<tweak>]]]
+        # Note: version must be updated in three places when doing a release. This
+        # annoying process ensures that amalgamate, CMake, and meson all report the
+        # correct version.
+        # 1. /meson.build
+        # 2. /include/json/version.h
+        # 3. /CMakeLists.txt
+        # IMPORTANT: also update the SOVERSION!!
+        VERSION 1.9.2 # <major>[.<minor>[.<patch>[.<tweak>]]]
```
BillyDonahue: CMake's file command can read a variable from a file.

    file (STRINGS "version.txt" JSONCPP_VERSION)
    file (STRINGS "soversion.txt" JSONCPP_SOVERSION)

I don't know how to do it in meson.
I guess a worst-case would be invoking cat.

    r = run_command('cat','version.txt')
    JSONCPP_VERSION=r.stdout()

    r = run_command('cat','soversion.txt')
    JSONCPP_SOVERSION=r.stdout()

I don't know meson. It feels like we should be able to do this though!

baylesj: I'll spend some time this week looking into that, I definitely agree with you.

PR #994, opened 2019-07-30
Modified fuzz.cpp and Added a dict

Google-Autofuzz:
The old fuzz.cpp didn't pass the afl engine test. Also, added a dictionary to increase the coverage of the test.


Google-Autofuzz:
This new fuzz target fix the afl engine test when we upstream to OSS-Fuzz. The old fuzz target has some issue when afl tested with the word "input". Thank you. Link: https://github.com/google/oss-fuzz/pull/2577#issuecomment-510217141

src/test_lib_json/fuzz.cpp
```diff
@@ -3,47 +3,31 @@
...
-  builder.settings_["allowComments_"] = hash_settings & (1 << 1);
-  builder.settings_["strictRoot_"] = hash_settings & (1 << 2);
-  builder.settings_["allowDroppedNullPlaceholders_"] = hash_settings & (1 << 3);
-  builder.settings_["allowNumericKeys_"] = hash_settings & (1 << 4);
-  builder.settings_["allowSingleQuotes_"] = hash_settings & (1 << 5);
-  builder.settings_["failIfExtra_"] = hash_settings & (1 << 6);
-  builder.settings_["rejectDupKeys_"] = hash_settings & (1 << 7);
-  builder.settings_["allowSpecialFloats_"] = hash_settings & (1 << 8);
+  // Write with StyledWriter
+  Json::StyledWriter styled_writer;
```
dota17: Should we use a method that is not deprecated?

Google-Autofuzz: Feel free to add an other fuzzer for the new method.
I think it's worthwhile using the *deprecated* one, because people are still using it.

Google-Autofuzz:
Friendly ping :)

src/test_lib_json/fuzz.cpp
```diff
@@ -3,47 +3,31 @@
...
-extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
-  Json::CharReaderBuilder builder;
-
-  if (size < sizeof(uint32_t)) {
+#include "fuzz.h"
+#include "json/config.h"
+#include "json/features.h"
+#include "json/reader.h"
+#include "json/value.h"
+#include "json/writer.h"
```
dota17: we can use #include "json/json.h" instead of "json/features.h", "json/reader.h", "json/value.h", "json/writer.h"

dota17:
When I run this fuzz test case, there will be `LeakSanitizer has encountered a fatal error`. See the detail in #1013. I supposed there is something wrong in this fuzz.cpp.

Google-Autofuzz:
Closing this PR because it isn't better than the previous fuzz.cpp

PR #997, opened 2019-07-31
Cleanup versioning strategy relanding (#989)

baylesj:
* Cleanup versioning strategy

Currently, versioning is a mess. CMake and Meson have seperate build
version number storage locations, with no way of knowing you need to
have both. Plus, due to recent revisions the amalgamate script is broken
unless you build first, and may still be broken afterwards.

This PR fixes some issues with versioning, and adds comments clarifying
what has to be done when doing a release.

* Run clang format

* Update SOVERSION....


baylesj:
I think I fixed the Travis issues. I'll wait on the job and make sure there are no remaining issues.

PR #999, opened 2019-08-09
Add cxx17 string_view operator[] for Json::Value

ObjatieGroba:
Is it possible to add string_view?
How can we turn it off on older c++ versions?
Can it be another define "JSON_USE_CXX17"? Or compiler defines?

Thank you

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -12,8 +12,12 @@
 #include <ostream>
 #include <sstream>
 #include <string>
-#include <string_view>
 #include <type_traits>
+
+#if __cplusplus >= 201703L
+#include <string_view>
+#define JSON_USE_STRING_VIEW 1
+#endif
 
 /// If defined, indicates that json library is embedded in CppTL library.
 //# define JSON_IN_CPPTL 1
@@ -142,7 +146,9 @@ using Allocator = typename std::conditional<JSONCPP_USING_SECURE_MEMORY,
                                             SecureAllocator<T>,
                                             std::allocator<T>>::type;
 using String = std::basic_string<char, std::char_traits<char>, Allocator<char>>;
+#ifdef JSON_USE_STRING_VIEW
 using StringView = std::basic_string_view<char, std::char_traits<char>>;
+#endif
 using IStringStream = std::basic_istringstream<String::value_type,
                                                String::traits_type,
                                                String::allocator_type>;
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -12,6 +12,7 @@
 #include <ostream>
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <type_traits>
 
 /// If defined, indicates that json library is embedded in CppTL library.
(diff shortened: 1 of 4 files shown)


include/json/config.h
```diff
@@ -12,6 +12,7 @@
 #include <ostream>
 #include <sstream>
 #include <string>
+#include <string_view>
```
dota17: It seems the system doesn't support C++17 Standard now, so it cannot open include file `string_view`
but if you like, you could use the __plusplus marco switch : 
```
#if __cplusplus >= 201703L 
#include <string_view>
#endif 
```

include/json/config.h
```diff
@@ -141,6 +142,7 @@ using Allocator = typename std::conditional<JSONCPP_USING_SECURE_MEMORY,
                                             SecureAllocator<T>,
                                             std::allocator<T>>::type;
 using String = std::basic_string<char, std::char_traits<char>, Allocator<char>>;
+using StringView = std::basic_string_view<char, std::char_traits<char>>;
```
dota17: Only when the __cplusplus >= 201703L, can this semantics work.

baylesj:
We would have to include abseil or move to C++17 for this, so I'm voting to close for now.

ObjatieGroba:
> Have to include abseil or move to c++17 for this

Why? It is just StringView. There are many cases where we can use it. What difference between const string ref and StringView?

dota17:
> > Have to include abseil or move to c++17 for this
> 
> Why? It is just StringView. There are many cases where we can use it. What difference between const string ref and StringView?

string_view can avoid unnecessary memory allocation. but it only applied in the new c++17 standard and abseil lib.

ObjatieGroba:
> it only applied in the new c++17 standard and abseil lib

What about people who use current master with c++17 standart? For them it is new standart and std::string_view can be applied because it contains pointer and size. Where the problem is?
For usage in operator[] const reference same as StringView. There are no memory allocation on this operator.
Can we add std::string_view to functions, where we do not care about memory ownership?
It is really necessary for people who use c++17 because them have to do smth like that
```
std::string_view sv;
Json::Value json;
json[std::string(sv)] = ...
```

Thank you

ObjatieGroba:
Maybe it is possible to add as default disabled feature?
To enable just define JSON_USE_STRING_VIEW

dota17:
I guess baylesj temporarily closed this PR because we haven't decided to use  abseil C++ library or C++17 Standard library. 
Currently, we should focus on how to migrate to C++14, so we should not rush to use new C++ standard, I think?
And there is a [link](https://www.modernescpp.com/index.php/c-17-avoid-copying-with-std-string-view) about the detail of std::string_view, hope it can help you.

BillyDonahue:
supporting std::string_view as a value key should be doable but it would have to be carefully controlled by a #if and we'd have to carefully control the overload set to avoid ambiguities with other string-like types we're using.

I don't think abseil has much to do with the decision though, unless I'm missing something.

aberaud:
As more projects move to C++17, I believe this becomes increasingly necessary.

Additionally to the proposals in this PR, it would be great to have 

`bool Json::Value::getString(std::string_view& out) const;`

To be able to read a string Json value without unneeded allocations + copies.

aberaud:
GCC is now moving to C++17 by default, and C++20/22 include additional support for `std::string_view`.

Other libraries like asio have included support for `std::string_view` for years now.

Could we please list issues with the current patch so they could be solved, so this could move forward and get merged?

PR #1002, opened 2019-08-15
jsoncpp: Define JSON_USE_INT64_DOUBLE_CONVERSION for clang as well.

m-gupta:
The current check to define JSON_USE_INT64_DOUBLE_CONVERSION
works for GCC but not clang.

Clang does define __GNUC__ but with a value 4 which misses
the check for >= 6.

This avoids the -Wimplicit-int-float-conversion warning
when jsoncpp is built with a recent version of clang.

Signed-off-by: Manoj Gupta <manojgupta@google.com>


m-gupta:
Thanks a lot!

PR #1004, opened 2019-08-16
Improving Code Readability : change variable's type

dota17:


dota17:
I think the Location typedef is expressed as "location" for the input json string/doc.
In my opinion, Location should be used for index, location, etc, and not for a string.
It doesn't mean that we should use const char everywhere.

PR #1012, opened 2019-08-23
Reinforce readToken function

dota17:
Add case "+" in readToken to allow it to read explicit positive number token.


dota17:
> What is the use case for this? "+" tokens are not part of ECMAScript, RFC 7159, or json.org grammars. Is someone wanting this

In some special usage scenarios, like the dial codes ` {"BOSTON" : +1-617}`, we do need to parse these cases, so we can use this case with the switch `allowSpecialFloats_`,  just like we deal with `NaN` or `Infinity`.

baylesj:
For dial codes, I would expect string representation, e.g.

"BOSTON": "+1-617"

Since 1-617 isn't a valid number anyway.

I still don't think we need this.

PR #1014, opened 2019-08-27
Do not allow tokenError tokens after input if failIfExtra is set.

res2k:
(This is basically PR #1001 plus some change suggestion from it's discussion.)

Currently when failIfExtra is set and strictRoot is not set,
OurReader::parse() will accept trailing non-whitespace after the JSON value
as long as the first token is not a valid JSON token. This commit changes
this to disallow any non-whitespace after the JSON value.

This commit also suppresses the "Extra non-whitespace after JSON value."
error message if parsing was aborted after another error.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1051,12 +1051,9 @@ bool OurReader::parse(const char* beginDoc,
   nodes_.pop();
   Token token;
   skipCommentTokens(token);
-  if (features_.failIfExtra_) {
-    if ((features_.strictRoot_ || token.type_ != tokenError) &&
-        token.type_ != tokenEndOfStream) {
-      addError("Extra non-whitespace after JSON value.", token);
-      return false;
-    }
+  if (features_.failIfExtra_ && (token.type_ != tokenEndOfStream)) {
+    addError("Extra non-whitespace after JSON value.", token);
+    return false;
   }
   if (collectComments_ && !commentsBefore_.empty())
     root.setComment(commentsBefore_, commentAfter);


dota17:
@res2k  Hmmm, the old PR #1001 should be closed?

src/lib_json/json_reader.cpp
```diff
@@ -1051,12 +1051,9 @@ bool OurReader::parse(const char* beginDoc,
   nodes_.pop();
   Token token;
   skipCommentTokens(token);
-  if (features_.failIfExtra_) {
-    if ((features_.strictRoot_ || token.type_ != tokenError) &&
-        token.type_ != tokenEndOfStream) {
-      addError("Extra non-whitespace after JSON value.", token);
-      return false;
-    }
+  if (features_.failIfExtra_ && (token.type_ != tokenEndOfStream)) {
```
baylesj: Where did the check for strict root and token type go? Don't we need all the checks still?

res2k: `(features_.strictRoot_ || token.type_ != tokenError)` caused certain inputs to be unexpectedly accepted if `failIfExtra_` is on and `strictRoot_` is off. For example, take the input `123 foo`: 123 is accepted, but `foo` produces a tokenError. So there was no failure, despite "extra" data being present. The original PR #1001 explains the problem.

Generally, `strictRoot_` is also checked in a later (beyond the diff), so that's still honored.

If you think there could be some input that's handles incorrectly after this change I can check whether the unit tests cover that (and extend those if necessary).

PR #1022, opened 2019-09-16
Fixup Json::Value append methods

baylesj:


src/lib_json/json_value.cpp
```diff
@@ -1157,10 +1156,14 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
 }
 #endif
 
-Value& Value::append(const Value& value) { return (*this)[size()] = value; }
-
+Value& Value::append(const Value& value) { return append(Value(value)); }
 Value& Value::append(Value&& value) {
-  return (*this)[size()] = std::move(value);
+  JSON_ASSERT_MESSAGE(type() == nullValue || type() == arrayValue,
```
baylesj: This is the only logic change. PTAL.

src/lib_json/json_value.cpp
```diff
@@ -1155,10 +1155,14 @@ Value const& Value::operator[](CppTL::ConstString const& key) const {
 }
 #endif
 
-Value& Value::append(const Value& value) { return (*this)[size()] = value; }
-
+Value& Value::append(const Value& value) { return append(Value(value)); }
```
dota17: Are we going to do parameter checking here, so that we don't have to jump to next function when we encounter an unexpected parameter type？

baylesj: Considering the move version of this method does parameter checking, I'd rather optimize for the good case and not have duplicate parameter checking between multiple methods.

PR #1031, opened 2019-09-18
Supplement the testcase for comparing the Array and Null

Vincent-LP:


src/test_lib_json/main.cpp
```diff
@@ -1449,6 +1451,7 @@ JSONTEST_FIXTURE(ValueTest, compareArray) {
   JSONTEST_ASSERT_PRED(checkIsLess(l2aArray, l2bArray));
   JSONTEST_ASSERT_PRED(checkIsEqual(emptyArray, Json::Value(emptyArray)));
   JSONTEST_ASSERT_PRED(checkIsEqual(l1aArray, Json::Value(l1aArray)));
+  JSONTEST_ASSERT_PRED(checkIsEqual(l2aArray, Json::Value(l2aArray)));
```
dota17: is this testcase necessary？ I think this is same as next testcase `checkIsEqual(l2bArray, Json::Value(l2bArray))`

PR #1033, opened 2019-09-23
clang-tidy fixes

neheb:


src/lib_json/json_value.cpp
```diff
@@ -1446,7 +1447,7 @@ String Value::Comments::get(CommentPlacement slot) const {
 
 void Value::Comments::set(CommentPlacement slot, String comment) {
   if (!ptr_) {
-    ptr_ = std::unique_ptr<Array>(new Array());
+    ptr_ = std::make_unique<Array>();
```
dota17: `std::make_unique` is introduced in C++14 Standard, so the compiler is reporting an error.

neheb: Removed

PR #1037, opened 2019-09-27
Add vcpkg installation instructions

grdowns:
JsonCpp is available as a port in vcpkg. Adding installation instructions will help users get started by providing a single command they can use to build JsonCpp and include it into their projects


baylesj:
Thanks for the commit! I'm happy to improve documentation.

PR #1041, opened 2019-10-06
Re-add JSONCPP_NORETURN replacing [[noreturn]]

es1x:
Restores Visual Studio 2013 compatibility.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -9,6 +9,18 @@
 #if !defined(JSON_IS_AMALGAMATION)
 #include "forwards.h"
 #endif // if !defined(JSON_IS_AMALGAMATION)
+
+// Conditional NORETURN attribute on the throw functions would:
+// a) suppress false positives from static code analysis
+// b) possibly improve optimization opportunities.
+#if !defined(JSONCPP_NORETURN)
+#if defined(_MSC_VER) && _MSC_VER == 1800
+#define JSONCPP_NORETURN __declspec(noreturn)
+#else
+#define JSONCPP_NORETURN [[noreturn]]
+#endif
+#endif
+
 #include <array>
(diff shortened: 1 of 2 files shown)


dota17:
Can you post more info?
I am wondering the errors or other messages about the Visual Studio 2013 compatibility.

es1x:
The `[[noreturn]]` clause causes the following compilation errors, apparently as a result of incomplete C++11 support:
Capture

dota17:
When I used VS 2013 to build, I got the same result.
The problem is that the newest code which include `[[noreturn]]` requires more C++11 features, but VS 2013 doesn't support C++11 attributes feature.
Related Solution Link: [C3409](https://stackoverflow.com/questions/32815196/c3409-empty-attribute-block-is-not-allowed-by-noreturn)

According to the git history, in the commit [408b466b](https://github.com/open-source-parsers/jsoncpp/commit/408b466b57f4cf0744c2ab3de1c700dc83fc24fc#diff-180360612c6b8c4ed830919bbb4dd459),  the test for Visual Studio 12 2013 had been removed.
Maybe there are some other places/commits that require more C++11 features.

And I found some places that we may need to change if we do not re-add JSONCPP_NORETURN.
etc, in include/json/config.h
```
#if defined(_MSC_VER) && _MSC_VER < 1800
#error                                                                         \
    "ERROR:  Visual Studio 12 (2013) with _MSC_VER=1800 is the oldest supported compiler with sufficient C++11 capabilities"
#endif
```
Maybe Visual Studio 12 (2013) with _MSC_VER=1800 is not the oldest anymore.

baylesj:
I think the value add of dropping a macro is not worth dropping support for your IDE. I'll look at adding more CI coverage.

PR #1045, opened 2019-10-08
Added emitUTF8 setting.

nicolaswilson:
This PR adds a new setting 'emitUTF8' which avoids unicode escaping the emitted JSON.


src/lib_json/json_writer.cpp
```diff
@@ -267,7 +267,7 @@ static String toHex16Bit(unsigned int x) {
   return result;
 }
 
-static String valueToQuotedStringN(const char* value, unsigned length) {
+static String valueToQuotedStringN(const char* value, unsigned length, bool emitUTF8 = false ) {
```
dota17: It's better to overload it instead of rewriting it, I guess?

src/lib_json/json_writer.cpp
```diff
@@ -313,21 +313,26 @@ static String valueToQuotedStringN(const char* value, unsigned length) {
...
-        result += toHex16Bit(cp);
-      } else { // codepoint is not in Basic Multilingual Plane
-               // convert to surrogate pair first
-        cp -= 0x10000;
-        result += "\\u";
-        result += toHex16Bit((cp >> 10) + 0xD800);
-        result += "\\u";
-        result += toHex16Bit((cp & 0x3FF) + 0xDC00);
+      if ( emitUTF8 ) {
+        result =+ *c;
```
dota17: `result =+ *c;` ？

dota17:
In addition, it's better to write some testcases to test this chaging.

src/lib_json/json_writer.cpp
nicolaswilson: It does result += *c; to write the UTF-8 directly, or if not 'emitUTF8' it goes into the block that does the unicode escaping.

src/lib_json/json_writer.cpp
nicolaswilson: I want to use the escaping of the tabs, newlines etc. in the first part of the function and then either write the UTF-8 characters or unicode escape them in the second part. Perhaps I should split the function into those separate pieces? What's your view?

nicolaswilson:
> In addition, it's better to write some testcases to test this change.

I'll add some tests this evening.

src/lib_json/json_writer.cpp
```diff
@@ -313,21 +313,26 @@ static String valueToQuotedStringN(const char* value, unsigned length) {
...
-        result += "\\u";
-        result += toHex16Bit((cp & 0x3FF) + 0xDC00);
+      if ( emitUTF8 ) {
+        result += *c;
+      }
+      else {
+        unsigned int cp = utf8ToCodepoint(c, end);
+        // don't escape non-control characters
+        // (short escape sequence are applied above)
+        if (cp < 0x80 && cp >= 0x20)
```
baylesj: Prefer constants over magic variables, e.g.

```
const unsigned int codepoint = utf8ToCodepoint(c, end);
const int FIRST_NON_CONTROL_CODEPOINT = 0x20;
const int LAST_NON_CONTROL_CODEPOINT = 0x7F;

if (FIRST_NON_CONTROL_CODEPOINT <= codepoint && codepoint <= LAST_NON_CONTROL_CODEPOINT) {
...
```


0x80 is the maximum value, right? If you are going for non-control characters, does DEL/delete (0x7f) count as a noncontrol character?

src/lib_json/json_writer.cpp
```diff
@@ -313,21 +313,26 @@ static String valueToQuotedStringN(const char* value, unsigned length) {
...
-        cp -= 0x10000;
-        result += "\\u";
-        result += toHex16Bit((cp >> 10) + 0xD800);
-        result += "\\u";
-        result += toHex16Bit((cp & 0x3FF) + 0xDC00);
+      if ( emitUTF8 ) {
+        result += *c;
+      }
+      else {
+        unsigned int cp = utf8ToCodepoint(c, end);
```
baylesj: Prefer const.

src/lib_json/json_writer.cpp
```diff
@@ -313,21 +313,26 @@ static String valueToQuotedStringN(const char* value, unsigned length) {
...
+        // don't escape non-control characters
+        // (short escape sequence are applied above)
+        if (cp < 0x80 && cp >= 0x20)
+          result += static_cast<char>(cp);
+        else if (cp < 0x10000) { // codepoint is in Basic Multilingual Plane
```
baylesj: Same here, prefer a constant.

Maybe `const int FIRST_SURROGATE_PAIR_CODEPOINT = 0x10000;`?

src/lib_json/json_writer.cpp
```diff
@@ -1159,6 +1167,7 @@ StreamWriter* StreamWriterBuilder::newStreamWriter() const {
   bool eyc = settings_["enableYAMLCompatibility"].asBool();
   bool dnp = settings_["dropNullPlaceholders"].asBool();
   bool usf = settings_["useSpecialFloats"].asBool();
+  bool emitUTF8 = settings_["emitUTF8"].asBool();
```
baylesj: prefer const. Bonus points if you make all the other settings_ bools const as well.

src/test_lib_json/main.cpp
```diff
@@ -1996,6 +1996,30 @@ JSONTEST_FIXTURE(StreamWriterTest, writeZeroes) {
...
+  // Create a Json value containing UTF-8 string with some chars that need escape (tab,newline).
+  Json::Value root;
+  root["test"] = "\t\n\xF0\x91\xA2\xA1\x3D\xC4\xB3\xF0\x9B\x84\x9B\xEF\xBD\xA7";
+
+  Json::StreamWriterBuilder b;
+
+  // Default settings - should be unicode escaped.
+  JSONTEST_ASSERT(Json::writeString(b, root) == "{\n\t\"test\" : \"\\t\\n\\ud806\\udca1=\\u0133\\ud82c\\udd1b\\uff67\"\n}");
+
+  // Enable UTF-8
```
baylesj: Nit: remove comments that just state what the line of code below them does, especifically the enable and disable UTF-8 comments.

src/lib_json/json_writer.cpp
baylesj: I would prefer the following solution:

```
static String valueToQuotedStringN(const char* value, unsigned int length) {
    valueToQuotedStringN(value, length, false);
}

static String valueToQuotedStringN(const char* value, unsigned int length, bool emitUTF8) {
...
```

memeda:
Great! Thanks！
How about change the `emitUTF8` to some other name? It is not clearly at the first glance for the `emit` word mean. How about `noUTF8Decode` or `toUnicode` (may be `toUnicodeFromUTF8` most clearly, but too long...) ?
Just a naive option, Thanks 🚀

src/lib_json/json_writer.cpp
```diff
@@ -313,21 +313,26 @@ static String valueToQuotedStringN(const char* value, unsigned length) {
...
-      } else { // codepoint is not in Basic Multilingual Plane
-               // convert to surrogate pair first
-        cp -= 0x10000;
-        result += "\\u";
-        result += toHex16Bit((cp >> 10) + 0xD800);
-        result += "\\u";
-        result += toHex16Bit((cp & 0x3FF) + 0xDC00);
+      if ( emitUTF8 ) {
+        result += *c;
+      }
```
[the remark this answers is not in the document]
nicolaswilson: Done.

src/lib_json/json_writer.cpp
nicolaswilson: I didn't develop this section, just the small 'if' clause above it. I've changed it to use constants, but can't answer your question about 0x7F.

src/lib_json/json_writer.cpp
nicolaswilson: As above.

src/lib_json/json_writer.cpp
nicolaswilson: As above.

src/lib_json/json_writer.cpp
nicolaswilson: Done.

src/test_lib_json/main.cpp
nicolaswilson: Done.

dota17:
Good for it.
Could you fix the clang-format error of travis-ci?Recently, we added clang format support to the travis bots.

src/lib_json/json_writer.cpp
cdunn2001: I'm glad that you're all considering binary compatibility. Thank you.

Yes, internal "statics" are hidden from the public API, so either way is fine with me.

This change is a nice idea.

Sublihim:
why didn't we add a description to `jsoncpp/include/json/writer.h`?

PR #1047, opened 2019-10-11
Fix fuzzer off by one error

baylesj:
Currently the fuzzer has an off by one error, as it passing a bad length
to the CharReader::parse method, resulting in a heap buffer overflow.

### DIFF
--- a/example/readFromString/readFromString.cpp
+++ b/example/readFromString/readFromString.cpp
@@ -2,8 +2,8 @@
 #include <iostream>
 /**
  * \brief Parse a raw string into Value object using the CharReaderBuilder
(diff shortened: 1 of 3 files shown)


cdunn2001:
@dota17 or @baylesj , could one of you become the "primary contact" for oss-fuzz?

* https://github.com/google/oss-fuzz/pull/575

dota17:
It seems that @baylesj had been added.

[/oss-fuzz/commit/54cc1468a3d8f](https://github.com/google/oss-fuzz/commit/54cc1468a3d8f84ec710914c4aaf793a64cda0ac#diff-d048999821c824e49b5f7772802ca2cf)

Issue #1043, opened 2019-10-08
Ignore trailing comma in objects and arrays


kimsey0:
**Is your feature request related to a problem? Please describe.**
I use the [new Windows Terminal](https://github.com/microsoft/terminal), which uses JsonCpp for parsing JSON configuration files, and currently requires those configuration files to be manually edited. It provides [a file with defaults](https://github.com/microsoft/terminal/blob/06bd7e22da183619b99b632e10e186d441b05e45/src/cascadia/TerminalApp/defaults.json), from which I copy paste lines with settings for individual profiles (represented as JSON objects) into my user configuration file, then change the default values. However, those setting lines end with a comma, so if I paste them at the end of an existing profile object and forget to remove the comma, JsonCpp will fail to parse the configuration file and the Windows Terminal will pop up with a "Failed to reload settings" dialog.

More generally, this problem occurs for both objects and arrays when manually constructing JSON files or copy-pasting objects from ECMAScript code, which ignores trailing commas in object and array literals.

**Describe the solution you'd like**
I want JsonCpp, either by default or with a configuration setting, to ignore one trailing comma in objects and arrays while deserializing. I don't expect JsonCpp to keep any reference of this trailing comma or to be able to recreate it while serializing.

**Describe alternatives you've considered**
The main alternative I see is to keep the current behavior and to strictly enforce the correctness of parsed JSON. However, JsonCpp already supports comments, which is non-standard, so supporting trailing commas doesn't seem at odds with the direction of the project.

Between the alternatives of ignoring a trailing comma by default or by configuration, this depends on whether JsonCpp is allowed to break backwards compatibility by parsing files that would previously have resulted in an syntax error. I don't know the project stance on this.

JsonCpp could optionally support retaining and serializing trailing commas, but I don't see any value in that, except being able to more faithfully reproduce the source when deserializing and reserializing a file.

**Additional context**
This feature was also requested back in 2006 [on Sourceforge](https://sourceforge.net/p/jsoncpp/feature-requests/4/). @cdunn2001 back then remarked:

> I see this is difficult the way readObject() and readArray() are written. Oh, well. It would be nice someday...

kimsey0:
Issue #723 would include this if implemented.

dota17:
I had test the setting `allowDroppedNullPlaceholders_` for this case, and I set `allowDroppedNullPlaceholders_` to `true`:

[...]

PR #1053, opened 2019-10-16
Number fixes

baylesj:
Found out today that our jsontestrunner hasn't been testing our CharReader, only the legacy Reader. I rushed out a fix for this, as part of cleaning up the logic for parsing numbers.


dota17:
+1 from me.

src/lib_json/json_reader.cpp
```diff
@@ -1587,8 +1582,8 @@ bool OurReader::decodeNumber(Token& token, Value& decoded) {
   if (isNegative) {
     // We use the same magnitude assumption here, just in case.
     const Value::UInt last_digit = value % 10;
-    decoded = -Value::LargestInt(value / 10) - last_digit;
-  } else if (value > Value::LargestUInt(Value::maxLargestInt)) {
+    decoded = -Value::LargestInt(value / 10) * 10 - last_digit;
+  } else if (value <= Value::LargestUInt(Value::maxLargestInt)) {
```
cdunn2001: Oh! This was a bug? Good catch.

dota17: Just wonder the bug, is there a example code to reproduce the bug?

src/jsontestrunner/main.cpp
```diff
@@ -126,15 +127,39 @@ static int parseAndSaveValueTree(const Json::String& input,
...
-                                 Json::Value* root) {
-  Json::Reader reader(features);
-  bool parsingSuccessful =
-      reader.parse(input.data(), input.data() + input.size(), *root);
-  if (!parsingSuccessful) {
-    printf("Failed to parse %s file: \n%s\n", kind.c_str(),
-           reader.getFormattedErrorMessages().c_str());
-    return 1;
+                                 Json::Value* root, bool use_legacy) {
+  if (!use_legacy) {
```
dota17: Just like @kimsey0 mentioned in #1043 

If `CharReader` has some new features that `Reader` doesn't have.There will be some questions if we do not differentiate the test/data between `CharReader` and `Reader`

baylesj: Thanks for bringing this to my attention. My suggested fix in the new revision to to prepend "legacy_" to any tests that should also run on the legacy `Reader` class. New tests that break compat should not start with "legacy_".

cdunn2001:
> Found out today that our jsontestrunner hasn't been testing our CharReader, only the legacy Reader.

I think we should prioritize that over all other work. And as implies @kimsey0 says in #1062, we really should separate the old Reader tests from CharReader tests, so we can alter them independently. We are more limited in the changes we can make to old Reader.

cdunn2001:
I think this only needs clang-format.

kimsey0:
@baylesj: Would you clang-format this?

dota17:
@kimsey0 Done.

@baylesj if you don't need branch `number-fixes` anymore, please delete it.

PR #1055, opened 2019-10-16
update pushError for Reader and CharReader

### LINKED ISSUE #1054: Deprecate and remove pushError()
Recently, I am supplementing the test case for pushError() (src/lib_json/json_reader.cpp) and find it is a public and isolated interface. I couldn’t match it with a suitable test scenario.

By searching the related records of pusherror(), I found this interface was introduced into jsoncpp in #57 and #58. After that, @cdunn2001 reverted this commit since 0.8.0 version. But this pusherror() still exicted in the master branch and influence the later version. 

In short, this public interface is a useless and harmless addition to jsoncpp. The submitter was not opposed to remove pusherror(). Maybe we could remove pusherror() in the current master branch. However, we could not exclude that some one is using this interface. So, I suggest:

1 Considering compatibility, deprecate the reader::pushError(). And I will supplement the corresponding test case by referring to the use case in #57.
2 Remove the OurReader::pusherror(). OurReader is an internal class and user cannot use this interface directly.

dota17:
fix #1054

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -894,9 +894,6 @@ public:
              bool collectComments = true);
   String getFormattedErrorMessages() const;
   std::vector<StructuredError> getStructuredErrors() const;
-  bool pushError(const Value& value, const String& message);
-  bool pushError(const Value& value, const String& message, const Value& extra);
-  bool good() const;
 
 private:
   OurReader(OurReader const&);      // no impl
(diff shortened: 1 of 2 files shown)


include/json/reader.h
```diff
@@ -135,7 +135,9 @@ class JSONCPP_DEPRECATED(
    * \param message The error message.
    * \return \c true if the error was successfully added, \c false if the Value
    * offset exceeds the document size.
+   * \deprecated Not suggest for use.
    */
+  JSONCPP_DEPRECATED("Not suggest for use.")
```
baylesj: Isn't ALL of Reader deprecated? Why does this need to be deprecated too?

If it does, can you say in the deprecation message what you are supposed to do instead?

baylesj:
LGTM.

PR #1056, opened 2019-10-16
COMP: Improve const correctness for ValueIterators

### LINKED ISSUE #914: Issues compiling with uClibc++

hjmjohnson:
The protected deref method had inconsistent interface
of being a const function that returned a non-const
reference.  Resolves #914.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -772,7 +772,14 @@ public:
   char const* memberName(char const** end) const;
 
 protected:
-  Value& deref() const;
+  /*! Internal utility functions to assist with implementing
+   *   other iterator functions. The const and non-const versions
+   *   of the "deref" protected methods expose the protected
+   *   current_ member variable in a way that can often be
+   *   optimized away by the compiler.
+   */
+  const Value& deref() const;
+  Value& deref();
 
   void increment();
 
@@ -895,9 +902,13 @@ public:
     return *this;
   }
 
-  reference operator*() const { return deref(); }
-
-  pointer operator->() const { return &deref(); }
+  /*! The return value of non-const iterators can be
+   *  changed, so the these functions are not const
+   *  because the returned references/pointers can be used
+   *  to change state of the base class.
+   */
+  reference operator*() { return deref(); }
+  pointer operator->() { return &deref(); }
 };
 
 inline void swap(Value& a, Value& b) { a.swap(b); }
--- a/src/lib_json/json_valueiterator.inl
+++ b/src/lib_json/json_valueiterator.inl
@@ -21,7 +21,8 @@ ValueIteratorBase::ValueIteratorBase(
     const Value::ObjectValues::iterator& current)
     : current_(current), isNull_(false) {}
 
-Value& ValueIteratorBase::deref() const { return current_->second; }
+Value& ValueIteratorBase::deref() { return current_->second; }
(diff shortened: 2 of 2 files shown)


include/json/value.h
```diff
@@ -895,9 +902,13 @@ class JSON_API ValueIterator : public ValueIteratorBase {
...
 
-  reference operator*() const { return deref(); }
-
-  pointer operator->() const { return &deref(); }
+  /*! The return value of non-const iterators can be
+   *  changed, so the these functions are not const
+   *  because the returned references/pointers can be used
+   *  to change state of the base class.
+   */
+  reference operator*() { return deref(); }
```
cdunn2001: Technically this broke binary compatibility, so this should have been a major version bump.

I guess we get away with this because probably nobody uses this method. If nobody complains, we can keep it.

BillyDonahue:
I have to argue with the original premise of this change.

I did make the argument in the thread for issue #914 but received no response.

https://github.com/open-source-parsers/jsoncpp/issues/914#issuecomment-493570302

Iterators should model pointer. All standard container itererators do as well.

All 4 of these possiblities exist:

```
    T*
    const T*
    T*const
    const T*const
```

The constness of the pointer is independent of the constness of the thing to which it points.
Same with iterators.  `deref()` should continue to be const method of iterator.

The right signatures are:

```
Value& ValueIterator::deref() const;
const Value& ConstValueIterator::deref() const;
```

Issue #1018, opened 2019-09-04
`add_compile_definitions` command not available in cmake 3.8.2


crazySocket:
**Describe the bug**
You cannot build library with cmake in version 3.8.2.

**To Reproduce**
Make sure your cmake version is 3.8.2. Build the library.

**Expected behavior**
CMake fails to configure build since `add_compile_definitions` is not available in version 3.8.2.

**Desktop (please complete the following information):**
 - OS: windows 10

baylesj:
Good catch. We say we support version 3.8.0, but sounds like we don't.

dota17:
In [CMake 3.12 Release Notes](https://cmake.org/cmake/help/v3.12/release/3.12.html?highlight=add_compile_definitions)， `add_compile_definitions` may be not added until CMake 3.12.

From PR #932 ,  `add_definitions` was changed to `add_compile_definitions`.
But in CMakeList.txt (root directory) , the oldestversion is still 3.8.0
```
set(JSONCPP_OLDEST_VALIDATED_POLICIES_VERSION "3.8.0")
```

Maybe we should change the above message or revert some change of the cmake cleanup.

cdunn2001:
I have found Cmake to be nearly impossible to maintain because different versions are so very different. I only both with Meson, and I generally accept any PR submitted for Cmake.

If someone has a fix for this, great. If the solution is to specify a later version of Cmake, that's fine with me.

hjmjohnson:
As this project is using C++11 features, cmake provides more robust support for C++11 features starting with version 3.12 and greater.

The new Cmake feature of "add_compiler_definitions" is certainly more appropriate, and more robust, but it a new feature that is only available in cmake 3.12 and greater.    Ubuntu 18.04 envirionments come standard with cmake 3.10, so using the bleeding edge features of cmake is probably not a good idea for a project like jsoncpp.  I'll have a look at fixing this later this week.

PR #1057, opened 2019-10-16
BUG: New CMake features used that break backward compatibility

hjmjohnson:
We desire for jsoncpp to compile and be readily available
with older  versions of cmake.  The use of newer cmake
commands requires conditional statements so that older
strategies can be used with older versions of cmake.

Resolves: #1018


hjmjohnson:
@crazySocket This should fix the problem. It provides both backward-compatible (rough solution) and newer more "elegant" solutions when cmake supports it.

I tested using cmake version 3.8.0 to verify that a typical build succeds.

Issue #1038, opened 2019-09-27
The project's code coverage is not very high


dota17:
Recently, I found that the project's code coverage is not very high. So I plan to improve it by adding some test cases. @baylesj 
Here are results from LCOV tool:
![LCOV](https://user-images.githubusercontent.com/50514813/65763058-17d7d000-e155-11e9-9c73-094e1d9df3f9.png)

baylesj:
This is definitely good to know, and that's pretty abysmal for a library like us. I'd love to figure out how to get this as part of TravisCI for each patch, definitely want to get these numbers up.

dota17:
As we know, there are two popular tools in github, which are Codecov and Coveralls. Recently, I have tested these two kind of tools, then I personally prefer to use the coveralls tool, becase it is more easy to configure, and its result is more compact, so I submit a PR about Coveralls. 

Here is the PR #1061 test result,
![image](https://user-images.githubusercontent.com/50514813/67012541-26961f00-f124-11e9-87b0-598bd1550d51.png)

dota17:
Well, the result did not proceed as expected, the CI build does triggle the Coveralls, and the coverage report has been sent to the back-end, but it doesn't leave a comments .
I will keep track of this issue.

dota17:
It seems that we should turn on the `Build pushed branches` button in the Travis CI's settings, because if we want to get the coverage change report, first, we should build pushed branch(usually, it's master branch), and then build pushed pull requests. Finally, the Coveralls can caculate the coverage is increased or decreased by comparing this two build result.

Furthermore, we should specify the build branches in .travis.yml : 
```yml
branches:
  only:
  - master
```
![image](https://user-images.githubusercontent.com/50514813/67183029-052e7f00-f413-11e9-9442-bc9ef9547830.png)

@baylesj  Please take a look. Can we turn on this button? Seems that it used to be turned on, I don't know why it's turned off now.

baylesj:
I think it's fine to try building pushed branches. We can see if it becomes a problem.

dota17:
Have turned it on, and it works! see the PR #1071.

PR #1060, opened 2019-10-17
Issue #1038: Use Coveralls tool in project.

dota17:
#1038


baylesj:
Let's give it a shot and see how it goes. I'll monitor to make sure it doesn't break anything.

PR #1062, opened 2019-10-17
Allow trailing comma in objects and arrays

### LINKED ISSUE #1043: Ignore trailing comma in objects and arrays

kimsey0:
Fixes #1043.

Allows one trailing comma in objects (`{ "count" : 1234, } = { "count" : 1234 }`) and arrays (`[1,] = [1]`), which is ignored.

Introduces new feature `allowTrailingCommas`, which can be used to switch this on and off. It is enabled by default, unless in strict mode.

Currently only allows trailing commas in arrays if `allowDroppedNullPlaceholders` is off, since otherwise, we'd break clients that depend on `[1,]` being parsed as `[1,null]`.

In this draft version, introduces the changes both in the legacy `Reader` and the modern `OurReader`. However, this can be changed if https://github.com/open-source-parsers/jsoncpp/pull/1053#discussion_r335987776 is handled and we allow having _modern only_ tests.


cdunn2001:
I like this feature. I personally use trailing commas in Python all the time. But (A) this PR currently fails for AppVeyor. And (B) it's still a draft. Is that because of the problem with old versus new Reader in the tests?

kimsey0:
(A) just seems to be because I haven't run clang-format. I'll get that done.

(B) is because of the problem with old versus new Reader. I have currently implemented the change in both the old and new Reader, but from https://github.com/open-source-parsers/jsoncpp/issues/1043#issuecomment-542957244, it sounds like I should revert the change to the old reader, since it's deprecated and shouldn't be changed. In that case, in order for the tests I have added to pass, we will need the change in https://github.com/open-source-parsers/jsoncpp/pull/1053/commits/836d954176f88325b82662903e4e365a4ebac03a to be merged, so that the tests are only run for the new Reader.

kimsey0:
@cdunn2001: Would you prefer to have support for trailing slashes in both the old and the new Reader, or just in the new one?

cdunn2001:
Trailing "slashes"? If you mean trailing "commas", just in New is fine. But we need to be sure we have test coverage first.

kimsey0:
Yes, I mean trailing commas. I've edited it in the comment above.

Let's keep this pull request as a draft until https://github.com/open-source-parsers/jsoncpp/pull/1053 is merged. Then I'll revert the changes to the old Reader.

cdunn2001:
We can merge #1053 after clang-format there. Then yes, we can merge this one.

dota17:
Ready for review! I think we can go to the next step now.

cdunn2001:
replaced by #1098

Issue #1046, opened 2019-10-11
Prefer pass-by-value over pass-by-rvalue


baylesj:
I would prefer not specifying rvalue only on the function parameter. It's similarly efficient to have a pass-by-value parameter that you can move in and out of.

If you declare as:
`
bool insert(ArrayIndex index, Value newValue);
`

You can call as
`
Value foo();
insert(index, std::move(foo))`

Some good material on this.
http://blogs.microsoft.co.il/sasha/2014/08/21/c-sink-parameter-passing/

cdunn2001:
I don't have any preference on rvalues. All that code came from @BillyDonahue, who might have an opinion?

dota17:
I tried to modify `Value& Value::operator=()`,  but before that, I'v noticed that there are some history discussions about this, which is #640 and #876, so I am not quit sure about whether or not to modify it with pass-by-value. @baylesj @BillyDonahue @cdunn2001, what's your opinions?

BillyDonahue:
I commented on pull #1067 .

BillyDonahue:
I'd like to get to consensus that this issue can be closed.

The `append` and `insert` of standard containers are well-designed. They provide overloads for `value_type const&` and for `value_type&&`. `Json::Value` is cheap but not free to move, and I don't want to incur even that cost unnecessarily (when the caller gives an rvalue argument). It's currently:

   - a 64-bit word union value_,
   - a word for the bits_ bitfield,
   - a unique_ptr for the comments_, then
   - ptrdiff_t for start_ and limit_.

So nothing expensive right now, but the `Value` type could become move expensive to move in the future, and I wouldn't want to get to that point and have to reintroduce the current conservative `const&/&&` overload set that we could have just left alone.

baylesj:
I don't understand the concern. If Json::Value is not free to move, surely it is not free to copy?

Let's consider three cases of objects:

1 - l-value object (e.g. a non-temporary class instance)
2 - r-value object (e.g. a temporary instance)
3 - an l-value passed as an r-value (e.g. std::move)

With a single pass by value, these objects are:
1 - copied, 2 - moved, 3 - moved

With `value_type const&` and `value_type&&`, these objects are:
1 - copied, 2 - moved, 3 - moved

For things where copying is actually cheaper than moving, e.g. `int`, pass by value is still better:

```cpp
void append(int foo) {
// passes the value of foo (not a reference), which is copied or moved as appropriate.
}
```

Versus:
[...]

PR #1067, opened 2019-10-24
Use universal references in append() method

dota17:
This litte patch fixes issue #1046.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -460,10 +460,8 @@ public:
   /// Return true if index < size().
   bool isValidIndex(ArrayIndex index) const;
   /// \brief Append value to array at the end.
-  ///
   /// Equivalent to jsonvalue[jsonvalue.size()] = value;
-  Value& append(const Value& value);
-  Value& append(Value&& value);
+  Value& append(Value value);
   /// \brief Insert value in array at specific index
   bool insert(ArrayIndex index, Value newValue);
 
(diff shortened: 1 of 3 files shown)


BillyDonahue:
This is not a "universal reference" (which are now called "forwarding references").
It's an rvalue reference.

Anyway, for an "append" member function I'm looking at copying the API of standard container members like vector::push_back, which will always have a const T& and a T&& overload.

The Microsoft.co.il blog referred to in #1046 is having an internal problem right now so I can't read it, but I haven't heard of the {const T&, T&&} overload set as being a problem for append methods before.

BillyDonahue:
@dota17 can we close this PR now?

dota17:
> @dota17 can we close this PR now?

@BillyDonahue 
if we close this PR, then how can we treat this issue #1046?
and how about the `insert` method?

BillyDonahue:
@dota17 you are right. We should get a consensus on #1046 before closing this.

dota17:
Closing.  See the dicussion in #1046.

cdunn2001:
I hope Billy and Jordan don't feel as if they've wasted their own time on this. Personally, this was very informative for me, including the experiments. So I'm glad you tried this, Chen.

dota17:
> I hope Billy and Jordan don't feel as if they've wasted their own time on this. Personally, this was very informative for me, including the experiments. So I'm glad you tried this, Chen.

Yes, I agree, the experiments for this issue are very important. Thanks to Billy and Jordan's experiments. we have a clear and and reliable conclusions for that now.

PR #1072, opened 2019-10-26
Add coverage badge in readme

dota17:


cdunn2001:
This is good. But aren't we still missing coverage of CharReader?

PR #1048, opened 2019-10-11
clang-tidy cleanups 2

neheb:

Issue #747, opened 2018-03-03
PVS Studio static analyser reports some issues


pavel-pimenov:
Report:
http://www.fly-server.ru/pvs-studio/jsoncpp

cdunn2001:
Here are the 2 `High` priority warnings:

json_writer.cpp:624 | High | V730 | Not all members of a class are initialized inside the constructor. Consider inspecting: indented_.
-- | -- | -- | --
json_value.cpp:934 | High | V607 | Ownerless expression '(* this)[newSize - 1]'.
-- | -- | -- | --

I have a fix for the first, but I have no idea for the second. What is the fix for V607, and what exactly is the problem?

pavel-pimenov:
V607 - false positive 

else if (newSize > oldSize)
(*this)[newSize - 1];  <- 

https://www.viva64.com/en/w/v607/

cdunn2001:
Ah. I was looking at the wrong commit locally.
* https://github.com/open-source-parsers/jsoncpp/blob/master/src/lib_json/json_value.cpp#L934

cdunn2001:
If you want to fix the lower priority warnings, go ahead.

PR #1077, opened 2019-11-01
PVS and Clang-tidy modify

dota17:
Fix some remaining lower priority warnings in issue #747:

+ V127. An overflow of the 32-bit variable is possible inside a long cycle which utilizes a memsize-type loop counter.          (int  --> size_t)

Because that PVS Studio report is old, I use clang-tidy tools to scan jsoncpp recently like  #1033 #1048 and find two warnings:
```c++
warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]
warning: use 'using' instead of 'typedef' [modernize-use-using]
```


cdunn2001:
These are all good changes, but I'd really rather see the test-coverage go up before we worry about cosmetic improvements.

src/jsontestrunner/main.cpp
```diff
@@ -56,7 +56,7 @@ static Json::String readInputTestFile(const char* path) {
     return "";
   fseek(file, 0, SEEK_END);
   long const size = ftell(file);
-  size_t const usize = static_cast<unsigned long>(size);
+  auto const usize = static_cast<unsigned long>(size);
```
BillyDonahue: I think clang-tidy messed up here. We had a static_cast to unsigned long captured by a variable declared as const size_t. The conversion to auto is only a no op when unsigned long is the same type as a size_t.
Really the better change here would probably be to `const auto usize = static_cast<size_t>(size);`

src/lib_json/json_value.cpp
```diff
@@ -529,7 +529,7 @@ bool Value::operator<(const Value& other) const {
   }
   case arrayValue:
   case objectValue: {
-    int delta = int(value_.map_->size() - other.value_.map_->size());
+    auto delta = int(value_.map_->size() - other.value_.map_->size());
```
BillyDonahue: This looks like a dangerous downcast of a potentially "negative" `size_t` to `int`, which is implementation-defined behavior. We should make the relatively easy change to make the calculation portable. We should also lose the training wheels on the following line's direct map comparison.

```
    auto thisSize = value_.map_->size();
    auto otherSize = other.value_.map_->size();
    if (thisSize != otherSize)
      return thisSize < otherSize;
    *value_.map_ < *other.value_.map_;
```

include/json/reader.h
```diff
@@ -217,16 +217,16 @@ class JSONCPP_DEPRECATED(
   void skipUntilSpace();
   Value& currentValue();
   Char getNextChar();
-  void getLocationLineAndColumn(Location location, int& line,
-                                int& column) const;
+  void getLocationLineAndColumn(Location location, size_t& line,
```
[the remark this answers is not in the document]
dota17: The warning comes from the PVS-Studio report:
> V127. An overflow of the 32-bit variable is possible inside a long cycle which utilizes a memsize-type loop counter. (int --> size_t)

dota17:
> These are all good changes, but I'd really rather see the test-coverage go up before we worry about cosmetic improvements.

Hi @cdunn2001 
Currently, what's our goal of the test-coverage? Is there a minimal threshold? At the present, we have gone up to 90%+ coverage on the whole.

Since the latest version is only a pre-release version, and some users feedback that there is a little problem with it, so will we need to release a stable version after the coverge reach the goal?

src/jsontestrunner/main.cpp
baylesj: Agreed.

baylesj: Yeah, I definitely don't like the way the code was previously.

src/lib_json/json_reader.cpp
```diff
@@ -761,8 +761,8 @@ Reader::Char Reader::getNextChar() {
   return *current_++;
 }
 
-void Reader::getLocationLineAndColumn(Location location, int& line,
```
baylesj: What's wrong with int here?

https://google.github.io/styleguide/cppguide.html#Integer_Types

>  Because of historical accident, the C++ standard also uses unsigned integers to represent the size of containers - many members of the standards body believe this to be a mistake, but it is effectively impossible to fix at this point. The fact that unsigned arithmetic doesn't model the behavior of a simple integer, but is instead defined by the standard to model modular arithmetic (wrapping around on overflow/underflow), means that a significant class of bugs cannot be diagnosed by the compiler. In other cases, the defined behavior impedes optimization.

BillyDonahue: That's a highly controversial point, and the Google style guide is not much more than an opinion piece on this topic. It's not an "accident" that C++ agrees with C on the type of its `sizeof` operator. It's also no accident that `container.size` in the STL is the same type as an array's `sizeof` operator.

In the surrounding text are listed some pitfalls of mixing arithmetic and a bit of an abdication on whether to give any advice at all w.r.t. choosing integer types, actually.

> Use care when converting integer types. Integer conversions and promotions can cause undefined behavior, leading to security bugs and other problems.

...

> That said, mixing signedness of integer types is responsible for an equally large class of problems.

The reader is left somewhat adrift. A canonical example of `size_t` causing trouble is walking a for loop backwards and underflowing at the last element. I believe it used to be in the style guide. This is not a realistically common bug, IMO. Yes, you have to know how to program in C++ to program in C++. It's not the type's fault.

For line and column numbers, I'm ok with either size_t or int. It doesn't matter much.
I'd prefer not changing it, really, as the lower-risk option.

src/jsontestrunner/main.cpp
```diff
@@ -56,7 +56,7 @@ static Json::String readInputTestFile(const char* path) {
     return "";
   fseek(file, 0, SEEK_END);
   long const size = ftell(file);
-  size_t const usize = static_cast<unsigned long>(size);
+  const auto usize = static_cast<size_t>(size);
```
BillyDonahue: All this buggy fuss.
Meanwhile this whole function is conceptually 4 easy lines. :)

```
static Json::String readInputTestFile(const char* path) {
    std::ifstream in(path, in.binary);
    std::stringstream os;
    os << in.rdbuf();
    return Json::String(os.str());
}
```

src/lib_json/json_value.cpp
```diff
@@ -141,7 +141,7 @@ static inline char* duplicateAndPrefixStringValue(const char* value,
                       "in Json::Value::duplicateAndPrefixStringValue(): "
                       "length too big for prefixing");
   unsigned actualLength = length + static_cast<unsigned>(sizeof(unsigned)) + 1U;
```
BillyDonahue: `size_t actualLength = sizeof(length) + length + 1;`

src/lib_json/json_value.cpp
```diff
@@ -529,9 +529,10 @@ bool Value::operator<(const Value& other) const {
   }
   case arrayValue:
   case objectValue: {
-    int delta = int(value_.map_->size() - other.value_.map_->size());
-    if (delta)
-      return delta < 0;
+    auto thisSize = value_.map_->size();
```
BillyDonahue: We could try doing it an easier way.

```
// order by map size, then by map contents.
auto tieFields = [](const Value& v) { return std::tie(v.map_->size(), *v.map_); };
return tieFields(*this) < tieFields(other);
```

cdunn2001:
> Currently, what's our goal of the test-coverage? Is there a minimal threshold? At the present, we have gone up to 90%+ coverage on the whole.

Well, I thought someone said the jsontestrunner used the old Reader, not the new CharReader. If that's true, we are missing a lot of *behavioral* coverage. But maybe I misunderstood that.

src/jsontestrunner/main.cpp
dota17: Yes, that's great! but this sould be in another PR.

PR #1075, opened 2019-10-28
Implemented as<T>() and is<T>() with accompanying tests

JulianvDoorn:
Solves issue #1066

Issue #1066, opened 2019-10-24
template-based as<type>() accessor


JulianvDoorn:
**Is your feature request related to a problem? Please describe.**
It's slightly disappointed when I have to write a helper function which allows to determine the type of the lookup at compile time.
**Describe the solution you'd like**
I wish to be able to call JsonValue::as<double>() as an alternative of JsonValue::asDouble().

**Describe alternatives you've considered**
I've written a function that uses metaprogramming features to determine the type which is being looked up at compile time. (assumes C++17, but I imagine an implementation is possible, SFINAE?):
```C++
template <class dat_type>
dat_type jsonValueAsType(const Json::Value& jsonVal) {
    if constexpr (std::is_same_v<dat_type, double>) {
        return jsonVal.asDouble();
    } else if constexpr (std::is_same_v<dat_type, float>) {
        return jsonVal.asFloat();
    }
}
```

**Additional context**
No additional context is required, I believe. But with focus of the C++ community to work towards better metaprogramming facilities, I think it is in place to add an JsonValue::as\<T\>() function.

dota17:
Please submit a PR if that's really what you want.

dota17:
Closing as fixed.

PR #1080, opened 2019-11-03
Issue 1066

BillyDonahue:
Julian I took out the macros and tried to compress the unit tests a bit.
It's just a suggestion. You can take it or leave it, really.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -399,6 +399,40 @@ public:
   double asDouble() const;
   bool asBool() const;
 
+  template <class T>
+  struct asLookupHelper;
+
+  #define defAsLookupHelper(type, lookup) \
+  template<> \
+  struct asLookupHelper< type > { \
+    static type as(const Value& val) { \
+      return val. lookup (); \
+    } \
+  }
+
+  defAsLookupHelper(const char*, asCString);
+  defAsLookupHelper(String, asString);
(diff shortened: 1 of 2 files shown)


cdunn2001:
TIL about "Draft" pull-requests. Anyway, this is interesting. (re: #1066)

JulianvDoorn:
I have no idea how these drafts work, am I meant to merge this or is this simply an addition for #1075 ? Also- the travis-ci build fails which prevents merging, I can't seem to reproduce it on my own machine.

dota17:
I think this PR is better.

From the ci report, it fails in the case `Linux xenial gcc cmake coverage`.
The Travis-CI tells that there are some errors when [building and testing with CMake](https://github.com/open-source-parsers/jsoncpp/wiki/Building#building-and-testing-with-cmake).
I try to reproduce it on my own machine. When i build with CMake, i got these following errors:
```
/src/lib_json/../../include/json/value.h:419:13: error: explicit specialization in non-namespace scope 'class Json::Value'
   template <> bool as<bool>() const { return asBool(); }
             ^
....
```
And i got some same build errors when building with meson/ninja.

BillyDonahue:
Oh, I forgot you can't specialize a template inside a class.
You just have to move the specializations outside the class.
I did that in 826228e

Issue #1081, opened 2019-11-03
Upgrading to Python 3


jakirkham:
**Is your feature request related to a problem? Please describe.**

It appears that Python 2.6 is used in some places like [here]( https://github.com/open-source-parsers/jsoncpp/blob/80bc776bae74261742b7c2d0b8dc31ec1718ba4a/makerelease.py#L8 ) and [here]( https://github.com/open-source-parsers/jsoncpp/blob/aebc7faa4fa69fa16ffe9082d03a3527120d5fa3/src/jsontestrunner/CMakeLists.txt#L1 ). However Python 2.6 has long since reach its EOL. Not to mention Python 2.7 will not be supported further in the new year.

**Describe the solution you'd like**

It would be good to move all scripts to support Python 3.

**Describe alternatives you've considered**

A good compromise would be to get all of the scripts to support Python 2/3, which should continue to work after Python 2 reaches its EOL.

**Additional context**

Here's the [Python EOL guide]( https://devguide.python.org/devcycle/#end-of-life-branches ).

cdunn2001:
I've deleted makerelease.py (#1082). Problem solved! :smile:

@hjmjohnson, Could you take care of py2.6 in CMake?

* https://github.com/open-source-parsers/jsoncpp/blob/aebc7faa4fa69fa16ffe9082d03a3527120d5fa3/src/jsontestrunner/CMakeLists.txt#L1

Note: We actually run jsontestrunner via Meson in TravisCI, using python3, so I don't think the python version actually matters.

... Oh, we actually still use CMake -- in the AppVeyor build. I think that runs jsontestrunner too -- yes, `jsontestrunner_exe.vcxproj` -- so we have a test for whatever change you make. I just hate dealing with CMake at all. I wonder if we could use Meson in AppVeyor.

hjmjohnson:
OK. Moving to python3 only for support.

PR #1085, opened 2019-11-04
DRAFT: Format python for python36

hjmjohnson:


cdunn2001:
This is mostly changing single-quote to double-quote. Harmless, but also pointless.

This also *removes* python2 support, which I assume is the reason for the AppVeyor failure.

Instead of changing this, we might be better off with a better test-framework and test-runners. We need to be able to run slightly different tests with different JSON Readers/Writers and configuration. (See other Issue.)

hjmjohnson:
@cdunn2001  I thought that the intent was to remove python 2 support since it is at end of life.  PR #1084 mandates the use of python3 as discussed in #1081.  If we want to continue supporting python 2.7, more work will need to be done to PR#1084.

hjmjohnson:
> This is mostly changing single-quote to double-quote. Harmless, but also pointless.

The point is to provide consistency by following a set of common formatting guides.  Black is one of may tools that can do this automatically.  My experience in other projects is that "giving control to an automated tool" for formatting greatly increased overall productivity, at the expense of a few "pointless" formatting compromises.  That's my 2cents worth of opinion.    This was 2 minutes of work, so I'm perfectly happy with a rejection of this PR.

cdunn2001:
I don't have a problem with dropping python2 support. But AppVeyor does need to pass.

As for running "black", I'm fine with it, but I think maybe we should hold off until we finish some serious refactoring or copy/pasting to ensure that we run separate tests for the old and new Readers. Git-history can be helpful for tracking the provenance of bugs, and a change this large muddies the history.

But after we're stable, I'd be in favor of this PR. (And I'm not *strongly* opposed today.)

hjmjohnson:
@cdunn2001 Thanks.  Let's revisit this after some stability isssues are resolved.  It really is 2-3 minutes of work to remake this PR.

baylesj:
I'm going to close this out as stale. Feel free to revisit this and reopen as appropriate.

PR #1087, opened 2019-11-07
clang-tidy fixes again

neheb:
I ran it through CMake this time. Coverage is greater.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,5 @@
 ---
-Checks:          'google-readability-casting,modernize-use-default-member-init,modernize-use-using,modernize-use-auto,readability-redundant-member-init'
+Checks:          'google-readability-casting,modernize-deprecated-headers,modernize-loop-convert,modernize-use-auto,modernize-use-default-member-init,modernize-use-using,readability-else-after-return,readability-redundant-member-init,readability-redundant-string-cstr'
 WarningsAsErrors: ''
 HeaderFilterRegex: ''
 AnalyzeTemporaryDtors: false
--- a/src/test_lib_json/fuzz.cpp
+++ b/src/test_lib_json/fuzz.cpp
@@ -9,7 +9,6 @@
 #include <json/config.h>
 #include <json/json.h>
 #include <memory>
-#include <stdint.h>
 #include <string>
 
 namespace Json {
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -1476,9 +1476,7 @@ void ValueTest::checkMemberCount(Json::Value& value,
   JSONTEST_ASSERT_PRED(checkConstMemberCount(value, expectedCount));
 }
 
-ValueTest::IsCheck::IsCheck()
-
-    = default;
+ValueTest::IsCheck::IsCheck() = default;
 
 void ValueTest::checkIs(const Json::Value& value, const IsCheck& check) {
   JSONTEST_ASSERT_EQUAL(check.isObject_, value.isObject());
@@ -3752,8 +3750,8 @@ JSONTEST_FIXTURE_LOCAL(FuzzTest, fuzzDoesntCrash) {
 int main(int argc, const char* argv[]) {
   JsonTest::Runner runner;
 
-  for (auto it = local_.begin(); it != local_.end(); it++) {
-    runner.add(*it);
+  for (auto& local : local_) {
+    runner.add(local);
   }
 
   return runner.runCommandLine(argc, argv);
--- a/src/jsontestrunner/main.cpp
+++ b/src/jsontestrunner/main.cpp
@@ -57,10 +57,10 @@ static Json::String readInputTestFile(const char* path) {
   if (!file)
     return "";
   fseek(file, 0, SEEK_END);
-  long const size = ftell(file);
-  const auto usize = static_cast<size_t>(size);
+  auto const size = ftell(file);
+  auto const usize = static_cast<size_t>(size);
   fseek(file, 0, SEEK_SET);
-  char* buffer = new char[size + 1];
+  auto buffer = new char[size + 1];
   buffer[size] = 0;
   Json::String text;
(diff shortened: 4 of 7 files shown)


dota17:
I think it's not a good time to `use trailing return type`. 
Other changes are good. And whether those rules were added to `.clang-tidy`?

neheb:
I agree. Will do so.

neheb:
There's an error. Not sure how to fix.

neheb:
Fixed finally.

BTW, I noticed a bug in CMake where running make clean removes important source files.

Is CMake to be deprecated or something? I've noticed the meson build files, which is interesting since CMake can generate Meson build files.

hjmjohnson:
> BTW, I noticed a bug in CMake where running make clean removes important source files.

Can you provide a description of how to replicate this problem?  Which important source files.   CMake should not have this behavior, and I have never encountered such a problem.  I'll look into fixing it if I can replicate and confirm that this is a problem.

neheb:
> > BTW, I noticed a bug in CMake where running make clean removes important source files.
> 
> Can you provide a description of how to replicate this problem? Which important source files. CMake should not have this behavior, and I have never encountered such a problem. I'll look into fixing it if I can replicate and confirm that this is a problem.

cmake . && make clean results in

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    example/readFromStream/errorFormat.json
	deleted:    example/readFromStream/readFromStream.cpp
	deleted:    example/readFromStream/withComment.json
	deleted:    example/readFromString/readFromString.cpp
	deleted:    example/streamWrite/streamWrite.cpp
	deleted:    example/stringWrite/stringWrite.cpp

src/test_lib_json/main.cpp
```diff
@@ -1464,8 +1464,7 @@ JSONTEST_FIXTURE_LOCAL(ValueTest, nonIntegers) {
 void ValueTest::checkConstMemberCount(const Json::Value& value,
                                       unsigned int expectedCount) {
   unsigned int count = 0;
-  Json::Value::const_iterator itEnd = value.end();
-  for (Json::Value::const_iterator it = value.begin(); it != itEnd; ++it) {
+  for ([[maybe_unused]] auto it : value) {
```
neheb: Anyone got a solution for this?

[[maybe_unused]] is ignored in VS2015. VS2017 demands a c++17 flag (as this is a c++17 feature).

GCC seems to work.

__attribute__((unused)) works in GCC but not MSVC.

hjmjohnson:
> cmake . && make clean

AHH!  You are doing an in-source build, which is generally considered bad form, precisely because of this situation.  We need to have some logic added to prevent insource builds.

from the jsoncpp directory, make a separate directory targeted for you build ( `../jsoncpp-release-build`) and then run `cd ../jsoncpp-release-build; cmake ../jsoncpp`.

hjmjohnson:
@neheb See #1091 for code that stops insource builds early so that build artifacts do not contaminate the source tree.

src/test_lib_json/fuzz.cpp
```diff
@@ -44,7 +43,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size) {
   std::unique_ptr<Json::CharReader> reader(builder.newCharReader());
 
   Json::Value root;
-  const char* data_str = reinterpret_cast<const char*>(data);
+  const auto data_str = reinterpret_cast<const char*>(data);
```
BillyDonahue: `data_str` changed type here in a subtle way, from `const char*` to `const char* const`.

src/lib_json/json_writer.cpp
```diff
@@ -159,7 +159,8 @@ String valueToString(double value, bool useSpecialFloats,
 
   // try to ensure we preserve the fact that this was given to us as a double on
   // input
-  if (buffer.find('.') == buffer.npos && buffer.find('e') == buffer.npos) {
+  if (buffer.find('.') == Json::String::npos &&
```
BillyDonahue: Downgrade. `npos` is a property of the object you just did a `find` on. If the string type of that object changes, then the particular `npos` you compare the result with should follow.

.clang-tidy
```diff
@@ -1,11 +1,19 @@
 ---
-Checks:          'google-readability-casting,modernize-use-default-member-init,modernize-use-using,readability-redundant-member-init'
+Checks:          'google-readability-casting,modernize-deprecated-headers,modernize-loop-convert,modernize-use-auto,modernize-use-default-member-init,modernize-use-using,readability-else-after-return,readability-redundant-member-init,readability-static-accessed-through-instance,readability-redundant-string-cstr,readability-uppercase-literal-suffix'
```
BillyDonahue: I think this is adding too much. Some of these are not helpful, IMHO.
The modernize transformations should not be applied to unit tests without careful hand-review.

`readability-static-accessed-through-instance` is particularly incorrect, I have to say.

src/test_lib_json/main.cpp
```diff
@@ -1526,9 +1522,9 @@ JSONTEST_FIXTURE_LOCAL(ValueTest, compareInt) {
 }
 
 JSONTEST_FIXTURE_LOCAL(ValueTest, compareUInt) {
-  JSONTEST_ASSERT_PRED(checkIsLess(0u, 10u));
-  JSONTEST_ASSERT_PRED(checkIsLess(0u, Json::Value::maxUInt));
-  JSONTEST_ASSERT_PRED(checkIsEqual(10u, 10u));
+  JSONTEST_ASSERT_PRED(checkIsLess(0U, 10U));
```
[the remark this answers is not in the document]
neheb: It comes from here: https://wiki.sei.cmu.edu/confluence/pages/viewpage.action?pageId=87152241

neheb:
Rebased and removed the uppercase literal suffix commit since it seems unwanted.

BillyDonahue:
LGTM

PR #1091, opened 2019-11-10
ENH: Prevent cmake in source builds

hjmjohnson:
Building directly inside the root of the source tree
can cause problems where the build intermediate files
overwrite or conflict with the intended source code
files.

This modification identifies this problem and
issues failure messages and suggestions to over
come the problem with more robust build suggestion.


hjmjohnson:
See #1087 for motivation for this improvement.

baylesj:
It looks like this change requires built bot reconfiguration.  If you can change the AppVeyor config to not build in the source directory, I'm happy to approve this patch.

dota17:
@hjmjohnson I think this issue can move forward now. Would you like to take a little time to look at this PR to move it forward?

cdunn2001:
@hjmjohnson, we've merged the clang-tidy fixes. Maybe that will help you fix VisualStudio? Otherwise, let's close this.

baylesj:
@cdunn2001, I think? I've fixed the appveyor issue, it's a valid one. I know our CMake support is deprecated but I'm still interested in landing this patch.

baylesj:
I think this is a good step forward.

baylesj:
Travis is being slow, but it's all green so going to submit.

gizmocuz:
Sorry for asking, but we use your library as a sub project and build this via:

  add_subdirectory (extern/jsoncpp EXCLUDE_FROM_ALL)
  target_link_libraries(domoticz jsoncpp_static)

With this change it's not possible to build this anymore

Could you make this an Option which is default set to YES, but when used as a external library in other projects we can set this to NO ?

brucemiranda:
This has caused me a problem Building jsoncpp today. In the end I had to comment the two new lines added in the CMakelists within the json folder.

gizmocuz:
I forked this repo, and now use my copy with also these two lines removed
This is not standard CMake, and it should not cause problems when building in the main folder and overwrite intermediate and source code files.
If it does, something else is wrong.
Sure it is best practice to create a build folder, but I would not force this...

@baylesj , @hjmjohnson , hope you can reconsider this patch

cdunn2001:
@baylesj , it was a good try, but I guess we need to revert at least some part of this. What do you think? The objections from these users seem valid.

PR #1092, opened 2019-11-11
Change all typedefs into using decls

BillyDonahue:
Remove the unused `size_t` typedefs from the Iterator classes.
size_t is a standard C typedef so we don't ever want to hide that.


dota17:
This looks like same as the #1077

BillyDonahue:
Yeah I guess it's the same as #1077, which I hadn't seen. I'll withdraw this one then.

PR #1093, opened 2019-11-11
some test coverage for Value::iterator

BillyDonahue:


BillyDonahue:
Brings json_valueiterator.inl to 100% coverage.

src/test_lib_json/main.cpp
```diff
@@ -3394,18 +3395,85 @@ JSONTEST_FIXTURE_LOCAL(BuilderTest, settings) {
...
+    auto it = json.begin();
+    for (;; ++it, ++i) {
+      auto dist = it - json.begin();
+      JSONTEST_ASSERT_EQUAL(i, dist);
+      if (it == json.end())
+        break;
+    }
+  }
+  {
+    Json::Value j; // empty
```
dota17: Can we directly use `Json::Value empty;`?

src/test_lib_json/main.cpp
```diff
@@ -3394,18 +3395,85 @@ JSONTEST_FIXTURE_LOCAL(BuilderTest, settings) {
...
+}
+
+JSONTEST_FIXTURE_LOCAL(IteratorTest, nullValues) {
+  {
+    Json::Value json;
+    auto end = json.end();
+    auto endCopy = end;
+    JSONTEST_ASSERT(endCopy == end);
+    endCopy = end;
+    JSONTEST_ASSERT(endCopy == end);
```
dota17: I don't understand why repeatly assert this.

BillyDonahue: I'm testing how the copy assignment operator= functions with `null` iterators. So the first assert is checking that copy construction yields equal iterators. The second is checking that copy assignment yields equal iterators as well.

src/test_lib_json/main.cpp
BillyDonahue: Done.

PR #1095, opened 2019-11-12
update charReader test case

dota17:
update charReader test case
Coverage of json_reader.cpp raises up to 89.52%

### DIFF
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -22,11 +22,7 @@
 #include <sstream>
 #include <string>
 
-#if __cplusplus >= 201103L || (defined(_CPPLIB_VER) && _CPPLIB_VER >= 520)
 using CharReaderPtr = std::unique_ptr<Json::CharReader>;
-#else
-typedef std::auto_ptr<Json::CharReader> CharReaderPtr;
-#endif
 
 // Make numeric limits more convenient to talk about.
 // Assumes int type in 32 bits.
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -2979,14 +2979,6 @@ JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseObjectWithErrors) {
                             "  Missing '}' or object member name\n");
     JSONTEST_ASSERT_EQUAL("value", root["property"]);
   }
-  {
-    char const doc[] = "{ \"property\" : \"value\" , }";
-    bool ok = reader->parse(doc, doc + std::strlen(doc), &root, &errs);
-    JSONTEST_ASSERT(!ok);
-    JSONTEST_ASSERT(errs == "* Line 1, Column 26\n"
-                            "  Missing '}' or object member name\n");
-    JSONTEST_ASSERT_EQUAL("value", root["property"]);
-  }
 }
 
 JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseArrayWithErrors) {
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -22,6 +22,12 @@
 #include <sstream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


src/test_lib_json/main.cpp
```diff
@@ -2834,14 +2834,195 @@ JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseWithNoErrorsTestingOffsets) {
...
+    JSONTEST_ASSERT(ok);
+    JSONTEST_ASSERT(errs.empty());
+    JSONTEST_ASSERT_EQUAL("", root[0]);
+  }
+  {
+    char const doc[] = "[\"\\u8A2a\"]";
+    bool ok = reader->parse(doc, doc + std::strlen(doc), &root, &errs);
+    JSONTEST_ASSERT(ok);
+    JSONTEST_ASSERT(errs.empty());
+    JSONTEST_ASSERT_EQUAL("訪", root[0]);
```
BillyDonahue: I'm not a fan of leaving  raw unicode in C++ source. I can't type it or modify it without a keyboard and knowledge I don't have. Can we use a C++ unicode escape for this instead and leave the raw glyph maybe as a comment?

src/test_lib_json/main.cpp
```diff
@@ -2834,14 +2834,195 @@ JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseWithNoErrorsTestingOffsets) {
...
+    JSONTEST_ASSERT_EQUAL(2u, root.size());
+    JSONTEST_ASSERT_STRING_EQUAL("x\ty", root["a"].asString());
+    JSONTEST_ASSERT_STRING_EQUAL("x\\y", root["b"].asString());
+  }
+  delete reader;
+}
+
+JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseComment) {
+  Json::CharReaderBuilder b;
+  Json::CharReader* reader(b.newCharReader());
```
BillyDonahue: use unique_ptr for the CharReader* in your tests in case the test asserts cause the function to exit via throwing.

src/test_lib_json/main.cpp
dota17: Yes, I aggree with you.

src/test_lib_json/main.cpp
dota17: Yes, I will changed it all.

src/test_lib_json/main.cpp
```diff
@@ -22,6 +22,12 @@
 #include <sstream>
 #include <string>
 
+#if __cplusplus >= 201103L || (defined(_CPPLIB_VER) && _CPPLIB_VER >= 520)
```
BillyDonahue: Remove #if and assume a full C++11 implementation.

src/test_lib_json/main.cpp
```diff
@@ -2998,56 +3162,85 @@ JSONTEST_FIXTURE_LOCAL(CharReaderFailIfExtraTest, issue107) {
...
   JSONTEST_ASSERT_EQUAL("value", root[1u]);
-  delete reader;
 }
 JSONTEST_FIXTURE_LOCAL(CharReaderFailIfExtraTest, commentAfterBool) {
   Json::CharReaderBuilder b;
   Json::Value root;
   char const doc[] = " true /*trailing\ncomment*/";
   b.settings_["failIfExtra"] = true;
-  Json::CharReader* reader(b.newCharReader());
+  CharReaderPtr reader(b.newCharReader());
```
BillyDonahue: (no action required now)

It would be great if the setup of the reader object were delegated out to a helper function, perhaps a member of the Fixture (that's supposed to be what they're for), so we wouldn't have to make so many changes to add or maintain tests. Most of these tests are doing the same thing with different strings and asserting slightly different expectations.

src/test_lib_json/main.cpp
cdunn2001: But I think we might need to alter the AppVeyor tests in that case. Try it and see. If we have to drop support for older VisualStudio, that's fine, but we'll have to state that clearly in the release notes.

PR #1098, opened 2019-11-14
Allow trailing comma in objects and arrays

cdunn2001:
replaces #1062 

rebased and bumped the fuzz number:
`src/test_lib_json/fuzz.cpp`:
```cpp
builder.settings_["allowTrailingCommas_"] = hash_settings & (1 << 10);
```


dota17:
As we discussioned in old PR: 
https://github.com/open-source-parsers/jsoncpp/pull/1062/files#issuecomment-549434023 
should we apply this new feature in the old `Reader`?

baylesj:
I don't think we should apply any new features in the old `Reader`, unless it's a massive security vulnerability or something.

dota17:
> I don't think we should apply any new features in the old `Reader`, unless it's a massive security vulnerability or something.

Agreed.
And I should point out directly that these changes involve the old Reader class, we should exclude it.

baylesj:
Thanks for doing #1126.

PR #1099, opened 2019-11-14
Add option JSONCPP_WITH_EXAMPLE

theirix:
Allows to conditionally build examples as it has been done for tests.
Useful for packaging.


CMakeLists.txt
```diff
@@ -82,6 +82,7 @@ option(JSONCPP_WITH_WARNING_AS_ERROR "Force compilation to fail if a warning occ
 option(JSONCPP_WITH_STRICT_ISO "Issue all the warnings demanded by strict ISO C and ISO C++" ON)
 option(JSONCPP_WITH_PKGCONFIG_SUPPORT "Generate and install .pc files" ON)
 option(JSONCPP_WITH_CMAKE_PACKAGE "Generate and install cmake package files" ON)
+option(JSONCPP_WITH_EXAMPLE "Compile JsonCpp example" ON)
```
baylesj: This means default on, right? Do we actually want to always build the example? I would vote for "opt in" not "opt out."

theirix: Me too, we can check to OFF by default. I have based a PR on existing behaviour (to build an example).

Issue #1100, opened 2019-11-14
Drop CPPTL support


baylesj:
For context, see: https://github.com/open-source-parsers/jsoncpp/issues/1097

Basically, we don't need to support CPPTL anymore.

PR #1101, opened 2019-11-14
Issue 1100: Drop CPPTL support

baylesj:
CPPTL support is no longer relevant to JsonCpp, and can be removed from
the library. This patch removes all mentions of CPPTL, by removing all
definitions and code sections conditionally compiled only when JsonCpp
is used with CPPTL. Include guards are also renamed to not refer to
CPPTL where appropriate.

See issue #1100


include/json/json.h
```diff
@@ -6,7 +6,7 @@
 #ifndef JSON_JSON_H_INCLUDED
 #define JSON_JSON_H_INCLUDED
 
-#include "autolink.h"
```
cdunn2001: I don't even know what `autolink.h` was!

Issue #1102, opened 2019-11-14
JsonChecker tests broken


baylesj:
After the  recent trailing commas change, json checker tests aren't broken.

A fix means making these tests run as part of checkin again, and resolving the actual test failures as well.

LocutusOfBorg:
Hello, I have some test failures on Debian experimental, version 1.9.2
first error is:
```
Testing EscapeSequenceTest/writeEscapeSequence: FAILED
* Detail of EscapeSequenceTest/writeEscapeSequence test failure:
/<<PKGBUILDDIR>>/src/test_lib_json/main.cpp(3370): expected == result
  Expected: '["\"","\\","\b","\f","\n","\r","\t","\u0278","\ud852\udf62"]
  '
  Actual  : '["\"","\\","\b","\f","\n","\r","\t","ɸ","𤭢"]
  '
101/102 tests passed (1 failure(s))
```
And happens on arm64, armhf, armel. ppc64el, s390x (and others)
https://buildd.debian.org/status/fetch.php?pkg=libjsoncpp&arch=arm64&ver=1.9.2-2&stamp=1580476809&raw=0
https://buildd.debian.org/status/fetch.php?pkg=libjsoncpp&arch=armhf&ver=1.9.2-2&stamp=1580476979&raw=0
https://buildd.debian.org/status/fetch.php?pkg=libjsoncpp&arch=armel&ver=1.9.2-2&stamp=1580477812&raw=0
https://buildd.debian.org/status/fetch.php?pkg=libjsoncpp&arch=ppc64el&ver=1.9.2-2&stamp=1580476937&raw=0
https://buildd.debian.org/status/fetch.php?pkg=libjsoncpp&arch=s390x&ver=1.9.2-2&stamp=1580476882&raw=0

LocutusOfBorg:
The second kind of error is on mips64el:
https://buildd.debian.org/status/fetch.php?pkg=libjsoncpp&arch=mips64el&ver=1.9.2-2&stamp=1580514060&raw=0
```

2/3 Test #1: jsoncpp_readerwriter ................***Failed    2.54 sec
Unhandled exception:
Exceeded stackLimit in readValue().
Failed to parse input file: 
* Line 1, Column 5
  Missing ',' or ']' in array declaration

TESTING: /<<PKGBUILDDIR>>/test/data/legacy_test_array_03.json FAILED


Failure details:
[...]

PR #1103, opened 2019-11-14
Issue 1102: Fixup test suite, fix broken tests

baylesj:
A recent PR broken the JsonChecker tests by adding support for trailing
commas. This didn't end up breaking the build, because those tests
aren't run, except locally and only using CMake.

This patch fixes the tests by adding exclusions for trailing comma
tests, as well as updates Meson to run these tests as part of `ninja
test`.

See issue #1102.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,8 +27,13 @@ Then,
     #LIB_TYPE=static
     meson --buildtype ${BUILD_TYPE} --default-library ${LIB_TYPE} . build-${LIB_TYPE}
     ninja -v -C build-${LIB_TYPE}
-    cd build-${LIB_TYPE}
-    meson test --no-rebuild --print-errorlogs
+
+    ninja -C build-static/ test
+
+    # Or
+    #cd build-${LIB_TYPE}
+    #meson test --no-rebuild --print-errorlogs
+
     sudo ninja install
 
 ## Building and testing with other build systems
--- a/meson.build
+++ b/meson.build
@@ -103,7 +103,7 @@ test(
     '-B',
     join_paths(meson.current_source_dir(), 'test/runjsontests.py'),
     jsontestrunner,
-    join_paths(meson.current_source_dir(), 'test/data')]
+    join_paths(meson.current_source_dir(), 'test/data')],
   )
 test(
   'jsonchecker_jsontestrunner',
@@ -113,5 +113,6 @@ test(
     join_paths(meson.current_source_dir(), 'test/runjsontests.py'),
     '--with-json-checker',
     jsontestrunner,
-    join_paths(meson.current_source_dir(), 'test/data')]
+    join_paths(meson.current_source_dir(), 'test/data')],
+    workdir : join_paths(meson.current_source_dir(), 'test/data'),
   )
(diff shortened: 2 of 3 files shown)


cdunn2001:
Hmmm. Using this branch locally, I have this:
```
[7/8] Running all tests.
1/3 unittest_jsoncpp_test                   OK       0.34 s
2/3 unittest_jsontestrunner                 FAIL     2.24 s (exit status 1)
3/3 jsonchecker_jsontestrunner              FAIL     2.60 s (exit status 1)
```
... Oh, if I delete the build-dir and start over, then it works. That's odd. Meson/Ninja usually do a good job of auto-updating for changes. Very suspicious, but it does work.
```
1/3 unittest_jsoncpp_test                   OK       0.15 s
2/3 unittest_jsontestrunner                 OK       3.17 s
3/3 jsonchecker_jsontestrunner              OK       4.04 s
```
Thanks for fixing this!

... Oh, hold on. It fails if you run a second time:
```
$ ninja -C build-static/ test
$ ninja -C build-static/ test
1/3 unittest_jsoncpp_test                   OK       0.03 s
2/3 unittest_jsontestrunner                 FAIL     1.17 s (exit status 1)
3/3 jsonchecker_jsontestrunner              OK       3.74 s
```
So we still have something to fix ...

cdunn2001:
I've verified that `ninja test` is repeatable on`master`, so something fishy is happening on this branch.

baylesj:
I'm digging back into this... looks like a fun problem. Basically both the `unittest_jsontestrunner` and `jsonchecker_jsontestrunner` are both intermittently flaky. Maybe a timing issue? I'll update once I have it sorted.

baylesj:
I figured it out! It's a race condition--for some reason the `runjsontests.py` script is trying to read the test files before they are done. I'm working on improving the hygeine of the test harness, and hopefully will have a 100% reliable configuration soon.

baylesj:
I think this is ready for another look!

dota17:
> I think this is ready for another look!

Is there any progress？

cdunn2001:
This worked:
```
cd build-static; meson test --no-rebuild --print-errorlogs
```
But this failed (from top-dir):
```
ninja -C build-static/ test

ninja: Entering directory `build-static/'
[0/1] Running all tests.
1/3 unittest_jsoncpp_test                   OK       0.06 s
2/3 unittest_jsontestrunner                 FAIL     2.05 s (exit status 1)
3/3 jsonchecker_jsontestrunner              OK       2.67 s
...
FAILED: meson-test
/usr/local/Cellar/meson/0.52.0/libexec/bin/meson test --no-rebuild --print-errorlogs
```
I'll update **meson**, but I don't think that's the issue...

cdunn2001:
After upgrading python/meson/ninja on my Mac, I don't see a failure anymore. Weird. But the tests seem slow... Ok. I see the problem. Meson runs 3 tests: C++ unit-tests, a python round-trip checker on `test/data/*.json`, and the same python round-trip checker on those same `*.json` in addition to any expected-failure tests plus (some of) the tests under `jsonchecker`. The problem is that the non-fail tests under `test/data/*.json` are run twice. That's easy to prevent.

This whole thing should be done differently, but I try to solve just one specific problem at a time. For now I will revert your last 2 commits and skip the repetition of those tests. Tomorrow....

PR #1105, opened 2019-11-15
Modify parseFromStream method

dota17:


cdunn2001:
I don't see one way as clearly better than the other. Which is more efficient? In runtime or memory?

BillyDonahue:
Getline is worse. It reads the file scanning for a character and rdbuf just copies blocks at a time.
of course this is irrelevant if the getline idea doesn't work. And I don't think it works.

What's happening is that the EOF happens to be an int -1, casting it to a char yields -1 or 255 depending on whether char is signed, and we just luck out in that that bit pattern doesn't appear in the files we read with this function, so we get the whole file. But it happens to work I guess. But it's far from semantically correct. We should just close this.

PR #1106, opened 2019-11-15
small fix Value::isConvertibleTo

pavel-pimenov:


cdunn2001:
What's being fixed? It looks like it now just falls through from the realvalue case to the booleanvalue case. It seems to remove a redundant line, but that line added clarity. I must be reading this wrong. Is there any actual change? Please explain.

Also, if there is an error, there should be a failing test that this fixes.

pavel-pimenov:
V6067. Two or more case-branches perform the same actions
https://www.viva64.com/en/w/v6067/

изображение

PR #1108, opened 2019-11-28
Add test cases for Reader 

dota17:
Add test cases for Reader in src\lib_json\json_reader.cpp.

### DIFF
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -2653,16 +2653,193 @@ JSONTEST_FIXTURE_LOCAL(ReaderTest, parseWithNoErrors) {
   JSONTEST_ASSERT(reader.getStructuredErrors().empty());
 }
 
+JSONTEST_FIXTURE_LOCAL(ReaderTest, parseObject) {
+  Json::Reader reader;
+  Json::Value root;
+  {
+    bool ok = reader.parse("{\"property\"}", root);
+    JSONTEST_ASSERT(!ok);
+    JSONTEST_ASSERT(reader.getFormattedErrorMessages() ==
+                    "* Line 1, Column 12\n  Missing ':' after object "
+                    "member name\n");
+    std::vector<Json::Reader::StructuredError> errors =
+        reader.getStructuredErrors();
+    JSONTEST_ASSERT(errors.size() == 1);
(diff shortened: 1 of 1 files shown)


src/test_lib_json/main.cpp
```diff
@@ -2751,13 +2892,6 @@ JSONTEST_FIXTURE_LOCAL(ReaderTest, parseChineseWithOneError) {
   JSONTEST_ASSERT(reader.getFormattedErrorMessages() ==
                   "* Line 1, Column 19\n  Syntax error: value, object or array "
                   "expected.\n");
-  std::vector<Json::Reader::StructuredError> errors =
```
BillyDonahue: why are these structured error verifications being deleted?

src/test_lib_json/main.cpp
```diff
@@ -2653,16 +2653,126 @@ JSONTEST_FIXTURE_LOCAL(ReaderTest, parseWithNoErrors) {
...
+    bool ok = reader.parse("[ \"\\uD801\" ]", root);
+    JSONTEST_ASSERT(!ok);
+    JSONTEST_ASSERT(reader.getFormattedErrorMessages() ==
+                    "* Line 1, Column 3\n"
+                    "  additional six characters expected to "
+                    "parse unicode surrogate pair.\n"
+                    "See Line 1, Column 10 for detail.\n");
+  }
+  {
+    bool ok = reader.parse("[ \"\\uD801\\d1234\" ]", root);
```
BillyDonahue: lowercase hex for the \u digits.

src/test_lib_json/main.cpp
dota17: I think getStructuredErrors() is not necessary for every testcase. Removing structured error verifications makes the testcase more clear and is convenient to add new testcase. In line 2819, we keep one structured error verifications for offset test.

BillyDonahue: let's keep structured error checking and remove formatted error checking, which has less information anyway.
The structured errors can be checked by a helper function so we can make the tests compact.

BillyDonahue:
Suggested changes submitted as PR to your fork. https://github.com/dota17/jsoncpp/pull/75

PR #1124, opened 2019-12-21
Meson updates

SoapGentoo:
Hi Chris,
we'd like to disable tests for most users in Gentoo, hence I've made them optional (but still default enabled). I've modernised some parts of the `meson.build` too.

### DIFF
--- a/include/json/version.h
+++ b/include/json/version.h
@@ -9,10 +9,10 @@
 // 3. /CMakeLists.txt
 // IMPORTANT: also update the SOVERSION!!
 
-#define JSONCPP_VERSION_STRING "1.9.2"
+#define JSONCPP_VERSION_STRING "1.9.3"
 #define JSONCPP_VERSION_MAJOR 1
 #define JSONCPP_VERSION_MINOR 9
-#define JSONCPP_VERSION_PATCH 2
+#define JSONCPP_VERSION_PATCH 3
 #define JSONCPP_VERSION_QUALIFIER
 #define JSONCPP_VERSION_HEXA                                                   \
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -44,13 +45,12 @@ else
 endif
 
 jsoncpp_lib = library(
-  'jsoncpp',
-  [ jsoncpp_headers,
-    'src/lib_json/json_tool.h',
```
dota17: i am new at meson.
just make sure if this file `json_tool.h` will be removed or a mistake, because there are some methods used in `json_writer.cpp` and `json_reader.cpp`

SoapGentoo: a (compiled) target doesn't need to include `.h` files. `.h` files aren't compiled and hence don't have to be part of the target.

SoapGentoo:
@dota17 thanks for merging, could you tag the latest commit as `1.9.3` so we can rely on it in Gentoo?

cdunn2001:
Hi, @SoapGentoo. I haven't been paying any attention to this project for some time. I think we're about to have a new release though, FWIW. Thanks for the Meson fix. It seems like half the Issues here are with Cmake.

include/json/version.h
```diff
@@ -9,10 +9,10 @@
 // 3. /CMakeLists.txt
 // IMPORTANT: also update the SOVERSION!!
 
-#define JSONCPP_VERSION_STRING "1.9.2"
+#define JSONCPP_VERSION_STRING "1.9.3"
```
jay0511, line 12: 1

JSONCPP_VERSION_STRING

jay0511:
JSONCPP_VERSION_STRING

Issue #737, opened 2018-01-29
jsoncpp v1.8.4 tries to ccache when not installed


rkeene:
jsoncpp v1.8.4 tries to use "ccache" to compile, even though such a tool does in the $PATH.  FIND_PROGRAM in cmake apparently tries to search other paths, even though doing so will result in a broken result.

```
+ cmake -DCMAKE_INSTALL_PREFIX= -DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=OFF -DJSONCPP_LIB_BUILD_STATIC=OFF -DJSONCPP_LIB_BUILD_SHARED=ON -DJSONCPP_WITH_TESTS=OFF -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSROOT=/home/rkeene/devel/aurae/node/root/packages/jsoncpp/workdir-rRBl3QaJjuAX/.fakeroot -DCMAKE_C_COMPILER=x86_64-knightpoint-linux-gcc -DCMAKE_CXX_COMPILER=x86_64-knightpoint-linux-g++ -DCMAKE_FIND_ROOT_PATH_MODE_PROGRAM=NEVER -DCMAKE_FIND_ROOT_PATH_MODE_LIBRARY=ONLY -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=ONLY -DCMAKE_FIND_ROOT_PATH_MODE_PACKAGE=ONLY -DCMAKE_CROSSCOMPILING=1 .
-- The C compiler identification is GNU 7.3.0
-- The CXX compiler identification is GNU 7.3.0
-- Check for working C compiler: /home/rkeene/devel/aurae/common/compiler/online/x86_64-knightpoint-linux/bin/x86_64-knightpoint-linux-gcc
-- Check for working C compiler: /home/rkeene/devel/aurae/common/compiler/online/x86_64-knightpoint-linux/bin/x86_64-knightpoint-linux-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /home/rkeene/devel/aurae/common/compiler/online/x86_64-knightpoint-linux/bin/x86_64-knightpoint-linux-g++
-- Check for working CXX compiler: /home/rkeene/devel/aurae/common/compiler/online/x86_64-knightpoint-linux/bin/x86_64-knightpoint-linux-g++ -- works
[...]

PR #1131, opened 2020-01-09
removing ccache from cmake

conjam:
seeing issues with ccache when using CMake -- removed it as per https://github.com/open-source-parsers/jsoncpp/issues/737.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,12 +186,6 @@ elseif(CMAKE_CXX_COMPILER_ID STREQUAL "Intel")
     endif()
 endif()
 
-find_program(CCACHE_FOUND ccache)
-if(CCACHE_FOUND)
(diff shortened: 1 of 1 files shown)


dota17:
@conjam 
Thanks for your PR.
But from https://github.com/open-source-parsers/jsoncpp/issues/737#issuecomment-361382227, is it better to change to not search outside $PATH?

conjam:
> @conjam
> Thanks for your PR.
> But from [#737 (comment)](https://github.com/open-source-parsers/jsoncpp/issues/737#issuecomment-361382227), is it better to change to not search outside $PATH?

from my understanding, find_program() in cmake has ... issues. I played around with setting PATH, but it still breaks my build, even with the work around mentioned in the thread (I was trying to use bazel wrapping cmake). In the end I just ported this to bazel -- could submit PR if you like, figure that would just cause you headaches down the road. 

Regardless, some folks seem to be against this so I'll close it.

PR #1133, opened 2020-01-24
Remove '=delete' from template methods for Xcode 8

dgobbi:
For Apple clang-800.0.42.1, which was released with Xcode 8 in September 2016, the '=delete' on the 'is' and 'as' methods introduced in 53c8e2cb causes the following errors for value.h:

    inline declaration of 'as<bool>' follows non-inline definition
    inline declaration of 'is<bool>' follows non-inline definition

etcetera for the other specializations of 'is' and 'as'.

I understand, of course, that this use of '=delete' serves a purpose: it allows the C++ compiler to detect when a program uses specializations of `is()` and `as()` that aren't implemented.  However, this patch will be useful (necessary, in fact) for people who build their projects with Xcode 8.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -21,6 +21,24 @@
 #endif
 #endif
 
+// Support for '= delete' with template declarations was a late addition
+// to the c++11 standard and is rejected by clang 3.8 and Apple clang 8.2
+// even though these declare themselves to be c++11 compilers.
+#if !defined(JSONCPP_TEMPLATE_DELETE)
+#if defined(__clang__) && defined(__apple_build_version__)
+#if __apple_build_version__ <= 8000042
+#define JSONCPP_TEMPLATE_DELETE
+#endif
+#elif defined(__clang__)
+#if __clang_major__ == 3 && __clang_minor__ <= 8
+#define JSONCPP_TEMPLATE_DELETE
+#endif
+#endif
+#if !defined(JSONCPP_TEMPLATE_DELETE)
+#define JSONCPP_TEMPLATE_DELETE = delete
+#endif
+#endif
+
 #include <array>
 #include <exception>
 #include <map>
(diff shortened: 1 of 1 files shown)


BillyDonahue:
Is the existing code correct C++ or not?
It's a little complex. A discussion is here:

https://stackoverflow.com/questions/33256589/specialized-template-function-with-deleted-general-case-fails-to-compile-with

Current draft spec includes the proposed language change: https://eel.is/c++draft/temp.expl.spec


If it's correct C++, then this is a toolchain workaround, which is fine, but based on the discussion regarding the 4-year-old XCode8 compatibility, I expected to find `#ifdefs` around the code changes.

Messing around to find out when clang stopped complaining about specialization of deleted primary function templates.
    https://gcc.godbolt.org/z/tF6_gG
Looks like around clang-3.9

If you don't want ifdefs, we could still do what we want, some other way.

dgobbi:
Thanks for the response and the insight.  I will amend the commit to add ifdefs for clang and Apple clang.  My omission of the ifdefs was to check your attachment to this particular use of '=delete'.

BillyDonahue:
XCode 8 is so old as to be a very special case, so be sure to check the version of XCode in your ifdefs, not just the vendor macros' existence, so that we're giving the same declarations everywhere that isn't old and weird.

BillyDonahue:
Looks good to me!

dgobbi:
Thanks for the rapid responses!

baylesj:
LGTM.

cdunn2001:
Wow. My instinct would have been to tell Xcode 8 users that they have to use an old release, maybe even `0.10.z`. But since a user did all the work, fine. It's just that all these ifdefs can get in the way later. This project is barely maintainable as it is.

We could easily re-break Xcode 8 someday. The only things we guarantee are the things we test, which are the TravisCI builds and the AppVeyor builds.

PR #1142, opened 2020-02-19
Use stream error flags instead of exception

mrjoel:
The operator>> overload should use stream error flags instead
of throwing an exception for error handling.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1973,7 +1973,7 @@ IStream& operator>>(IStream& sin, Value& root) {
   String errs;
   bool ok = parseFromStream(b, sin, &root, &errs);
   if (!ok) {
-    throwRuntimeError(errs);
+    sin.setstate(std::ios::failbit);
   }
   return sin;
 }


src/lib_json/json_reader.cpp
```diff
@@ -1973,7 +1973,7 @@ IStream& operator>>(IStream& sin, Value& root) {
   String errs;
   bool ok = parseFromStream(b, sin, &root, &errs);
   if (!ok) {
-    throwRuntimeError(errs);
+    sin.setstate(std::ios::failbit);
```
dota17: I think throwing error messages is useful to users. they should know some error messages.

cdunn2001:
I have to agree with @dota17. This is an interesting idea, but the error is usually in the content of the stream, not in the stream itself. In fact, I think an error on the stream would remain set, so if you trap the exception, you could still see the stream error, if any. I *think* you've misinterpreted the meaning of a stream error, but I could be wrong. Thanks for the PR anyway. It's very helpful that it's so small and specific.

PR #1150, opened 2020-03-24
Make `throw.*Error` print `msg` when built with JSON_USE_EXCEPTION=0

bcsgh:
Tested: used this to debug an issue in my code.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -13,6 +13,7 @@
 #include <cmath>
 #include <cstddef>
 #include <cstring>
+#include <iostream>
 #include <sstream>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


src/lib_json/json_value.cpp
```diff
@@ -211,8 +212,14 @@ JSONCPP_NORETURN void throwLogicError(String const& msg) {
   throw LogicError(msg);
 }
 #else // !JSON_USE_EXCEPTION
-JSONCPP_NORETURN void throwRuntimeError(String const& msg) { abort(); }
-JSONCPP_NORETURN void throwLogicError(String const& msg) { abort(); }
+JSONCPP_NORETURN void throwRuntimeError(String const& msg) {
+  std::cerr << msg << std::endl;
```
dota17: I'm afraid noreturn function should not return anything.

bcsgh: I'm not seeing the issue. These functions *don't* return.

They do have side-effects before aborting, but then there is nothing wrong with a no-return function having side effects.

BillyDonahue: We are a library and we don't control std::cerr. We must not reach into the app and decide what information should be put to std::cerr. The app might have a completely different logging system to which this error should be put. It's an app-level decision. The most we should do is allow for a user-defined error handling callback. Then if they want to write to `std::cerr` they can install the behavior to do so, but we should not be making that decision for them.

PR #1155, opened 2020-04-11
clang-tidy fixes again

neheb:
Easier to fun with meson.

### DIFF
--- a/src/jsontestrunner/main.cpp
+++ b/src/jsontestrunner/main.cpp
@@ -111,7 +111,7 @@ static void printValueTree(FILE* fout, Json::Value& value,
     Json::Value::Members members(value.getMemberNames());
     std::sort(members.begin(), members.end());
     Json::String suffix = *(path.end() - 1) == '.' ? "" : ".";
-    for (auto name : members) {
+    for (const auto& name : members) {
       printValueTree(fout, value[name], path + suffix + name);
     }
   } break;
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -2592,7 +2592,7 @@ JSONTEST_FIXTURE_LOCAL(StreamWriterTest, indentation) {
 JSONTEST_FIXTURE_LOCAL(StreamWriterTest, writeZeroes) {
   Json::String binary("hi", 3); // include trailing 0
   JSONTEST_ASSERT_EQUAL(3, binary.length());
-  Json::String expected("\"hi\\u0000\""); // unicoded zero
+  Json::String expected(R"("hi\u0000")"); // unicoded zero
   Json::StreamWriterBuilder b;
   {
     Json::Value root;
@@ -2866,7 +2866,7 @@ JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseWithNoErrors) {
   CharReaderPtr reader(b.newCharReader());
   Json::String errs;
   Json::Value root;
-  char const doc[] = "{ \"property\" : \"value\" }";
+  char const doc[] = R"({ "property" : "value" })";
   bool ok = reader->parse(doc, doc + std::strlen(doc), &root, &errs);
   JSONTEST_ASSERT(ok);
   JSONTEST_ASSERT(errs.empty());
@@ -2914,14 +2914,14 @@ JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseString) {
     JSONTEST_ASSERT_EQUAL("", root[0]);
   }
   {
-    char const doc[] = "[\"\\u8A2a\"]";
+    char const doc[] = R"(["\u8A2a"])";
     bool ok = reader->parse(doc, doc + std::strlen(doc), &root, &errs);
     JSONTEST_ASSERT(ok);
     JSONTEST_ASSERT(errs.empty());
     JSONTEST_ASSERT_EQUAL(u8"\u8A2a", root[0].asString()); // "訪"
   }
   {
-    char const doc[] = "[ \"\\uD801\" ]";
+    char const doc[] = R"([ "\uD801" ])";
     bool ok = reader->parse(doc, doc + std::strlen(doc), &root, &errs);
     JSONTEST_ASSERT(!ok);
     JSONTEST_ASSERT(errs == "* Line 1, Column 3\n"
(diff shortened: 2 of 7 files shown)


src/jsontestrunner/main.cpp
```diff
@@ -111,8 +111,8 @@ static void printValueTree(FILE* fout, Json::Value& value,
     Json::Value::Members members(value.getMemberNames());
     std::sort(members.begin(), members.end());
     Json::String suffix = *(path.end() - 1) == '.' ? "" : ".";
-    for (auto name : members) {
-      printValueTree(fout, value[name], path + suffix + name);
+    for (const auto& name : members) {
+      printValueTree(fout, value[name], path + suffix.append(name));
```
BillyDonahue, line 115: Wait, this didn't come from clang-tidy did it?
The suffix grows on every iteration of this loop now.
That isn't what the code was doing before.

src/lib_json/json_writer.cpp
```diff
@@ -159,7 +159,7 @@ String valueToString(double value, bool useSpecialFloats,
 
   // try to ensure we preserve the fact that this was given to us as a double on
   // input
-  if (buffer.find('.') == buffer.npos && buffer.find('e') == buffer.npos) {
+  if (buffer.find('.') == Json::String::npos && buffer.find('e') == Json::String::npos) {
```
BillyDonahue, line 162: Not an error. The thing we're looking for `npos` in is the thing we called `find` on.
Clang-tidy is wrong. The buffer might be a Json::String and it might be a std::string and it might be a std::string_view. We don't have to care about the decltype(b) on every line.

src/test_lib_json/main.cpp
```diff
@@ -122,14 +122,14 @@ struct ValueTest : JsonTest::TestCase {
 
 Json::String ValueTest::normalizeFloatingPointStr(const Json::String& s) {
   auto index = s.find_last_of("eE");
-  if (index == s.npos)
+  if (index == Json::String::npos)
```
BillyDonahue, line 125: not an error

src/test_lib_json/main.cpp
```diff
@@ -122,14 +122,14 @@ struct ValueTest : JsonTest::TestCase {
...
     return s;
   std::size_t signWidth = (s[index + 1] == '+' || s[index + 1] == '-') ? 1 : 0;
   auto exponentStartIndex = index + 1 + signWidth;
   Json::String normalized = s.substr(0, exponentStartIndex);
   auto indexDigit = s.find_first_not_of('0', exponentStartIndex);
   Json::String exponent = "0";
-  if (indexDigit != s.npos) { // nonzero exponent
+  if (indexDigit != Json::String::npos) { // nonzero exponent
```
BillyDonahue, line 132: not an error

src/test_lib_json/main.cpp
```diff
@@ -3122,9 +3122,9 @@ JSONTEST_FIXTURE_LOCAL(CharReaderStrictModeTest, dupKeys) {
   Json::CharReaderBuilder b;
   Json::Value root;
   char const doc[] =
-      "{ \"property\" : \"value\", \"key\" : \"val1\", \"key\" : \"val2\" }";
+      R"({ "property" : "value", "key" : "val1", "key" : "val2" })";
   {
-    b.strictMode(&b.settings_);
+    Json::CharReaderBuilder::strictMode(&b.settings_);
```
BillyDonahue, line 3127: These were using `b` deliberately so we can change the type of `b` without updating other lines.

src/test_lib_json/main.cpp
```diff
@@ -3286,11 +3286,11 @@ struct CharReaderAllowDropNullTest : JsonTest::TestCase {
     return [=](const Value& root) { JSONTEST_ASSERT_EQUAL(root, v); };
   }
 
-  ValueCheck objGetAnd(std::string idx, ValueCheck f) {
+  static ValueCheck objGetAnd(std::string idx, ValueCheck f) {
```
BillyDonahue: Why do we add static here?

The fact that the body of the function HAPPENS to not use any nonstatic members of the enclosing class is an implementation detail. I would say these helpers reserve the right to modify the instance of the object on which they are called. clang-tidy outsmarting itself here again.

src/jsontestrunner/main.cpp
neheb: The const did.

The append did not. clang-tidy did not autofix it but suggested std::append().

neheb:
Removed the static changes.

src/jsontestrunner/main.cpp
neheb: warning: string concatenation results in allocation of unnecessary temporary strings; consider using 'operator+=' or 'string::append()' instead [performance-inefficient-string-concatenation]

neheb:
Removed noexcept changes,

dota17:
I think the benefit of these macros is to maintain backward compatibility.

neheb:
This is currently not true as explicit and noexcept are already used in the code.

src/test_lib_json/main.cpp
```diff
@@ -3286,11 +3286,11 @@ struct CharReaderAllowDropNullTest : JsonTest::TestCase {
     return [=](const Value& root) { JSONTEST_ASSERT_EQUAL(root, v); };
   }
 
-  ValueCheck objGetAnd(std::string idx, ValueCheck f) {
+  ValueCheck objGetAnd(const std::string idx, const ValueCheck f) {
```
BillyDonahue, line 3289: I don't want to start adding `const` to locals just to do it. It's extra syntax and doesn't help anything. Code was really quite fine before change.

src/test_lib_json/main.cpp
```diff
@@ -3286,11 +3286,11 @@ struct CharReaderAllowDropNullTest : JsonTest::TestCase {
...
     return [=](const Value& root) { f(root.get(idx, true)); };
   }
 
-  ValueCheck arrGetAnd(int idx, ValueCheck f) {
+  ValueCheck arrGetAnd(int idx, const ValueCheck f) {
```
BillyDonahue, line 3293: revert

BillyDonahue:
This looks good to me.  Thanks for being patient!

PR #1162, opened 2020-04-28
allowBom -> skipBom

dota17:
Modify condition and the optional name.


dota17:
The original name `allowBom` is confusing, so I changed it.

PR #1171, opened 2020-05-12
clang-tidy + any_of usage

neheb:


src/lib_json/json_reader.cpp
```diff
@@ -77,10 +78,8 @@ Features Features::strictMode() {
 // ////////////////////////////////
 
 bool Reader::containsNewLine(Reader::Location begin, Reader::Location end) {
-  for (; begin < end; ++begin)
-    if (*begin == '\n' || *begin == '\r')
-      return true;
-  return false;
+  return std::any_of(begin, end,
+                     [](const char& b) { return b == '\n' || b == '\r'; });
```
BillyDonahue, line 82: `(auto b)`

Don't assume char unnecessarily and leave off the references.

src/lib_json/json_reader.cpp
```diff
@@ -998,10 +997,8 @@ class OurReader {
 
 bool OurReader::containsNewLine(OurReader::Location begin,
                                 OurReader::Location end) {
-  for (; begin < end; ++begin)
-    if (*begin == '\n' || *begin == '\r')
-      return true;
-  return false;
+  return std::any_of(begin, end,
+                     [](const char& b) { return b == '\n' || b == '\r'; });
```
BillyDonahue, line 1001: `(auto b)`

src/lib_json/json_writer.cpp
```diff
@@ -176,14 +177,11 @@ String valueToString(bool value) { return value ? "true" : "false"; }
...
 
-  char const* const end = s + n;
-  for (char const* cur = s; cur < end; ++cur) {
-    if (*cur == '\\' || *cur == '\"' ||
-        static_cast<unsigned char>(*cur) < ' ' ||
-        static_cast<unsigned char>(*cur) >= 0x80)
-      return true;
-  }
-  return false;
+  return std::any_of(s, s + n, [](const char& cur) {
```
BillyDonahue, line 180: `(unsigned char c)`

The name `cur` is longer and not helpfully so, since it's a 1-liner.
Save on some ugly static_casts. All comparisons we need can be done in the unsigned char domain.

src/lib_json/json_writer.cpp
```diff
@@ -1217,9 +1215,7 @@ bool StreamWriterBuilder::validate(Json::Value* invalid) const {
   std::set<String> valid_keys;
   getValidWriterKeys(&valid_keys);
   Value::Members keys = settings_.getMemberNames();
-  size_t n = keys.size();
-  for (size_t i = 0; i < n; ++i) {
-    String const& key = keys[i];
+  for (const auto& key : keys) {
```
BillyDonahue, line 1218: `getMemberNames()` is inefficient.
We should be able to do with without a `keys` vector.
Iterate directly through `settings_`, copying elements into `inv`.

src/lib_json/json_reader.cpp
neheb: auto lambdas are a C++14 feature. This is a C++11 project.

src/lib_json/json_reader.cpp
neheb: See above.

src/lib_json/json_reader.cpp
BillyDonahue: Oops, it's been a while. `char` then? Hey same # of letters.

neheb: I'd keep the const. Compilers do interesting things with const. I assume the reference type is fine.

src/lib_json/json_writer.cpp
neheb: I tried this patch:

```
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -1212,12 +1212,7 @@ bool StreamWriterBuilder::validate(Json::Value* invalid) const {
   Json::Value& inv = *invalid;
   std::set<String> valid_keys;
   getValidWriterKeys(&valid_keys);
-  Value::Members keys = settings_.getMemberNames();
-  for (const auto& key : keys) {
-    if (valid_keys.find(key) == valid_keys.end()) {
-      inv[key] = settings_[key];
-    }
-  }
+  std::copy_if(settings_.begin(), settings_.end(), inv.begin(), [=](const Json::Value& key){return valid_keys.find(key) == valid_keys.end();});
   return inv.empty();
 }
 Value& StreamWriterBuilder::operator[](const String& key) {
```

Not working.

src/lib_json/json_reader.cpp
BillyDonahue: Really `char` is enough to say what we want. I like code that's short and to the point. The rest of the type decorations are extra code to read. There's nothing interesting that a compiler is going to do with const or reference decorations here. Taking refs off of register-sized types can make them more efficient, and doesn't count as an odr-use.

src/lib_json/json_writer.cpp
BillyDonahue: You can't do it with iterator-hiding algorithms, because *iterator doesn't expose object keys. You also can't insert at Value.begin() like that. It is not an insert_iterator.

Please pull this change for a working refactor of the validate function:

https://github.com/BillyDonahue/jsoncpp/commit/3b416b03ab533fa6c2418e2d0c8223031206ac49

neheb:
I think everything has been adressed.

src/lib_json/json_reader.cpp
```diff
@@ -1925,9 +1922,7 @@ bool CharReaderBuilder::validate(Json::Value* invalid) const {
   std::set<String> valid_keys;
```
BillyDonahue, line 1922: We should be able to apply the same static set technique here as in the Writer, and remove the getValidReaderKeys function.

src/lib_json/json_writer.cpp
```diff
@@ -176,14 +177,9 @@ String valueToString(bool value) { return value ? "true" : "false"; }
...
-  char const* const end = s + n;
-  for (char const* cur = s; cur < end; ++cur) {
-    if (*cur == '\\' || *cur == '\"' ||
-        static_cast<unsigned char>(*cur) < ' ' ||
-        static_cast<unsigned char>(*cur) >= 0x80)
-      return true;
-  }
-  return false;
+  return std::any_of(s, s + n, [](unsigned char c) {
+    return c == '\\' || c == '\"' || c < ' ' || c >= 0x80;
```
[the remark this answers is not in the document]
neheb: Done.

src/lib_json/json_reader.cpp
neheb: Done.

BillyDonahue:
Thanks, Rosen!

PR #1172, opened 2020-05-12
optimize JsonWriter::validate 

BillyDonahue:
Based on discussion started in #1171

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -1198,34 +1198,30 @@ StreamWriter* StreamWriterBuilder::newStreamWriter() const {
                                      endingLineFeedSymbol, usf, emitUTF8, pre,
                                      precisionType);
 }
-static void getValidWriterKeys(std::set<String>* valid_keys) {
-  valid_keys->clear();
-  valid_keys->insert("indentation");
-  valid_keys->insert("commentStyle");
-  valid_keys->insert("enableYAMLCompatibility");
-  valid_keys->insert("dropNullPlaceholders");
-  valid_keys->insert("useSpecialFloats");
-  valid_keys->insert("emitUTF8");
-  valid_keys->insert("precision");
-  valid_keys->insert("precisionType");
-}
+
 bool StreamWriterBuilder::validate(Json::Value* invalid) const {
-  Json::Value my_invalid;
-  if (!invalid)
-    invalid = &my_invalid; // so we do not need to test for NULL
-  Json::Value& inv = *invalid;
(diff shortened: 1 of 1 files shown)


src/lib_json/json_writer.cpp
```diff
@@ -1198,34 +1198,30 @@ StreamWriter* StreamWriterBuilder::newStreamWriter() const {
...
-  std::set<String> valid_keys;
-  getValidWriterKeys(&valid_keys);
-  Value::Members keys = settings_.getMemberNames();
-  size_t n = keys.size();
-  for (size_t i = 0; i < n; ++i) {
-    String const& key = keys[i];
-    if (valid_keys.find(key) == valid_keys.end()) {
-      inv[key] = settings_[key];
-    }
+  static const auto& valid_keys = *new std::set<String>{
```
BillyDonahue, line 1203: There's no sense making a fresh valid_keys set each time validate is called.
Everybody can share one.

src/lib_json/json_writer.cpp
```diff
@@ -1198,34 +1198,30 @@ StreamWriter* StreamWriterBuilder::newStreamWriter() const {
...
+      "precisionType",
+  };
+  for (auto si = settings_.begin(); si != settings_.end(); ++si) {
+    auto key = si.name();
+    if (valid_keys.count(key))
+      continue;
+    if (invalid)
+      (*invalid)[std::move(key)] = *si;
+    else
+      return false;
```
BillyDonahue, line 1220: If `invalid==nullptr`, then the caller will only get a pass/fail result.
We might as well return as soon as we know the answer.

src/lib_json/json_writer.cpp
```diff
@@ -1198,34 +1198,30 @@ StreamWriter* StreamWriterBuilder::newStreamWriter() const {
...
-  valid_keys->insert("enableYAMLCompatibility");
-  valid_keys->insert("dropNullPlaceholders");
-  valid_keys->insert("useSpecialFloats");
-  valid_keys->insert("emitUTF8");
-  valid_keys->insert("precision");
-  valid_keys->insert("precisionType");
-}
+
 bool StreamWriterBuilder::validate(Json::Value* invalid) const {
-  Json::Value my_invalid;
```
BillyDonahue, line 1213: This object is always a waste.
In the `invalid==nullptr` case, we populate this Value just to throw it away.
In the `invalid!=nullptr` case, we create it and never use it.

BillyDonahue:
Subsumed by PR #1171.

Issue #1175, opened 2020-05-19
Control characters are not escaped when emitUTF8 = true


TheStormN:
**Describe the bug**
In JsonCPP 1.9.2. There is a configuration option `emitUTF8` which can be used to force the library to not escape UTF-8 characters. However, when using it, the control characters are also not escaped which breaks the output json string and other software have issues parsing it. Currently to get non-escaped UTF-8 with escaped control characters, I had to downgrade to 1.8.3.

**To Reproduce**
```
Json::Value root(Json::objectValue);

root["unprintable"] = std::string{0x01, 0x02, 0x03, 0x10};

Json::StreamWriterBuilder builder;
builder.settings_["emitUTF8"] = true;
auto result_str = Json::writeString(builder, root);
```

**Expected behavior**
The `result_str` should be `"{"unprintable":"\u0001\u0002\u0003\u0010"}"` but I get the same characters as in the input string.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Meson - no idea, I'm using Conan
 - Ninja version - no idea, I'm using Conan

**Additional context**
Version 1.8.3 does not suffer from that issue.

TheStormN:
Ok it seems that it was easy to fix that bug. I submitted a PR. @dota17 @BillyDonahue It would be nice if we can get a new release with that fix, so I'll be able to fetch it from Conan and get back to the latest version.

PR #1176, opened 2020-05-19
Fixed control chars escaping with enabled emitUTF8

TheStormN:
Fixes https://github.com/open-source-parsers/jsoncpp/issues/1175

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -309,36 +309,28 @@ static String valueToQuotedStringN(const char* value, unsigned length,
     // Should add a flag to allow this compatibility mode and prevent this
     // sequence from occurring.
     default: {
-      if (emitUTF8) {
-        if (isControlCharacter(*c)) {
-          result += "\\u";
-          result += toHex16Bit(*c);
+      unsigned int codepoint =
+          emitUTF8 ? static_cast<unsigned char>(*c) : utf8ToCodepoint(c, end);
+      const unsigned int FIRST_NON_CONTROL_CODEPOINT = 0x20;
+      const unsigned int LAST_NON_CONTROL_CODEPOINT = 0x7F;
+      const unsigned int FIRST_SURROGATE_PAIR_CODEPOINT = 0x10000;
+
+      if ((codepoint < FIRST_NON_CONTROL_CODEPOINT) ||
+          (!emitUTF8 && (codepoint >= LAST_NON_CONTROL_CODEPOINT))) {
+        auto appendHexChar = [&result](unsigned ch) {
+          result.append("\\u").append(toHex16Bit(ch));
+        };
+        if (codepoint < FIRST_SURROGATE_PAIR_CODEPOINT) {
+          // codepoint is in Basic Multilingual Plane
+          appendHexChar(codepoint);
         } else {
-          result += *c;
+          // codepoint is not in Basic Multilingual Plane
+          codepoint -= FIRST_SURROGATE_PAIR_CODEPOINT;
+          appendHexChar(0xD800 + (codepoint >> 10));
+          appendHexChar(0xDC00 + (codepoint & 0x3FF));
         }
       } else {
-        unsigned int codepoint = utf8ToCodepoint(c, end);
-        const unsigned int FIRST_NON_CONTROL_CODEPOINT = 0x20;
-        const unsigned int LAST_NON_CONTROL_CODEPOINT = 0x7F;
-        const unsigned int FIRST_SURROGATE_PAIR_CODEPOINT = 0x10000;
-        // don't escape non-control characters
-        // (short escape sequence are applied above)
-        if (FIRST_NON_CONTROL_CODEPOINT <= codepoint &&
-            codepoint <= LAST_NON_CONTROL_CODEPOINT) {
-          result += static_cast<char>(codepoint);
-        } else if (codepoint <
-                   FIRST_SURROGATE_PAIR_CODEPOINT) { // codepoint is in Basic
-                                                     // Multilingual Plane
-          result += "\\u";
-          result += toHex16Bit(codepoint);
-        } else { // codepoint is not in Basic Multilingual Plane
-                 // convert to surrogate pair first
-          codepoint -= FIRST_SURROGATE_PAIR_CODEPOINT;
-          result += "\\u";
-          result += toHex16Bit((codepoint >> 10) + 0xD800);
-          result += "\\u";
-          result += toHex16Bit((codepoint & 0x3FF) + 0xDC00);
-        }
+        result += *c;
       }
     } break;
     }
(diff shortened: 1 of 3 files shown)


src/lib_json/json_writer.cpp
```diff
@@ -310,7 +310,12 @@ static String valueToQuotedStringN(const char* value, unsigned length,
     // sequence from occurring.
     default: {
       if (emitUTF8) {
-        result += *c;
+        if (isControlCharacter(*c)) {
```
dota17, line 313: Good catch.
But I think there is no need to add more method.
You can refer these code : https://github.com/nlohmann/json/blob/a861bdb81f1c5ef0adc270952001096b189dbe9e/include/nlohmann/detail/output/serializer.hpp#L406-#L428

`emitUTF` maybe not a good name.

TheStormN: From the code you are referring I guess you mean that I should embed that code as part of the current one which is doing the utf8 encoding?

And yeah, `encodeUTF8` would be much better name, but regardless of the encoding, control and special characters should be encoded/escaped.

TheStormN: Ok, I've squeezed it a bit, but `utf8ToCodepoint` should not be called when we don't want to do special processing for utf-8, because it offsets the pointer, and this leaved me no choice but to use the ternary if.

src/lib_json/json_writer.cpp
```diff
@@ -309,31 +309,26 @@ static String valueToQuotedStringN(const char* value, unsigned length,
...
           result += "\\u";
           result += toHex16Bit(codepoint);
-        } else { // codepoint is not in Basic Multilingual Plane
-                 // convert to surrogate pair first
-          codepoint -= FIRST_SURROGATE_PAIR_CODEPOINT;
+        } else {
+          // codepoint is not in Basic Multilingual Plane
           result += "\\u";
-          result += toHex16Bit((codepoint >> 10) + 0xD800);
+          result += toHex16Bit((codepoint >> 10u) + 0xD7C0u);
```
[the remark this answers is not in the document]
TheStormN: Well, the `const` wasn't the only reason, it was actually one operations (minus) less. I don't see what's the point to do it in 2 steps, but for the sake of this bugfix I don't care much so I will change it as you want.

src/lib_json/json_writer.cpp
```diff
@@ -309,31 +309,26 @@ static String valueToQuotedStringN(const char* value, unsigned length,
...
-        const unsigned int FIRST_SURROGATE_PAIR_CODEPOINT = 0x10000;
-        // don't escape non-control characters
-        // (short escape sequence are applied above)
-        if (FIRST_NON_CONTROL_CODEPOINT <= codepoint &&
-            codepoint <= LAST_NON_CONTROL_CODEPOINT) {
-          result += static_cast<char>(codepoint);
-        } else if (codepoint <
-                   FIRST_SURROGATE_PAIR_CODEPOINT) { // codepoint is in Basic
-                                                     // Multilingual Plane
+      const unsigned int codepoint = emitUTF8 ? *c : utf8ToCodepoint(c, end);
```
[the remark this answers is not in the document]
TheStormN: The code you've proposed does not pass the unit tests. I've tried to fix but it wasn't something obvious. If you can do it, I'll be glad to close this PR and let you fix it.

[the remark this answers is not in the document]
TheStormN: I did some rework on my own code. Check if you like it more now, or if you can fix your example.

src/lib_json/json_writer.cpp
```diff
@@ -309,31 +309,26 @@ static String valueToQuotedStringN(const char* value, unsigned length,
...
+        } else {
+          // codepoint is not in Basic Multilingual Plane
+          codepoint -= 0x10000;
+          appendHexChar(0xD800 + (codepoint >> 10));
+          appendHexChar(0xDC00 + (codepoint & 0x3FF));
         }
+      } else if (codepoint < 0x20) {
+        appendHexChar(codepoint);
+      } else {
+        result += static_cast<char>(codepoint);
```
dota17, line 331: Mostly used range is `FIRST_NON_CONTROL_CODEPOINT ` to `FIRST_SURROGATE_PAIR_CODEPOINT `, it should be the first to judge.

src/lib_json/json_writer.cpp
```diff
@@ -309,31 +309,26 @@ static String valueToQuotedStringN(const char* value, unsigned length,
     // Should add a flag to allow this compatibility mode and prevent this
     // sequence from occurring.
     default: {
-      if (emitUTF8) {
-        result += *c;
-      } else {
-        unsigned int codepoint = utf8ToCodepoint(c, end);
-        const unsigned int FIRST_NON_CONTROL_CODEPOINT = 0x20;
-        const unsigned int LAST_NON_CONTROL_CODEPOINT = 0x7F;
-        const unsigned int FIRST_SURROGATE_PAIR_CODEPOINT = 0x10000;
```
dota17, line 318: It is better to keep so that people know what they mean.

TheStormN: I've removed them based on @BillyDonahue suggestion. Now I reverted them back, but please lets no go back and forth all the time.

src/lib_json/json_writer.cpp
TheStormN: It is well efficient as it is. I know it might not be very obvious because tabs are rendered as only 2 spaces but I will give a quick explanation.


- First topmost `if` is taking care of any UTF-8 symbols(any character larger than 127) in case encoding is desired.
- In case encoding is not desired, the `else if` is triggered which will take care of encoding any non-printable characters and since an `unsigned` is used there is no need for lower bounds check.
- Lastly, any other symbol is directly added to the `result` which may include non-encoded UTF-8 characters is case `emitUTF8` is true.

src/lib_json/json_writer.cpp
BillyDonahue: I don't think these need names, particularly names with that style. The 0x1000 is really just a bit pattern. It would be like naming a variable ONE_HUNDRED or something. Frankly I'd rather use comments than variable names if there's risk of confusion for readers. These are locals and make no difference beyond the body of this function.

dota17: For me, I just need to know what they mean. So it is ok to use comments.
I find you create a new PR for this. Both are great.

src/lib_json/json_writer.cpp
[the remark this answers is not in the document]
BillyDonahue: I fixed my example, at your invitation, (PR #1178)
I'd like to go with my version of the control flow here.

I also added a pretty good test we should integrate.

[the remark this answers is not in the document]
TheStormN: We all have our approaches and opinions on what's best. At first I wanted to fix the bug with minimal diff, as you can see in my first commit, later a several refactoring approaches were pointed to which I tried to comply...

Anyway I mostly care about the result, so as long as your PR is merged, this can be closed.

TheStormN:
Closed in favor of https://github.com/open-source-parsers/jsoncpp/pull/1178

PR #1178, opened 2020-05-21
Escape control chars even if emitting UTF8

### LINKED ISSUE #1175: Control characters are not escaped when emitUTF8 = true

BillyDonahue:
See #1176
Fixes #1175

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -262,6 +262,14 @@ static String toHex16Bit(unsigned int x) {
   return result;
 }
 
+static void appendRaw(String& result, unsigned ch) {
+  result += static_cast<char>(ch);
+}
+
+static void appendHex(String& result, unsigned ch) {
+  result.append("\\u").append(toHex16Bit(ch));
+}
+
 static String valueToQuotedStringN(const char* value, unsigned length,
                                    bool emitUTF8 = false) {
   if (value == nullptr)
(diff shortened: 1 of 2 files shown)


src/lib_json/json_writer.cpp
```diff
@@ -262,6 +262,14 @@ static String toHex16Bit(unsigned int x) {
   return result;
 }
 
+static void appendRaw(String& result, unsigned ch) {
+  result += static_cast<char>(ch);
+}
+
+static void appendHex(String& result, unsigned ch) {
```
dota17, line 269: Can we use `inline ` method?

BillyDonahue: I think it made a difference 20 years ago, but it makes no difference for a function like this. I try hard to avoid adding keywords that have no effect. You can try it on gcc.godbolt.org or something. The optimizers are amazing.

src/lib_json/json_writer.cpp
```diff
@@ -309,31 +317,25 @@ static String valueToQuotedStringN(const char* value, unsigned length,
     // Should add a flag to allow this compatibility mode and prevent this
     // sequence from occurring.
     default: {
+      unsigned codepoint;
```
TheStormN, line 320: You know, uninitialized variables are bad practice, even if in the current code a value will always be assigned, this might not be the case in future revisions. Even the compilers are issuing warnings for such variables.

src/lib_json/json_writer.cpp
```diff
@@ -309,31 +317,25 @@ static String valueToQuotedStringN(const char* value, unsigned length,
...
-          result += toHex16Bit((codepoint >> 10) + 0xD800);
-          result += "\\u";
-          result += toHex16Bit((codepoint & 0x3FF) + 0xDC00);
-        }
+        codepoint = utf8ToCodepoint(c, end); // modifies `c`
+      }
+
+      if (codepoint < 0x20) {
+        appendHex(result, codepoint);
+      } else if (codepoint < 0x80 || emitUTF8) {
```
TheStormN, line 329: You have double checking for `emitUTF8`. First on initialization and second here. My PR had only one such check, Not a big deal, just mentioning.

src/test_lib_json/main.cpp
```diff
@@ -2640,6 +2640,56 @@ JSONTEST_FIXTURE_LOCAL(StreamWriterTest, unicode) {
...
+      return "\\r";
+    case '\t':
+      return "\\t";
+    default:
+      return nullptr;
+    }
+  };
+
+  Json::StreamWriterBuilder b;
+  b.settings_["emitUTF8"] = true;
```
TheStormN, line 2676: You can also add a validation with that setting disabled, just to make sure the encoding is proper and prevent future regressions.

src/lib_json/json_writer.cpp
BillyDonahue: I'm not worried about it.
I know they're issuing warnings if the variable is read before initialization, but here the compiler should be able to prove codepoint to be safe. I don't want to have to invent some value like =0 here, that's IMO worse because it shows the reader a value that looks like it has a meaning but it doesn't. It also opens us up to the same class of maintenance bugs in it would be trying to avoid. If the variable somehow escapes without being overwritten, it will survive with the invalid sentinel value.  I could wrap the codepoint initializer in an immediately-invoked-lambda, but I think that's just too clever for this kind of code.

src/lib_json/json_writer.cpp
```diff
@@ -309,31 +317,25 @@ static String valueToQuotedStringN(const char* value, unsigned length,
...
+        appendRaw(result, codepoint);
+      } else if (codepoint < 0x10000) {
+        // Basic Multilingual Plane
+        appendHex(result, codepoint);
+      } else {
+        // Extended Unicode. Encode 20 bits as a surrogate pair.
+        codepoint -= 0x10000;
+        appendHex(result, 0xd800 + ((codepoint >> 10) & 0x3ff));
```
TheStormN, line 337: Using ` & 0x3ff` is not really need for the first code point.

src/lib_json/json_writer.cpp
TheStormN: `= 0` is not needed, you can just use braced initialization `{}` which will not give wrong impression to the reader.
What you say does have a point, but in case the variable gets used without an initialization, you will be having buggy random behavior(based on what random value it have), while when initialized you will gave buggy, but fixed behavior. :)

src/lib_json/json_writer.cpp
TheStormN: I agree that optimizes are amazing these days, and yet again I will do my best to not rely on the optimizer if that will not make the code looks bad. Using `inline` or `constexpr` would not make the code worse here. Optimizer is good, but is not part of any standard, so you never can be sure what and where it will optimize given tons of different target architectures and several different compilers.

src/test_lib_json/main.cpp
BillyDonahue: Done.

src/lib_json/json_writer.cpp
BillyDonahue: Yeah I don't like it either. It's clearer to separate the emitUTF8 path. Thanks.

src/lib_json/json_writer.cpp
BillyDonahue: Defense in depth, here. It's an interesting question. We can only write 20 bits of the 32 bit codepoint, as a limitation of json's encoding. We would need to know that codepoint has no bits above bit 20 set, or the UTF-8 output would be broken.

PR #1180, opened 2020-05-28
Fixes Oss-Fuzz issue: 21916

kabeer27:


dota17:
@kabeer27  Could you please provide more detail information about heap-buffer-overflow error? 
At least, I didn't see any error report about this in [Oss-Fuzz](https://github.com/google/oss-fuzz).

As I can see, the `begin_` will not exceed `end_`, if `end_ - begin_ < 3`, it will not execute `begin_ += 3;`. Considering edge case, if `end_ - begin_ = 3`, that's to say there is nothing except 3 bytes bom characters, if we skip it, then `begin_ = end_`(that's when they're closest), at last, we parse an empty string, this  is the expected behaviour.

kabeer27:
@dota17 `strncmp(begin_, "\xEF\xBB\xBF", 3)` will not execute `begin_ += 3`, when there are <=3 bytes but what happens is `strncmp` is only given a pointer `begin_` with no size information, so it will still blindly read past the bounds provided by `end_` causing a read outside buffer, hence a heap-buffer-overflow, it is reading one extra byte comparing it with the given string then exiting the condition, it will not execute `begin_ += 3` but it will read that extra byte which is the problem.

You can reproduce the steps by running the fuzzer and you will hit this shallow bug in <= 1 second,
 
`Crash State: | `
`Json::OurReader::parse`
`Json::OurCharReader::parse`
`fuzz.cpp`

`Crash Type: | Heap-buffer-overflow READ 1`

dota17:
I am trying to run the fuzzer.
Besides, I just run the testcases in my local env:
```
  const std::string rawJson = ""; // "\xEF", "\xEF\xBB"
  const int rawJsonLength = static_cast<int>(rawJson.length());
  JSONCPP_STRING err;
  Json::Value root;
  Json::CharReaderBuilder builder;
  Json::CharReader* reader(builder.newCharReader());
  bool ok = reader->parse(rawJson.c_str(), rawJson.c_str() + rawJsonLength, &root, &err);
  if (!ok) {
    std::cout << "error : " << err << std::endl; 
  }
 delete reader;
```

Everything was fine in my local env. It didn't raise `heap buffer overflow`

kabeer27:
Are you compiling with ASan (Address Sanitizer flag)? 
If not then it will run normally and not detect the heap buffer overflow, so either enable asan while compiling or run the fuzzer.

To detect it using the fuzzer:
after you have cloned oss-fuzz, from /oss-fuzz dir
`python infra/helper.py build_image jsoncpp`
`python infra/helper.py build_fuzzers jsoncpp`
`python infra/helper.py run_fuzzer jsoncpp jsoncpp_fuzzer`

dota17:
I can't obtain some docker images, but I'v made some tests with `-fsanitize=address`, whether the parsed string's length is 0, or 3, or longer, the results are the same(no overflow in these test cases), even if I applied your patch.

so I need more information to see if it's really a bug, it's better to post some log or picture.

I can merge this path into master, after all, it's a small change, and it doesn't have a bad impact on other code.

kabeer27:
image
So basically it crashes as soon as it starts, i was working on improving the coverage of the fuzzer, as per the statistics this makes it crash 99.78%, and it was hitting this shallow bug on all those crashes, can you also let me know the active maintainers so that i can add them to the auto cc list of oss-fuzz project.yaml? The previous maintainer cdunn2001 seems to have asked to be removed from the auto cc list, and there is no one else on the list.

dota17:
you can add me into the list. see #1047

dota17:
Merged.  thanks for reporting!

PR #1183, opened 2020-05-30
Issue 1182: Fix fuzzing bug

baylesj:
This patch fixes a fuzzing bug by resolving a bad fallthrough in the
setComment logic.

The result is that we get a proper error instead of an assert, making
the library friendlier to use and less likely to cause issue for
consumers.

See related Chromium project bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=989851

Issue: 1182


BillyDonahue:
Looks good.

PR #1185, opened 2020-06-03
meson: Don't specifically look for python3

diizzyy:
Not all distributions provide Python as python3 and as Meson already depends on 3.5+ just use what Meson uses.
References: https://mesonbuild.com/Getting-meson.html
https://mesonbuild.com/Python-module.html#find_installation

Signed-off-by: Daniel Engberg <daniel.engberg.lists@pyret.net>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -73,7 +73,7 @@ if meson.is_subproject() or not get_option('tests')
   subdir_done()
 endif
 
-python = import('python').find_installation('python3')
+python = import('python').find_installation()
 
 jsoncpp_test = executable(
   'jsoncpp_test', files([


neheb:
ACK

baylesj:
Could we run into an issue where python2 and python3 are installed side by side (pretty common IMO), and this picks python2 incorrectly?

diizzyy:
No, https://mesonbuild.com/Getting-meson.html
"Meson is implemented in Python 3, and requires 3.5 or newer."
...and as for the function itself
https://mesonbuild.com/Python-module.html#find_installation
"That argument is optional, if not provided then the returned python installation will be the one used to run meson."

jtojnar:
`python3` does not refer to the binary name but the python version, see the [docs you linked](https://mesonbuild.com/Python-module.html#find_installation):

> If [`name_or_path` argument is] provided, it can be: […]
> * One of `python2` or `python3`: in either case, the module will try some alternative names: `py -2` or `py -3` on Windows, and `python` everywhere. In the latter case, it will check whether the version provided by the sysconfig module matches the required major version

I would like it very much if Meson just deprecated calling the method without arguments since Python version running Meson should be an implementation detail if we follow Meson’s [design goals](https://mesonbuild.com/FAQ.html#why-is-meson-not-just-a-python-module-so-i-could-code-my-build-setup-in-python):

> In addition to those reasons, not exposing Python or any other "real" programming language makes it possible to port Meson's implementation to a different language.

diizzyy:
The issue is that it doesn't work on systems that for instance keeps Python versioned (```python3.7``` etc) and since Meson itself already depends on Python 3.5 I don't see how it would accidently pickup python2 and still work.

Issue also discussed here:
https://github.com/mesonbuild/meson/issues/4608
https://bugzilla.gnome.org/show_bug.cgi?id=783198
https://github.com/GStreamer/gstreamer/blob/master/meson.build#L508

jtojnar:
Then that is a Meson bug and should be fixed there. Calling `find_intallation` without arguments has bugged me for quite some time so thanks for getting me to finally open [an issue](https://github.com/mesonbuild/meson/issues/7415).

jtojnar:
I am not worried about it picking up `python2` but about using a wrong `python3` installation (in our case, we run Meson with minimal Python). In this case [`runjsontests.py`](https://github.com/open-source-parsers/jsoncpp/blob/c8453d39d1d98ddafd15267d9d55223cdc48f2b1/test/runjsontests.py) should be probably fine but hidden coupling like this has a non-negligible chance of causing issues down the line.

cdunn2001:
Wow. What a crazy problem. @jtojnar is obviously correct (as he demonstrates in [libjcat](https://github.com/hughsie/libjcat/pull/33/files)), but until Meson is fixed this seems like a harmless change we can undo later. I'm going to merge this.

PR #1189, opened 2020-06-10
avoid isprint, because it is locale specific

### LINKED ISSUE #1187: Regression: isAnyCharRequiredQuoting() is broken on Windows in JsonCpp 1.9.3 for some locales.
**Describe the bug**
The tab character is not escaped on Windows with various locales. This bug is a regression in 1.9.3, it is not present in 1.9.2.

**To Reproduce**
```c++
setlocale(LC_ALL, "English_United States.1252");

Json::Value result(Json::objectValue);
result["test"] = "\tTest123";

assert(Json::writeString(builder, result) == "{\"test\":\"\\tTest123\"}" );
```

**Expected behavior**
The tab character should be escaped.

**Desktop:**
 - OS: Windows 10 1909
 - Meson version - Using Conan
 - Ninja version - Using Conan

**Additional context**
`isAnyCharRequiredQuoting()` uses `std::isprint()` which gives different results on Windows. The version of that function in 1.9.2 was working properly. Perhaps you can just revert it?

Update: It seems that the locale does not matter, to reproduce the bug you only need to have \t as a special character among the normal string.

Update 2: It seems that the locale does matter. Had to restart the PC to get proper results after changing the locale several times.

BillyDonahue:
`std::isprint` is locale-specific and the JSON-spec is not.
In particular, isprint('\t') is true in Windows CP1252.

Has bitten others, e.g. https://github.com/laurikari/tre/issues/64

Fixes #1187

### DIFF
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -179,7 +179,7 @@ static bool doesAnyCharRequireEscaping(char const* s, size_t n) {
   assert(s || !n);
 
   return std::any_of(s, s + n, [](unsigned char c) {
-    return c == '\\' || c == '"' || c < 0x20 || c > 0x7F
+    return c == '\\' || c == '"' || c < 0x20 || c > 0x7F;
   });
 }
 
--- a/src/lib_json/json_writer.cpp
+++ b/src/lib_json/json_writer.cpp
@@ -175,11 +175,11 @@ String valueToString(double value, unsigned int precision,
 
 String valueToString(bool value) { return value ? "true" : "false"; }
 
-static bool isAnyCharRequiredQuoting(char const* s, size_t n) {
+static bool doesAnyCharRequireEscaping(char const* s, size_t n) {
   assert(s || !n);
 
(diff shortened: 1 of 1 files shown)


TheStormN:
Btw, how about some unit tests(for Windows only) just to make sure this does not happen again?

BillyDonahue:
Good idea. I don't have a Windows dev setup though.
You came up with a good repro in the issue description, can you contribute that in unit test form? I can merge it into this PR.

TheStormN:
> Good idea. I don't have a Windows dev setup though.
> You came up with a good repro in the issue description, can you contribute that in unit test form? I can merge it into this PR.

Here you are: https://github.com/open-source-parsers/jsoncpp/pull/1190 :)

TheStormN:
@BillyDonahue You forgot to merge my PR into yours...

BillyDonahue:
My PR is the same as master right now, so it should be the same thing.
I decided it was easier to just do two PRs in sequence. I'll merge yours as soon as CI is through with it.

Issue #1170, opened 2020-05-09
Inconsistent package contents dependent on build system


AMDmi3:
**Describe the bug**
Using different build systems lead to different package contents:
* Meson
  - no CMake modules are generated
  - both shared and static libraries are possible with `--default-library=both`
* CMake:
  - CMake modules are generated
  - only either of shared or static libraries is possible
* additionally, shared libarary is installed as `libjsoncpp.so.1.9.2` with cmake and `libjsoncpp.so.22` with meson

This makes it much harder to use jsoncpp, as jsoncpp consumer may not rely on any inconsistent component present. For instance, a project which relies on CMake modules will fail to build on FreeBSD which uses meson to build jsoncpp.

**To Reproduce**
Build, install with cmake and meson, compare produced files

**Expected behavior**
Single build system, or consistent behavior with both build systems. At the very least, a way to build all components which consumers may requiered (cmake modules AND static library AND shared library).

mr700:
The only static/shared library problem with cmake can be solved by calling cmake two times and packaging the relevant files.

AMDmi3:
> The only static/shared library problem with cmake can be solved by calling cmake two times and packaging the relevant files.

Not really, this is usually solved with two separate targets for static and shared library.

dota17:
All the cmake support is user-provided, not official. We actually use meson in Travis CI.

> At the very least, a way to build all components which consumers may requiered (cmake modules AND static library AND shared library).

Welcome a PR.

dota17:
Related PR: #827 

I found that before this PR,  there are different targets for static and shared library. And the target name wasn't unique. Also, you could find these target in `1.8.4` or `0.10.7`.

> Not really, this is usually solved with two separate targets for static and shared library.

Thus, if we want to change the targets, please consider the influences of #827.
What is more, could we add [`OBJECT LIBRARY`](https://cmake.org/cmake/help/v3.8/command/add_library.html#object-libraries) except static and shared library?

PR #1197, opened 2020-07-03
Enhance cmake script

dota17:
User @AMDmi3 put forward a requirement in issue #1170 , that is to use cmake to compile both static and shared libraries. After some investigation, I found that this is a very practical usage scenario. At the same time, I also know a common software usage scenario: In a large software system, the JSON parsing library(like JsonCpp) is the basic library of the system. Client may only need the `xxx.o` files of these basic libraries, and then link them into a shared library(`yyy.so` file) to use. Therefore, I made this patch.

During this time, I noticed the PR #827, I think the names of static libraries and dynamic libraries should be different. When users call them externally, they should be very clear about what type of libraries they are using, and a different names can guarantee this.
So, basically, the part of this patch is a rollback to that patch.

Cc: @cdunn2001 @hjmjohnson

### DIFF
--- a/.travis_scripts/cmake_builder.sh
+++ b/.travis_scripts/cmake_builder.sh
@@ -66,7 +66,7 @@ cmake --version
 echo ${CXX}
 ${CXX} --version
 _COMPILER_NAME=`basename ${CXX}`
-if [ "${BUILD_TYPE}" == "shared" ]; then
+if [ "${LIB_TYPE}" = "shared" ]; then
   _CMAKE_BUILD_SHARED_LIBS=ON
 else
   _CMAKE_BUILD_SHARED_LIBS=OFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,19 +59,19 @@ if(CCACHE_EXECUTABLE)
     set(CMAKE_C_COMPILER_LAUNCHER "${CCACHE_EXECUTABLE}" CACHE PATH "ccache" FORCE)
 endif()
 
-project(JSONCPP
+project(jsoncpp
         # Note: version must be updated in three places when doing a release. This
         # annoying process ensures that amalgamate, CMake, and meson all report the
         # correct version.
         # 1. ./meson.build
         # 2. ./include/json/version.h
         # 3. ./CMakeLists.txt
-        # IMPORTANT: also update the JSONCPP_SOVERSION!!
+        # IMPORTANT: also update the PROJECT_SOVERSION!!
         VERSION 1.9.3 # <major>[.<minor>[.<patch>[.<tweak>]]]
         LANGUAGES CXX)
 
-message(STATUS "JsonCpp Version: ${JSONCPP_VERSION_MAJOR}.${JSONCPP_VERSION_MINOR}.${JSONCPP_VERSION_PATCH}")
-set(JSONCPP_SOVERSION 24)
+message(STATUS "JsonCpp Version: ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
+set(PROJECT_SOVERSION 24)
 
 option(JSONCPP_WITH_TESTS "Compile and (for jsoncpp_check) run JsonCpp test executables" ON)
 option(JSONCPP_WITH_POST_BUILD_UNITTEST "Automatically run unit-tests as a post build step" ON)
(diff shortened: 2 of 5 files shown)


gizmocuz:
@dota17 , we are using this great library as a submodule in our project (https://github.com/domoticz/domoticz)
Since this change it seems only the shared libary is build, we would like to have the static library

```
[  1%] Building CXX object extern/jsoncpp/src/lib_json/CMakeFiles/jsoncpp_lib.dir/json_reader.cpp.o
[  1%] Building CXX object extern/jsoncpp/src/lib_json/CMakeFiles/jsoncpp_lib.dir/json_value.cpp.o
[  1%] Building CXX object extern/jsoncpp/src/lib_json/CMakeFiles/jsoncpp_lib.dir/json_writer.cpp.o
[  1%] Linking CXX shared library ../../../../lib/libjsoncpp.so
[  1%] Built target jsoncpp_lib
```

https://travis-ci.org/github/domoticz/domoticz/builds/707882330

I tried adding the following lines at the top of our CMakeList.txt file, but it did not make a difference:

```
option(BUILD_SHARED_LIBS "Build jsoncpp_lib as a shared library." OFF)
option(BUILD_STATIC_LIBS "Build jsoncpp_lib as a static library." ON)
```

Could you tell us what to add ?

Maybe the option names could use a slight name change like BUILD_JSONCPP_STATIC_LIBS ?

Thanks in advance!

dota17:
@gizmocuz  Thank you for your timely feedback! It seems that you only need the static library, that ok, but we don't need to change the option name, cause I find real problem, I will fix it today.

gizmocuz:
@dota17 , thanks for your fast reply and fix ! Appreciated very much !!!

gizmocuz:
@dota17 , unfortunately this change only builds the shared library not the static library

I also tried adding the following lines at the top of our CMakeList.txt file, but no luck....

option(BUILD_SHARED_LIBS "Build jsoncpp_lib as a shared library." OFF)
option(BUILD_STATIC_LIBS "Build jsoncpp_lib as a static library." ON)

There is a small discussion about it here:
https://github.com/domoticz/domoticz/issues/4264

We are building from source following this tutorial:
https://www.domoticz.com/wiki/Build_Domoticz_from_source

dota17:
This is strange, in my local compile, when I turned off `BUILD_SHARED_LIBS`, then it only have libjsoncpp.a.

I think you should forced to set the option `BUILD_SHARED_LIBS` off in the top of your CMakeList.txt file by using `set`, rather than add this line option(BUILD_SHARED_LIBS "Build jsoncpp_lib as a shared library." OFF) in your CMake file, cause I'm adraid it can't effect the inner CMakeList.txt setting.

SpaceIm:
This PR will break projects relying on CMake imported target, which was `jsoncpp_lib`. Now we have two different imported targets: `jsoncpp_shared` and `jsoncpp_static`.
It's also worth noting that package managers like `conan` or `vcpkg` prefer build scripts building either static or shared, not both, and specifically having the same imported target name for static and shared (avoid delegating shared/static logic in downstream dependencies, since it's usually final app decision to control shared/static of each dependency of the dependency graph).

dota17:
This PR don't aim to let users use both of them at the same time, but provide these packages, as available choices.
`conan` or `vcpkg` can still pick their default kind of package without any change.

SpaceIm:
Correct, but what about breaking change regarding CMake imported target?

cdunn2001:
@SpaceIm , if you have an important fix, please submit it.

@dota17, of course we appreciate the help, but in general I avoid touching the cmake crap at all. I have found it nearly impossible to maintain with all the versions of cmake's history and all the crazy usage patterns. Cmake maintenance was far more time-consuming than the sum of all other issues. I concentrate on Meson, and I simply accept Cmake patches from users if they seem to know what they're doing.

> I think the names of static libraries and dynamic libraries should be different. When users call them externally, they should be very clear about what type of libraries they are using, and a different names can guarantee this.

I disagree with this. In UNIX, it is normal to have something like `/usr/lib/` with both static and shared libraries of the same name. The static library will not include versions, so that's a difference. A good way to select shared versions is to use **pkg-config** and `*.pc` files. (Meson actually generates those, in fact.) I haven't looked at your change closely because I honestly don't care about Cmake at all, not even slightly. But if the breaking change was to rename the library, I think that's a mistake. As for "imported targets", I have no idea.

dota17:
Yes, it's normal to have the same name for static/shared library packages, it is easy to fix, I'll fix it later :-)

SpaceIm:
The modification would be to revert this modification, see this issue: https://github.com/open-source-parsers/jsoncpp/issues/1276

cdunn2001:
Please submit a PR with a revert, or work with the users on #1271 which has been merged.

We honestly have no opinion on **cmake** changes. But I can tell you that supporting **cmake** is a never-ending problem. Most issues relate to **cmake**. It's unbelievable.

If you submit a fix and the tests pass, we'll merge it, even if it's a revert.

PR #1199, opened 2020-07-05
Fix generation of pkg-config file with absolute includedir/libdir.

lopsided98:
The CMake variables `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` are not guaranteed to be relative paths. If they are absolute, an invalid pkg-config file is generated.

See https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files for more information.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,8 @@ if(NOT DEFINED CMAKE_BUILD_TYPE AND NOT DEFINED CMAKE_CONFIGURATION_TYPES)
         "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel Coverage.")
 endif()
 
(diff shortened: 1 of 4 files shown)


dota17:
LGTM.

CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 # vim: et ts=4 sts=4 sw=4 tw=0
 
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
```
dota17, line 3: Can this line move down a bit? I think there should be a better place.

lopsided98: I moved it down. If you have somewhere specific you want it let me know.

dota17:
Thanks.

PR #1203, opened 2020-07-14
Hot fix for building static lib

dota17:
Fix the problem introduced in #1197


dota17:
All checks have passed, merging.

Issue #1202, opened 2020-07-09
Keep order of fields as they are added, not sorting alphabetically


fedormsv:
To use json files in a context where human readability makes sense, order of the fields insertion can make sense. As jsoncpp is sorting them alphabetically, it becomes hard to read the file and find related fields in ddifferent ends of the object.
For example 
Json::Value root;
root["name"] = "Test;"
root["host"] = "127.0.0.1";
root["port"] = 80;

would be nice to have printed as

{
"name" : "Test,
"host" : "127.0.0.1",
"port" : 80
}

and not as 

{
"host" : "127.0.0.1",
"name" : "Test,
"port" : 80
}

I actually succeeded to have this feature by replacing a container used by Json::Value with a custom one.

dota17:
Pull requests are welcome.

dota17:
Relate discussion : https://github.com/open-source-parsers/jsoncpp/issues/237

samsonsite1:
Can someone please point out exactly where in the json code the fields are being sorted alphabetically?
And what changes must be made to stop this from happening?

fedormsv:
It's std::map container used to store fields that sorts them. To solve the issue you need another container.
Take care: this map is also used for storing arrays elements, and there is a requrement to get an element with max key.

I solved the issue in our project creating custom container, something like LinkedMap.

BillyDonahue:
> Can someone please point out exactly where in the json code the fields are being sorted alphabetically?

It's implicated by this typedef for the Json::Value::ObjectValues map type:

    typedef std::map<CZString, Value> ObjectValues;

And `std::map` uses sorted keys. We currently have no option to track insertion order of the keys within a `Json::Value`.

It would be neat to let the Reader emit SAX parse events instead of just building up a Json::Value and returning it when it's finished. Then the parse and the storage of its results can be decoupled for special cases like yours. But that's perhaps a job for a different json library.

samsonsite1:
Thanks for the reply.  I haven't used std::map much to realize it's sorting keys while inserting them.

Is jsoncpp dependent on this sorting behavior in any way?

Because I'm willing to make a compromise.  I can use std::map, but only if I can change the rules.  I want to give priority to a small list of keywords, so they are always sorted at the top, while everything else is sorted as usual.  So, I would like to able to pass std::map my own comparison function.

Would that cause any problems?

fedormsv:
[...]

PR #1204, opened 2020-07-14
historic_map container to keep subvalues in order of insertion (#1202)

fedormsv:
Solution for keeping items in order of their insertions, not resorting alphabetically

### DIFF
--- a/include/json/historic_map.h
+++ b/include/json/historic_map.h
@@ -1,0 +1,430 @@
+#ifndef HISTORIC_MAP_H
+#define HISTORIC_MAP_H
+
+#include <list>
+#include <map>
+#include <iterator>
+#include <utility>
+
+template <class Key, 
+          class Value,
+          class Compare = std::less<Key>
+         >
+class historic_map
+{
+public:
+  typedef Key                                                   key_type;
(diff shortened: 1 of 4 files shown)


BillyDonahue:
`historic_map` doesn't conform to the style of the surrounding Json library, which is making it difficult to review.

I would have a lot to say about its implementation if we go forward with it, as there are several code problems. But I hope we can save that effort by agreeing that we don't need an entirely new from scratch container type. Containers are very hard to get right. I hope we can add a much smaller amount of code to retain insertion order in the Json::Value::ObjectValues map. There's only one declared instance of it, as a private member of Json::Value. So a lot of the generic boilerplate can go away.

Would need a deliberate test of the container, beyond the incidental expectation adjustments. We are carefully monitoring code coverage stats in the project.

I would like to bikeshed a better name. Java calls this sort of thing a LinkedMap. The name "historic" isn't quite right, as the map itself isn't recorded in a history.

I would also say that the users have been given an expectation of sorted object keys. If we were to provide an ability to iterate by insertion order, that would have to be done in an opt-in way, preserving the existing lexicographical order behavior as a default.

fedormsv:
till further refactoring

PR #1212, opened 2020-08-19
include/json: check MSVC before adding pragma warning directive

Biswa96:


include/json/writer.h
```diff
@@ -360,8 +360,8 @@ JSON_API OStream& operator<<(OStream&, const Value& root);
 
 #pragma pack(pop)
 
-#if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING)
+#if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING) && defined(_MSC_VER)
 #pragma warning(pop)
-#endif // if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING)
+#endif // if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING) && defined(_MSC_VER)
```
BillyDonahue, line 365: Let's just delete the comments on these `#endif` statements if the block is small and leafy like this. It's lame to keep them up to date as the `#if` condition evolves.

BillyDonahue: I just looked at json/src/config.h and we already do what I'm requesting.
The JSONCPP_DISABLE_DLL_INTERFACE_WARNING is only defined if _MSC_VER is defined.
This is by design. I don't think your PR will improve anything. Do you have a motivating case like a build failure you can provide?

Biswa96:
I have change the code and add a little explanation.

include/json/config.h
```diff
@@ -34,18 +34,21 @@
...
 #endif // if defined(_MSC_VER)
 
 #elif defined(JSON_DLL)
 #if defined(_MSC_VER) || defined(__MINGW32__)
 #define JSON_API __declspec(dllimport)
-#define JSONCPP_DISABLE_DLL_INTERFACE_WARNING
 #endif // if defined(_MSC_VER)
 #endif // ifdef JSON_DLL_BUILD
 
+// Disable MSVC specific compiler warnings
```
BillyDonahue, line 47: This macro used to only  get defined when `JSON_DLL_BUILD` is defined.
I think it should stay that way, as it's clearly only relevant to DLL builds. Move it back up into the preceding `#if` block.

Biswa96: But it was defined in both `JSON_DLL_BUILD` and `JSON_DLL` sections. What's the issue with current change?

BillyDonahue: Sorry, I misread and thought JSON_DLL was nested inside JSON_DLL_BUILD.
So looking at the if/else structure here, we have 3 relevant modes. JSON_DLL, JSON_DLL_BUILD, and everything else.

I think JSONCPP_DISABLE_DLL_INTERFACE_WARNING should only be engaged when we're making a DLL target AND we're using a MSFT compiler. I think we should only disable the warnings when we expect to generate them and need to silence them. We don't need this warning silenced if JSON_API doesn't expand into a dll-related declspec. If we're not building with DLL interface declspecs, we are silencing a warning unnecessarily and we don't need to do that. So I think we should be extremely conservative in turning off warnings. So:

```
#if defined(_MSC_VER) && (defined (JSON_DLL) || defined(JSON_DLL_BUILD))
#define JSONCPP_DISABLE_DLL_INTERFACE_WARNING
#endif
```

?

Biswa96: But `JSONCPP_DISABLE_DLL_INTERFACE_WARNING` is defined in include/json files. So, if a project includes those header files MSVC compiler shows warnings due to jsoncpp.

baylesj:
It's still not clear to me what the actual error that is happening in MSVC, and how this fixes it. @Biswa96 can you show the compiler output?

Biswa96:
The error is not happening with MSVC, it's with GCC and/or Clang. First read the commit message. The `#pragma warning` are for MSVC, so my preprocessor changes check it first then enables it. Without it GCC show multiple warnings like this:

```sh
In file included from ../include/json/reader.h:11,
                 from ../src/lib_json/json_reader.cpp:10:
../include/json/value.h:52: warning: ignoring '#pragma warning ' [-Wunknown-pragmas]
   52 | #pragma warning(push)
      |
../include/json/value.h:53: warning: ignoring '#pragma warning ' [-Wunknown-pragmas]
   53 | #pragma warning(disable : 4251)
      |
../include/json/value.h:932: warning: ignoring '#pragma warning ' [-Wunknown-pragmas]
  932 | #pragma warning(pop)
      |
```

BillyDonahue:
> The error is not happening with MSVC, it's with GCC and/or Clang. First read the commit message. The `#pragma warning` are for MSVC, so my preprocessor changes check it first then enables it. Without it GCC show multiple warnings like this:
> 
> ```shell
> In file included from ../include/json/reader.h:11,
>                  from ../src/lib_json/json_reader.cpp:10:
> ../include/json/value.h:52: warning: ignoring '#pragma warning ' [-Wunknown-pragmas]
>    52 | #pragma warning(push)
>       |
> ../include/json/value.h:53: warning: ignoring '#pragma warning ' [-Wunknown-pragmas]
>    53 | #pragma warning(disable : 4251)
>       |
> ../include/json/value.h:932: warning: ignoring '#pragma warning ' [-Wunknown-pragmas]
>   932 | #pragma warning(pop)
>       |
> ```

You said here: https://github.com/open-source-parsers/jsoncpp/pull/1212#discussion_r475855672

> But JSONCPP_DISABLE_DLL_INTERFACE_WARNING is defined in include/json files. So, if a project includes those header files MSVC compiler shows warnings due to jsoncpp.

What MSVC compiler warnings are we talking about in THAT case?

Biswa96:
The warnings in `#pragma warning(disable)`.

include/json/config.h
baylesj: Looping back around to this. I think the confusion is that the warnings you are seeing are NOT DLL_INTERFACE_WARNINGs. This should really be renamed to:

JSONCPP_DISABLE_MSVC_WARNINGS

PR #1217, opened 2020-09-05
* PVS Studio fix

pavel-pimenov:
http://oracle.fly-server.ru/pvs-studio/jsoncpp/


BillyDonahue (changes requested):
looks like a lot of unnecessary changes, largely to c++14. Is there actually a "fix" in here?

src/lib_json/json_value.cpp
```diff
@@ -840,6 +840,7 @@ bool Value::isConvertibleTo(ValueType other) const {
            (type() == realValue && InRange(value_.real_, 0, maxUInt)) ||
            type() == booleanValue || type() == nullValue;
   case realValue:
+    return isNumeric() || type() == booleanValue || type() == nullValue;
```
BillyDonahue: We could also use the `[[fallthrough]]` attribute here to document the intent instead of duplicating code, and that would shut PVS Studio up.
https://en.cppreference.com/w/cpp/language/attributes/fallthrough

I think the duplication here is probably an improvement though. It's almost coincidental that convertibility to realValue and convertibility to booleanValue are the same condition.

src/lib_json/json_value.cpp
```diff
@@ -1396,7 +1397,7 @@ String Value::Comments::get(CommentPlacement slot) const {
 
 void Value::Comments::set(CommentPlacement slot, String comment) {
   if (!ptr_) {
-    ptr_ = std::make_unique<Array>();
+    ptr_ = std::unique_ptr<Array>(new Array());
```
BillyDonahue: I think we should make a version of `make_unique` in this file with internal linkage.

It can forward to `std::make_unique` if `__cplusplus` indicates that we're on C++14. Then we won't have to deal with the `make_unique` suggestion again. People are going to keep running jsoncpp through linters and sending PRs if we keep the C++11 raw `new` `unique_ptr` creation idiom in place. :)

BillyDonahue (changes requested):
Thanks for the persistence.
I think you're on the right track with the make_unique but we have to stay C++11 compatible, so we should abstract it internally.

src/lib_json/json_writer.cpp
```diff
@@ -1168,9 +1168,7 @@ StreamWriter* StreamWriterBuilder::newStreamWriter() const {
   const bool emitUTF8 = settings_["emitUTF8"].asBool();
   unsigned int pre = settings_["precision"].asUInt();
   CommentStyle::Enum cs = CommentStyle::All;
-  if (cs_str == "All") {
-    cs = CommentStyle::All;
-  } else if (cs_str == "None") {
+  if (cs_str == "None") {
```
baylesj, line 1171: This seems wrong. Am I missing something here?

src/test_lib_json/jsontest.cpp
```diff
@@ -408,7 +408,7 @@ Json::String ToJsonString(const char* toConvert) {
   return Json::String(toConvert);
 }
 
-Json::String ToJsonString(Json::String in) { return in; }
+Json::String ToJsonString(const Json::String& in) { return in; }
```
baylesj, line 411: You also need to modify the header:

```
  jsoncpp_static.vcxproj -> C:\projects\jsoncpp\lib\Release\jsoncpp_static.lib
main.obj : error LNK2019: unresolved external symbol "class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl JsonTest::ToJsonString(class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> >)" (?ToJsonString@JsonTest@@YA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@V23@@Z) referenced in function "public: virtual void __thiscall TestCharReaderAllowSingleQuotesTestissue182::runTestCase(void)" (?runTestCase@TestCharReaderAllowSingleQuotesTestissue182@@UAEXXZ) [C:\projects\jsoncpp\src\test_lib_json\jsoncpp_test.vcxproj]
```

baylesj (changes requested):

src/lib_json/json_writer.cpp
pavel-pimenov: https://github.com/open-source-parsers/jsoncpp/blob/ceae0e3867fe16e1227b4a39fe6951ee005591dc/src/lib_json/json_writer.cpp#L1170

baylesj:

Issue #1209, opened 2020-07-28
string copy operation in parse function is useless


ayrb13:
https://github.com/open-source-parsers/jsoncpp/blob/5be07bdc5e2d5b7715ecbc73749af3e625674dcb/src/lib_json/json_reader.cpp#L93
there is string copy operation while we use reader to parse json from std::string. I think it is useless, we can directly use document param, there is no need to copy it.
could someone explain it for me, thanks

ayrb13:
if we remove this copy operation, the parsing time cost will be reduced by 20%

dota17:
`Reader` class is deprecated, we highly recommand that you use `OurReader` class, which is better designed than `Reader`

fedormsv:
Still Json::parseFromStream is potentially doing extra copy: if I have an std::string to parse JSON from, I'm still forced to wrap it into stringstream, pass this stream to operator>> or parseFromStream function.  There the provided stream is directed to istringstream and finaly str() return a string, actually a copy of original one. Having parseFromString or even from string_view would save time and memory.

cdunn2001:
@fedormsv , please submit a PR. Could be a good idea. This sounds like it's not a bug, so we probably won't spend time on it ourselves.

fedormsv:
PR is submitted. I'm not sure why it gets new number. 
https://github.com/open-source-parsers/jsoncpp/pull/1221

It doesn't seem to be really welcomed.

PR #1221, opened 2020-09-16
#1209 add parseFromString to avoid unnecessary stringstream content copy

fedormsv:
functions for parsing directly from string, without unnecessary pass through string->stringstream->string pipeline

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -47,6 +47,7 @@ ds283 <D.Seery@sussex.ac.uk>
 Egor Tensin <Egor.Tensin@gmail.com>
 eightnoteight <mr.eightnoteight@gmail.com>
 Evince <baneyue@gmail.com>
+Fedor Moiseev <fedormsv@gmail.com>
 filipjs <filipjs@users.noreply.github.com>
 findblar <ft@finbarr.ca>
 Florian Meier <florian.meier@koalo.de>
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -359,6 +359,20 @@ public:
   static void strictMode(Json::Value* settings);
 };
 
+/** Consume entire string and use its begin/end.
+ * Someday we might have a real StreamReader, but for now this
+ * is convenient.
+ */
+bool JSON_API parseFromString(CharReader::Factory const&, char const* str, size_t len, Value* root,
+                                 String* errs);
+
+/** Consume entire string and use its begin/end.
+ * Someday we might have a real StreamReader, but for now this
+ * is convenient.
+ */
+bool JSON_API parseFromString(CharReader::Factory const&,const String&, Value* root,
+                              String* errs);
+
 /** Consume entire stream and use its begin/end.
  * Someday we might have a real StreamReader, but for now this
  * is convenient.
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1968,16 +1968,27 @@ void CharReaderBuilder::setDefaults(Json::Value* settings) {
 //////////////////////////////////
 // global functions
 
+bool parseFromString(CharReader::Factory const& fact, char const* str, size_t len, Value* root,
+                     String* errs) {
+    char const* begin = str;
+    char const* end = str + len;
+    // Note that we do not actually need a null-terminator.
+    CharReaderPtr const reader(fact.newCharReader());
+    return reader->parse(begin, end, root, errs);
+}
+
+bool parseFromString(CharReader::Factory const& fact, const String& doc, Value* root,
+                     String* errs) {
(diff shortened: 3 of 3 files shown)


BillyDonahue:
i don't see how this is an optimization.

it looks like just adding some new parse variants. I don't really want to make a parse variant for each storage format a user might have, so we have a stream based parse and we have a contiguous char array parse. We don't favor String as source of that buffer. Users are free to write wrappers.

fedormsv:
if I have a string to parse Json from, I should actually do the following (non throwing version):

```
std::string stringToParse;
std::istringstream iss(stringToParse);
Json::Value value;
Json::parseFromStream(builder, iss, &value, nullptr);

```
With the parseFromString being as following
```
bool parseFromStream(CharReader::Factory const& fact, IStream& sin, Value* root,
                     String* errs) {
  OStringStream ssin;
  ssin << sin.rdbuf();
  String doc = ssin.str();
  char const* begin = doc.data();
  char const* end = begin + doc.size();
  // Note that we do not actually need a null-terminator.
  CharReaderPtr const reader(fact.newCharReader());
  return reader->parse(begin, end, root, errs);
}

```
we actually arrive to the following: otside the parseFromStream we wrap a string to a stringstream. Inside parseFromStream we are redirecting it to another stringstream, from which it gets copied to doc object. The latter is then used to get begin and end of the string buffer.

BillyDonahue:
> if I have a string to parse Json from, I should actually do the following (non throwing version):
> 
> ```
> std::string stringToParse;
> std::istringstream iss(stringToParse);
> Json::Value value;
> Json::parseFromStream(builder, iss, &value, nullptr);
> ```

This isn't what a user with a string should do.
They should make a reader and call reader->parse with the string bounds:

    std::unique_ptr<Json::CharReader> reader(builder.newCharReader());

    reader->parse(
            stringToParse.data(),
            stringToParse.data() + stringToParse.size(),
            &value,
            nullptr);

parseFromStream is just capturing a common pattern for a convenient but inefficient dump of an istream. That pattern is inconvenient to formulate from scratch so we wrote it down, but it's not a primary API for Jsoncpp. The reader range parse is really the fundamental kind of parse, and a user with a string can use it readily.

> With the parseFromString being as following
> 
> ```
> bool parseFromStream(CharReader::Factory const& fact, IStream& sin, Value* root,
>                      String* errs) {
>   OStringStream ssin;
>   ssin << sin.rdbuf();
>   String doc = ssin.str();
>   char const* begin = doc.data();
>   char const* end = begin + doc.size();
>   // Note that we do not actually need a null-terminator.
>   CharReaderPtr const reader(fact.newCharReader());
>   return reader->parse(begin, end, root, errs);
> }
> ```
> 
> we actually arrive to the following: outside the parseFromStream we wrap a string to a stringstream. Inside parseFromStream we are redirecting it to another stringstream, from which it gets copied to doc object. The latter is then used to get begin and end of the string buffer.

fedormsv:
If I compare
```
std::unique_ptr<Json::CharReader> reader(builder.newCharReader());

reader->parse(
        stringToParse.data(),
        stringToParse.data() + stringToParse.size(),
        &value,
        nullptr);

```

with 
`
parseFromString(builder, stringToParse, &value, nullptr);
`

I will say length as well as one reference to stringToParse makes difference. 

And it makes even more sense comparing to now deprecated now deprecated API where it was possible to have 
`Json::Reader().parse(stringToParse, value);
`

So it evolved to dynamic allocation instead of stack, manual memory management (with or without smart pointers) and function call that needs refering the same string 3 times. 

Wrapping at least memory management and references to string seemed natural. 
I would probably refactor newCharReader to return unique_ptr, but I understand it's potentially problematic for DLL build.

fedormsv:
BTW, my PR was based on the suggestion to submit it in https://github.com/open-source-parsers/jsoncpp/issues/1209.
I'm ok with retracting

cdunn2001:
My only concern would be to avoid binary incompatible changes, and I don't see any of that. This should have a minor version bump though.

@BillyDonahue , you've looked at the code pretty closely. You can merge or close this.

BillyDonahue:
I'm gonna close it.

cdunn2001:
Sorry, @fedormsv. I _thought_ it sounded like a good idea. I didn't mean to waste your time. We do appreciate the help.

BillyDonahue:
Same here.

I just think that taking a `String`, one of our typedefs, is not quite right.
And std::string is convertible from other types, which would make a string copy happen anyway, and we were trying to avoid that. I feel we should emphasize the primary parse mechanism being a contiguous range of `const char*` and not get into how that range is stored or allocated. In C++17 a `std::string_view` would be a good choice but we aren't there yet.

Issue #1193, opened 2020-06-16
Volatile is deprecated in C++20


Troels51:
**Describe the bug**
In C++20 a lot of the volatile keyword is being deprecated ([P1152R3](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1152r3.html))
For example [deallocate](https://github.com/open-source-parsers/jsoncpp/blob/632044ad956b9ca55703d98d852bc815077f6afc/include/json/allocator.h#L41) has this problem

**To Reproduce**
1. Use a new compiler, fx gcc10
2. Set CMAKE_CXX_STANDARD to 20
3. Build

**Expected behavior**
Expected the build to complete

**Desktop (please complete the following information):**
- gcc version 10.1.0

**Additional context**
I ran into this problem because i am using jsoncpp in a project that was setting CXX_STANDARD to 20, and then when we upgraded gcc we were getting compile errors.
For example
```
jsoncpp-src/include/json/allocator.h:41:36: error: ‘volatile’-qualified parameter is deprecated [-Werror=volatile]
   41 |   void deallocate(volatile pointer p, size_type n) {
      |                   ~~~~~~~~~~~~~~~~~^
```

dota17:
I think we are far from ready to adapt to the C + + 20 standard. But if you like, you can submit a PR to fix this, and I'd love to review it then.

cdunn2001:
@dawesc, any opinion? This was originally your code, apparently.

https://en.cppreference.com/w/cpp/language/cv

> Some uses of volatile are deprecated: ...
>
> volatile object type as function parameter type or return type;

ledergec:
This seems to be the only reason which prevent to compile with c++20 standard using clang 10. The following patch fixes the issue:

diff --git a/include/json/allocator.h b/include/json/allocator.h
index d5f987e..b6c1fdf 100644
--- a/include/json/allocator.h
+++ b/include/json/allocator.h
@@ -35,11 +35,12 @@ public:
    * Release memory which was allocated for N items at pointer P.
    *
    * The memory block is filled with zeroes before being released.
-   * The pointer argument is tagged as "volatile" to prevent the
-   * compiler optimizing out this critical step.
+   * The pointer argument passed to memset is tagged as "volatile" to prevent
+   * the compiler optimizing out this critical step.
    */
-  void deallocate(volatile pointer p, size_type n) {
-    std::memset(p, 0, n * sizeof(T));
[...]

PR #1228, opened 2020-10-10
Fix c++20 compilation problem for clang10

ledergec:
This fixes the following issue:
https://github.com/open-source-parsers/jsoncpp/issues/1193

### DIFF
--- a/include/json/allocator.h
+++ b/include/json/allocator.h
@@ -35,11 +35,10 @@ public:
    * Release memory which was allocated for N items at pointer P.
    *
    * The memory block is filled with zeroes before being released.
-   * The pointer argument is tagged as "volatile" to prevent the
-   * compiler optimizing out this critical step.
    */
-  void deallocate(volatile pointer p, size_type n) {
-    std::memset(p, 0, n * sizeof(T));
+  void deallocate(pointer p, size_type n) {
+    // memset_s is used because memset may be optimized away by the compiler
+    memset_s(p, n * sizeof(T), 0, n * sizeof(T));
     // free using "global operator delete"
     ::operator delete(p);
   }


include/json/allocator.h
```diff
@@ -35,11 +35,12 @@ template <typename T> class SecureAllocator {
...
    * The memory block is filled with zeroes before being released.
-   * The pointer argument is tagged as "volatile" to prevent the
-   * compiler optimizing out this critical step.
+   * The pointer argument passed to memset is tagged as "volatile" to prevent
+   * the compiler optimizing out this critical step.
    */
-  void deallocate(volatile pointer p, size_type n) {
-    std::memset(p, 0, n * sizeof(T));
+  void deallocate(pointer p, size_type n) {
+    volatile pointer do_not_optimize_memset_p = p;
```
BillyDonahue, line 42: I really don't see how this ever prevented an optimizer from omitting the `memset` call.

Again this is just a `volatile` pointer to `T`, NOT a pointer to `volatile T`.

Pointer values are not modified by a `memset` call, so from the optimizer's PoV, it doesn't matter if the pointer changes or not. A local variable or function parameter (equivalently) that is not leaked outside the function has a provable scope of visible effects that the optimizer can exploit whether it's marked `volatile` or not. My opinion about this variable is that the `volatile` applied only to the parameter local and didn't affect the function signature before. Same as a `const` parameter is ignored.  It can be removed without an ABI break. Further, the `do_not_optimize_memset_p` has no effect and we might as well just memset p directly.

ledergec: What you write sounds reasonable to me. However, the codes intention is stated clearly in the comment. Hence, the question: is this actually a bug?

BillyDonahue: I think this may have been a "faith-based" use of `volatile`. It happens. It should be replaced with something that actually works.

I think we want the C11 `memset_s` on platforms on which it's available.
It is guaranteed not to be optimized away, and `memset` has no such guarantee, regardless of any irrelevant `volatile` qualifications on the pointer argument.

https://en.cppreference.com/w/c/string/byte/memset

ledergec: I updated the PR with memset_s. Indeed according to the c++ doc this should never get optimized away.

I hope my usage is correct.

Thanks for your comments.

ledergec:
I feel like this is ready to be merged? Anything blocking?

BillyDonahue:
Looks good to me.

BillyDonahue:
FYI: There are Github magic words that would make this issue auto-fix when the PR is merged.
So, rather than "This fixes the following issue: #1193", you'd say:

Fixes #1193

PR #1229, opened 2020-10-10
Try meson/ninja from pypi

cdunn2001:
On both linux and osx. If this works, we might be able to simplify.
However, we still want to test cmake, so there is only so much we
can simplify.

If this works, we might be able to stop using **homebrew**, if `clang-format` is available by default.


cdunn2001:
I am trying to figure out why **TravisCI** notification does not appear here, only **AppVeyor**. Possibly relevant: https://github.com/travis-ci/travis-ci/issues/10204

cdunn2001:
Ok, I see TravisCI now. I added a "Branch Protection Rule" here: https://github.com/open-source-parsers/jsoncpp/settings/branches

cdunn2001:
I guess we don't need to wait for appveyor as long as everything else is good.

PR #1230, opened 2020-10-12
Add example code of Visual Studio 2017 on windows OS.

gaozilai:
On windows OS， in the Microsoft Visual Studio 2017 IDE, open the solution file jsonTest.sln, directly compile and run.


dota17:
I don't think this is a good PR.

gaozilai:
maybe you are right.
but for the VS users, they don't need any extra operations. All they need to do is have a click.

example/vs2017/jsonTest/jsonTest/jsonTest.vcxproj.filters
```diff
@@ -0,0 +1,66 @@
...
+  <ItemGroup>
+    <Filter Include="源文件">
+      <UniqueIdentifier>{4FC737F1-C7A5-4376-A066-2A32D752A2FF}</UniqueIdentifier>
+      <Extensions>cpp;c;cc;cxx;def;odl;idl;hpj;bat;asm;asmx</Extensions>
+    </Filter>
+    <Filter Include="头文件">
+      <UniqueIdentifier>{93995380-89BD-4b04-88EB-625FBE52EBFB}</UniqueIdentifier>
+      <Extensions>h;hh;hpp;hxx;hm;inl;inc;ipp;xsd</Extensions>
+    </Filter>
+    <Filter Include="资源文件">
```
BillyDonahue, line 12: I don't understand what these say. I can't merge it.

BillyDonahue (changes requested):
As a Windows GUI app, this is not as useful as a portable console app would be.
We can't maintain the comments in Chinese.

We're already part of `vcpkg`, so it should be easy to get a demo app built without so much IDE filler data.
https://github.com/microsoft/vcpkg/tree/master/ports/jsoncpp

I think we need to close this one.

example/vs2017/jsonTest/jsonTest/jsonTest.vcxproj.filters
dota17: Even if I can understand these Chinese Characters, but they should never appear in the code or in the comments.

gaozilai:
ok i close it ,thanks

Issue #1215, opened 2020-09-04
Cannot parse Json


huddhudd:
Cannot parse {“A”:"黒"}

BillyDonahue:
Please be more specific.
That's supposed to work. Do you have a minimal repro?
Can you add it to a unit test to demonstrate the failure you're seeing?

huddhudd:
Please try parse     {“A”：“黒”} 

“黒”
There are hundreds of similar to this

dota17:
As the prompt information showed when you open this new issue, you need to provide some key information to tell us.
These information can help us solve issue.

----------------------------------------------------------------------------------------------------------------------------------
**Describe the bug**
A clear and concise description of what the bug is.

**To Reproduce**
Steps to reproduce the behavior:
1.
2.

**Expected behavior**
A clear and concise description of what you expected to happen.

**Desktop (please complete the following information):**
 - OS: [e.g. iOS]
 - Meson version
 - Ninja version

**Additional context**
Add any other context about the problem here.

gaozilai:
All the following Chinese characters cannot be parsed:
"黒", "玕", "盶", "砛", "芢" and so on, less than 255.
All they codes are including '0x5c'.

for example,  {"A":"芢"} , The data stream in memory is `7b 22 41 22 3a 22 c6 5c 22 7d 00`.
Originally c6 and 5c were together, now 5c and 22 are together. They become an escape character, `"` was missing.

BillyDonahue:
I feel like this problem is understood as an encoding issue now and can be closed?
Reopen if I'm wrong.

PR #1232, opened 2020-10-14
fix bug #1215

gaozilai:
fix bug #1215


src/lib_json/json_reader.cpp
```diff
@@ -1417,6 +1417,8 @@ bool OurReader::readString() {
   Char c = 0;
   while (current_ != end_) {
     c = getNextChar();
+    if (c < 0)
```
dota17, line 1420: This check isn't make any sense, type `char` will not less than zero.

gaozilai: Are you sure? What about `unsigned char`?

Actually, there was an error report without that. And it works fine with it.

BillyDonahue: > This check isn't make any sense, type `char` will not less than zero.


the `Reader::Char` typedef is just char. It is implementation defined whether char is signed or unsigned. I agree it's concerning to see it compared with c<0. On some platforms it's impossible. Others it's fine. So it's not portable code.

Note that char, signed char, and unsigned char are three distinct types in the C++ type system.

(chiming in from myphone)

gaozilai: In GBK or GB2312, Chinese characters are composed of two BYTEs, A BYTE greater than 127 followed by any BYTE.

BYTE is unsigned char.
`BYTE` than 127 is a `char` less than 0.

BillyDonahue: > In GBK or GB2312, Chinese characters are composed of two BYTEs, A BYTE greater than 127 followed by any BYTE.
> 
> 
> 
> BYTE is unsigned char.
> 
> `BYTE` than 127 is a `char` less than 0.
> 
> 

Ok there's your problem. We don't decode GBK. That's not a valid JSON string encoding. You will have to convert to UTF-8 before feeding to a JsonCpp Reader.

BYTE may have gone out as an unsigned char, but isn't unsigned char unless the receiver doing the decoding decides to look at it as an unsigned char.

gaozilai: Yes, you are very correct! Thank you!

dota17:
Thank you for your patience:) @gaozilai 
By the way, you should avoid making any changes in `master` branch, keep it clean and keep it track with upstream master branch, then if you want to submit PR, please checkout a new `dev` branch from the master branch. 
And if you've made some changes of code, it's better to write some test cases to make sure your changes really work.

gaozilai:
> Thank you for your patience:) @gaozilai
> By the way, you should avoid making any changes in `master` branch, keep it clean and keep it track with upstream master branch, then if you want to submit PR, please checkout a new `dev` branch from the master branch.
> And if you've made some changes of code, it's better to write some test cases to make sure your changes really work.

ok, got it. (Although I don't know how to do yet）
I am a svn user, work on the stand-alone machine. I learned so much, I knew nothing about git and json a week ago.
My practice is over, I will not make any changes anymore.

`jsoncpp` is perfect!
thank you

PR #1250, opened 2020-12-21
`Json::ValueIterator` operators `*` and `->` need to be const

### LINKED ISSUE #1249: `ValueIterator` operators `->` and `*` must be `const`
**Describe the bug**

As discovered in #1216, The C++20 `std::reverse_iterator<ValueIterator>` cannot perform `->`, at least not in GCC.
This is because `ValueIterator` operators `->` and `*` are `const`-incorrect.

**To Reproduce**

    Json::Value v;
    v["k1"] = "v1";
    std::reverse_iterator it(value.end());
    auto s = it->asString();  // This fails to compile, but `(*it).asString()`; is fine.

**Expected behavior**

`operator->` and `operator*` should be `const` in `ValueIterator` as they are in `ValueConstIterator`.
Dereferencing an iterator is always a logical const operation, as providing such access does not change the iterator.

 - OS: in my case, MacOS, using homebrew GCC-10.
 - Meson version N/A
 - Ninja version N/A

**Additional context**

A commit was made for this on #1216, but we could make it a standalone bugfix independent of C++20.
https://github.com/open-source-parsers/jsoncpp/commit/14e15b63d568b2461c48373d0afce810d3739d3c

BillyDonahue:
Fixes #1249.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -918,8 +918,8 @@ public:
    *  because the returned references/pointers can be used
    *  to change state of the base class.
    */
-  reference operator*() { return deref(); }
-  pointer operator->() { return &deref(); }
+  reference operator*() const { return const_cast<reference>(deref()); }
+  pointer operator->() const { return const_cast<pointer>(&deref()); }
 };
 
 inline void swap(Value& a, Value& b) { a.swap(b); }


cdunn2001:
I tried admin-privs for "rebase and merge" here for the first time. We use that at work all the time in BitBucket. I'll have to check whether it worked here as expected ...

Hmmm. It worked, but it didn't add this ticket number to the commit message. Not sure how important that is. Oh, well.

PR #1260, opened 2021-01-18
allow selection of legacy/new CXX11 ABI

yxlao:
Adds a CMake option to switch between old and new CXX11 ABI. The default behavior (using the new ABI) remains the same.

See: https://github.com/open-source-parsers/jsoncpp/blob/master/example/README.md

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -85,6 +85,7 @@ option(JSONCPP_WITH_STRICT_ISO "Issue all the warnings demanded by strict ISO C 
 option(JSONCPP_WITH_PKGCONFIG_SUPPORT "Generate and install .pc files" ON)
 option(JSONCPP_WITH_CMAKE_PACKAGE "Generate and install cmake package files" ON)
 option(JSONCPP_WITH_EXAMPLE "Compile JsonCpp example" OFF)
+option(JSONCPP_USE_CXX11_ABI "Set -D_GLIBCXX_USE_CXX11_ABI=1." ON)
 option(BUILD_SHARED_LIBS "Build jsoncpp_lib as a shared library." ON)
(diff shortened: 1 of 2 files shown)


cdunn2001:
You'll need to rebase this onto your own other PR (already merged).

Then, I guess there is no reason to object, though I don't know what the normal way to do this in cmake would be.

sergeyrachev:
CMake provides an abstraction layer to let developer support multiple compilers and target platforms. If your needs require to switch between ABI version you can pass compiler flags explicitly to CMake generation call with environment variable CXXFLAGS. Like `CXXFLAGS="-D_GLIBCXX_USE_CXX11_ABI=1" cmake <...>`
Or alternatively `cmake -DCMAKE_CXX_FLAGS="-D_GLIBCXX_USE_CXX11_ABI=1" <...>`

This approach allows to anyone who use the project to tune it with no inspection of how the project is built. CMake options are intended to affect the functionality of the project and generally should be more or less equal to any user. regardless target platform or prroject.

References: 
https://cmake.org/cmake/help/latest/envvar/CXXFLAGS.html
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html#variable:CMAKE_%3CLANG%3E_FLAGS
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_INIT.html#variable:CMAKE_%3CLANG%3E_FLAGS_INIT

cdunn2001:
@yxlao , what do you think of the idea from @sergeyrachev ?

yxlao:
@cdunn2001 Thanks, the new approach works for me. Closing this PR.

cdunn2001:
@yxlao, We just merged the PR from @sergeyrachev , with changes to **cmake** files. If that doesn't work for you, submit a new ticket and maybe mention that user, who has been very helpful.

PR #1265, opened 2021-02-04
Fix Value::resize to fill all array elements

### LINKED ISSUE #1264: Value::resize does not fill the array's elements
To implement array growth, `Value::resize` must create new empty array elements.
It does not do this completely. It only makes the last element. The interior elements of the array are left unpopulated.

The .size() will be reported to the expected value because that only looks at the key of the final element of the map, but this is misleading because the array is sparse. The sparseness is a clever optimization but iteration over the sparse array doesn't behave as expected. The sparse elements are just skipped by the iterator sequence.

 - OS: MacOS

Discovered while investigating #1253

BillyDonahue:
Fixes #1264

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -913,7 +913,7 @@ void Value::resize(ArrayIndex newSize) {
     clear();
   else if (newSize > oldSize)
     for (ArrayIndex i = oldSize; i < newSize; ++i)
-        (*this)[i];
+      (*this)[i];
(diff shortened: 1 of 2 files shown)


cdunn2001:
Oddly, #1266 says OSX compilation now has a warning after this fix. How did it get through our CI? Is TravisCI partly off? I know I saw a note about TravisCI moving VMs or something.

I don't have full internet access to check myself right now.

PR #1272, opened 2021-03-01
Needed for compiling with Embarcadero C++Builder 10.3, and probably w…

dacevedo:
Changes needed in order to compile with Embarcadero C++Builder 10.3.

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -80,8 +80,8 @@ extern JSON_API int msvc_pre1900_c99_snprintf(char* outBuf, size_t size,
 // These Macros are maintained for backwards compatibility of external tools.
 #if (defined(_MSC_VER) && _MSC_VER >= 1900) ||                                 \
     (defined(__GNUC__) && __cplusplus >= 201103L) ||                           \
+    (defined(__clang__) && __clang_major__ == 3 && __clang_minor__ >= 3)
 
-    (defined(__clang__) && (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 3)))
 #define JSONCPP_CXX_STD_11 1
 #else
 #define JSONCPP_CXX_STD_11 0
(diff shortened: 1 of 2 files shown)


src/lib_json/json_value.cpp
```diff
@@ -16,6 +16,7 @@
 #include <iostream>
 #include <sstream>
 #include <utility>
+#include <math.h>
```
BillyDonahue, line 19: why do we need both cmath and math.h?

dacevedo: Because Embarcadero C++Builder 10.3 complains that doesn't know about modf().
Maybe it could be improved with an #ifdef

BillyDonahue: What if we changed our `modf` call to `std::modf`? Would that help?
https://github.com/open-source-parsers/jsoncpp/blob/f2f5ce866ba4b0286b5d6ec7f595b5c0d6989381/src/lib_json/json_value.cpp#L1255

dacevedo: Indeed! It works with std::modf.
So #include <math.h> is not needed anymore.

include/json/config.h
```diff
@@ -80,8 +80,8 @@ extern JSON_API int msvc_pre1900_c99_snprintf(char* outBuf, size_t size,
 // These Macros are maintained for backwards compatibility of external tools.
 #if (defined(_MSC_VER) && _MSC_VER >= 1900) ||                                 \
     (defined(__GNUC__) && __cplusplus >= 201103L) ||                           \
-    (defined(__clang__) && __clang_major__ == 3 && __clang_minor__ >= 3)
 
+    (defined(__clang__) && (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 3)))
```
BillyDonahue, line 84: Need to remove the blank line above this line.
I think the blank line is going to ruin the continuity of the `#if`, and this won't do what it is supposed to do.

dacevedo: Of course. I just realized and was about to modify it.

PR #1277, opened 2021-03-17
Update json_value.cpp, fixing up spelling mistakes (boundry to boundary)

Guguant:
Here is a common spelling mistake that I've found while fixing up spelling mistakes.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -1401,7 +1401,7 @@ void Value::Comments::set(CommentPlacement slot, String comment) {
   if (!ptr_) {
     ptr_ = std::unique_ptr<Array>(new Array());
   }
-  // check comments array boundary.
+
   if (slot < CommentPlacement::numberOfCommentPlacement) {
     (*ptr_)[slot] = std::move(comment);
   }
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -1398,11 +1398,11 @@ String Value::Comments::get(CommentPlacement slot) const {
 }
 
 void Value::Comments::set(CommentPlacement slot, String comment) {
+  if (slot >= CommentPlacement::numberOfCommentPlacement) {
+    return;
+  }
   if (!ptr_) {
     ptr_ = std::unique_ptr<Array>(new Array());
-  }
-
-  if (slot < CommentPlacement::numberOfCommentPlacement) {
     (*ptr_)[slot] = std::move(comment);
   }
 }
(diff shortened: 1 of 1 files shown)


src/lib_json/json_value.cpp
```diff
@@ -1401,7 +1401,7 @@ void Value::Comments::set(CommentPlacement slot, String comment) {
   if (!ptr_) {
     ptr_ = std::unique_ptr<Array>(new Array());
   }
-  // check comments array boundry.
+  // check comments array boundary.
```
BillyDonahue, line 1404: Would also need capitalization. Let's just remove the comment.
It's fairly obvious why the if precedes the array element assignment.

Guguant: ok

src/lib_json/json_value.cpp
```diff
@@ -1401,7 +1401,7 @@ void Value::Comments::set(CommentPlacement slot, String comment) {
   if (!ptr_) {
     ptr_ = std::unique_ptr<Array>(new Array());
   }
-  // check comments array boundry.
+
```
BillyDonahue, line 1404: I don't really want the blank line either.
But while we're manipulating this function let's make the change more meaningful.

```c++
  if (slot >= CommentPlacement::numberOfCommentPlacement)
    return;
  if (!ptr_)
    ptr_ = std::unique_ptr<Array>(new Array());
  (*ptr_)[slot] = std::move(comment);
}
```

In other words, if we're not actually going to write to `ptr_`, we don't need to make one. We don't need to allocate it until we actually use it, and errant calls to `set` will then have no effect, which is a cleaner recovery I think.

Guguant: I agree with your opinion.

src/lib_json/json_value.cpp
```diff
@@ -1398,11 +1398,11 @@ String Value::Comments::get(CommentPlacement slot) const {
 }
 
 void Value::Comments::set(CommentPlacement slot, String comment) {
+  if (slot >= CommentPlacement::numberOfCommentPlacement) {
```
BillyDonahue, line 1401: This is not the code I suggested. Can you do it again?

Guguant: Yes. If return directly while `slot >= CommentPlacement::numberOfCommentPlacement`, the **CI** told that Some checks were not successful.

BillyDonahue: I'd be interested in the CI failure.

BillyDonahue:
I merged my version of this. Thanks for bringing it up.

PR #1282, opened 2021-04-08
Expose getLocationLineAndColumn() as a public API

KyleFromKitware:
`getLocationLineAndColumn()` is an internal API used by the error string code, but clients which use `Json::Value::getOffsetStart()` and `Json::Value::getOffsetLimit()` may want to use `getLocationLineAndColumn()` to turn the offset into a line and column. Expose the function as an external API.

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -392,6 +392,12 @@ bool JSON_API parseFromStream(CharReader::Factory const&, IStream&, Value* root,
  */
 JSON_API IStream& operator>>(IStream&, Value&);
 
+/** Get the line and column of a character within a string.
+ */
+void JSON_API getLocationLineAndColumn(char const* beginDoc, char const* endDoc,
+                                       ptrdiff_t location, int& line,
+                                       int& column);
+
 } // namespace Json
 
 #pragma pack(pop)
--- a/src/test_lib_json/main.cpp
+++ b/src/test_lib_json/main.cpp
@@ -3903,6 +3903,28 @@ JSONTEST_FIXTURE_LOCAL(FuzzTest, fuzzDoesntCrash) {
                              example.size()));
 }
 
+struct GetLocationLineAndColumnTest : JsonTest::TestCase {
+  void testLineAndColumn(const std::string& doc, ptrdiff_t location, int row,
+                         int column) {
+    int actualRow;
+    int actualColumn;
+    Json::getLocationLineAndColumn(doc.data(), doc.data() + doc.length(),
+                                   location, actualRow, actualColumn);
+    JSONTEST_ASSERT_EQUAL(row, actualRow);
+    JSONTEST_ASSERT_EQUAL(column, actualColumn);
+  }
+};
+
+JSONTEST_FIXTURE_LOCAL(GetLocationLineAndColumnTest, test) {
+  const std::string example = "line 1\nline 2\r\nline 3\rline 4";
+  testLineAndColumn(example, 0, 1, 1);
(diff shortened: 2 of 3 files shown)


src/test_lib_json/main.cpp
```diff
@@ -3903,6 +3903,28 @@ JSONTEST_FIXTURE_LOCAL(FuzzTest, fuzzDoesntCrash) {
...
+    int actualColumn;
+    Json::getLocationLineAndColumn(doc.data(), doc.data() + doc.length(),
+                                   location, actualRow, actualColumn);
+    JSONTEST_ASSERT_EQUAL(row, actualRow);
+    JSONTEST_ASSERT_EQUAL(column, actualColumn);
+  }
+};
+
+JSONTEST_FIXTURE_LOCAL(GetLocationLineAndColumnTest, test) {
+  const std::string example = "line 1\nline 2\r\nline 3\rline 4";
```
BillyDonahue, line 3919: This is not enough testing. There's only one input here.
There should be an exhaustive loop checking the results of every position.
This input only has one kind of newline.

include/json/reader.h
```diff
@@ -392,6 +392,12 @@ bool JSON_API parseFromStream(CharReader::Factory const&, IStream&, Value* root,
  */
 JSON_API IStream& operator>>(IStream&, Value&);
 
+/** Get the line and column of a character within a string.
```
BillyDonahue, line 395: It feels like a different general function that has nothing to do with JSON.

This isn't part of Reader's API, so I'm not sure it belongs in reader.h.

The out parameters should be a returned struct rather than int references, and the location should be a `size_t`, as it's encoding an array offset.

The documentation should be more complete about whether the line and column are 1 or 0 based, how it handles newlines, and how it handles unicode multibyte chars. I mean is 八 counted as one character or are all of its utf-8 bytes counted? If you're doing line and column, then you're implicitly talking about display position and it's a can of worms.

But really I have to question its role in the API. It doesn't really fit in.

BillyDonahue (changes requested):

KyleFromKitware:
I created this PR, along with #1281, because I would like to get more detailed line/column information from both the error (right now it only returns a string) and the values in case of success. This PR is not needed very badly, because clients that want to convert a `ptrdiff_t` to line/column can just copy this algorithm into their own code. I did the refactoring to expose this function as an API because it would seem silly to duplicate the code. We can close this if it's not a good fit.

#1281 is more important to me, because it exposes information that I can't get without parsing the returned error string.

BillyDonahue:
> I created this PR, along with #1281, because I would like to get more detailed line/column information from both the error (right now it only returns a string) and the values in case of success. This PR is not needed very badly, because clients that want to convert a `ptrdiff_t` to line/column can just copy this algorithm into their own code. I did the refactoring to expose this function as an API because it would seem silly to duplicate the code. We can close this if it's not a good fit.
> 
> #1281 is more important to me, because it exposes information that I can't get without parsing the returned error string.

The description field of this PR and #1281 are both empty. Can you write descriptions for them so we can understand the motivation?

KyleFromKitware:
> The description field of this PR and #1281 are both empty. Can you write descriptions for them so we can understand the motivation?

OK, done.

cdunn2001:
Let's re-open this after #1281 is done. I don't have any time to look at this myself, and #1281 is independent, right?

KyleFromKitware:
Yes, #1281 is independent of this.

baylesj:

PR #1284, opened 2021-04-14
CMake: conversion & sign-compare errors only if warnings as errors enabled

SpaceIm:
conversion & sign-compare errors should be enabled only if `JSONCPP_WITH_WARNING_AS_ERROR` is enabled also.


cdunn2001:
I merged another cmake change recently (#1271), so you'll need to rebase. After that, let's merge this.

PR #1289, opened 2021-05-05
Avoid getline(s, EOF)

### LINKED ISSUE #1288: Assumption that `EOF` is enclosed in parantheses (example: `(-1)`); otherwise compilation error
Previously discussed at https://github.com/DevSolar/pdclib/issues/20#issuecomment-824730642.

These 2 problems exist in jsoncpp:
1. `EOF` is used in a way that might break certain C implementations (in our case, pdclib for C + libcxx for C++, with clang as compiler). The code assumes that `EOF` will provide paranetheses for `dynamic_cast`:
    https://github.com/open-source-parsers/jsoncpp/blob/94cda30dbddc1859f111848fdd05dfb85d3287c7/src/lib_json/json_reader.cpp#L108
2. `EOF` isn't a `char`, but it's used as if it is. `EOF` is also not very strictly defined, so the behavior will be C implementation defined. Typically you get either a `char` (cast to `int`) *or* an `EOF` (`int`) as return value from functions like `getchar()`.

Point 1 is easily fixed, but 2 shows that it might not be necessary to fix it - the problematic code needs to be rewritten.

cdunn2001:
Fixes #1288

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -104,8 +104,7 @@ bool Reader::parse(std::istream& is, Value& root, bool collectComments) {
 
   // Since String is reference-counted, this at least does not
   // create an extra copy.
-  String doc;
-  std::getline(is, doc, static_cast<char> EOF);
+  String doc(std::istreambuf_iterator<char>(is), {});
   return parse(doc.data(), doc.data() + doc.size(), root, collectComments);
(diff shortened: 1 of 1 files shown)


cdunn2001:
Man, appveyor takes a long time. I guess we need it though. How else could we validate changes for MS VS?

PR #1299, opened 2021-06-10
CMake: Remove ancient version checks

ferdnyc:
The minimum version for the project is CMake 3.8.0 (as set in `cmake_minimum_required()` at the top of the root `CMakeLists.txt`), so there's no point in keeping legacy code for pre-3.0 or pre-2.8 CMake.


baylesj:
Good catch, always happy to see unnecessary code deleted!!

PR #1306, opened 2021-06-28
Create aliases with name space for static and shared library

mloy:
The previously, used add_library in the removed cmake file is something similar but it causes trouble when having jsoncpp included with cmake fetch_content.

Furthermore, alias is the simple way to go.


baylesj:

PR #1312, opened 2021-07-16
0.y.z

winnyrain:
When it parse a string that contain some special multibyte char, like Chinese "碶",  the parse method would return false. It is because the second byte is '\' 0x5c, the parser will skip one charactor .
 This bug  also appear in the other branch version.


baylesj:

PR #1333, opened 2021-09-21
Introduce CharReaderBuilder::ecma404Mode

axic:
Instead of a finished PR, this is an initial one to judge appetite for it.

I found out through testing that our assumption what strict means doesn't exactly match. Strictness according to ECMA-404 is a bit closer what we expected. For reference: [ECMA-404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) and [ECMA-262](https://tc39.es/ecma262/multipage/structured-data.html#sec-json.parse). Additionally [RFC4627](https://datatracker.ietf.org/doc/html/rfc4627#section-3) implies no BOM is needed.

If this feature is welcome, I can look into adding tests for it.


cdunn2001 (approved):
Please rebase.

I would suggest bumping the minor version number, but I doubt this will be a commonly used feature anyway. Bumping the micro number (and soversion) should suffice. We'll include this is the next release, probably `1.9.6`.

axic:
@cdunn2001 done!

meson.build
```diff
@@ -9,7 +9,7 @@ project(
   # 2. /include/json/version.h
   # 3. /CMakeLists.txt
   # IMPORTANT: also update the SOVERSION!!
-  version : '1.9.4',
```
axic, line 12: This was missed in c4904b2c0d461f84fba88a760161212651e4f536

axic:
@cdunn2001 let me know if there's anything else I need to do here?

axic:
@cdunn2001 @baylesj @BillyDonahue is anything else needed here?

axic:
Any chance to get this merged?

baylesj:
> Any chance to get this merged?

Thanks for waiting what turns out to have been an incredibly long time. Finally merging this now that I have fixed the CI system.

PR #1337, opened 2021-10-02
Add whitespaceOptions: colonBraceSameLine to settings

### LINKED ISSUE #1336: Braces/brackets on the same line
**Is your feature request related to a problem? Please describe.**
When printing JSON (`std::cout << json`, `StreamWriterBuilder`+`BuiltStyledStreamWriter`), if an object contains another object or an array, it looks like this:

```json
{
	"hello" : 
	{
		"world" : "!"
	},
	"world" : 
	[
		"!"
	]
}
```
while I want it to look like this:
```json
{
	"hello" : {
		"world" : "!"
	},
	"world" : [
		"!"
	]
}
```
Also, currently there's trailing whitespace after the colons which is annoying.

**Describe the solution you'd like**
To solve both problems at once, whenever the opening `{` or `[` of an object or array is printed after a colon of an object key:value pair, I want it to not move to a new line and indent and instead print it in the same line.

**Describe alternatives you've considered**
I've considered solving the trailing whitespace problem while keeping the brace/bracket the way it is, but it's easier this way and I prefer it.

**Additional context**
I think a setting should be added to `StreamWriterBuilder` called something like `whitespaceStyle`, and currently have two options: `colonBracesOnNewLine` and `colonBracesOnSameLine`.

This way instead of a boolean setting, in the future maybe one can add `whitespaceStyle: minimal` which would output the JSON as compactly as possible without any spaces. Or `commaBracesOnSameLine` or `bracketBraceOnSameLine` for having `{` be on the same line as a preceding `,` or `[` respectively.

I'm wondering whether combining values should be possible or not, for example using both `commaBracesOnSameLine` and `
bracketBraceOnSameLine` at the same time, so maybe `whitespaceStyle` can be an array. Not sure.

NotWearingPants:
Fixes #1336 by adding an array option called `"whitespaceOptions"` to `StreamWriterBuilder`, and a possible value called `"colonBraceSameLine"` which causes `{` and `[` following a colon to be placed in the same line.
The option is backed by a flag enum in `BuiltStyledStreamWriter`.

Example usage:

```cpp
builder.settings_["whitespaceOptions"] = Json::Value { Json::ValueType::arrayValue };
builder.settings_["whitespaceOptions"].append("colonBraceSameLine");

Json::Value json { Json::ValueType::objectValue };
json["hello"] = Json::Value { Json::ValueType::objectValue };
json["hello"]["world"] = "!";
json["world"] = Json::Value { Json::ValueType::arrayValue };
json["world"].append("!");

std::cout << Json::writeString(builder, json) << std::endl;
```
Result:
```json
{
        "hello": {
                "world": "!"
        },
        "world": [
                "!"
        ]
}
```


cdunn2001:
I like this. Thanks. Could use a test though.

Does anyone else have time to review this?

baylesj:
I'd be happy to review once there is test coverage associated with this.

include/json/value.h
```diff
@@ -123,6 +123,22 @@ enum CommentPlacement {
...
 };
 
+enum WhitespaceOptions {
+  colonBraceSameLine = 1 << 0, ///< braces just after colons on the same line
+};
+inline WhitespaceOptions operator|(
+  WhitespaceOptions left, WhitespaceOptions right) {
+    return static_cast<WhitespaceOptions>(static_cast<int>(left) | static_cast<int>(right));
+}
+inline void operator|=(
```
BillyDonahue, line 133: An `operator|=` should return `*this`.

NotWearingPants:
Rebased and fixed review comment

baylesj:
This needs to be formatted and have tests added.

PR #1339, opened 2021-10-05
Get decimal separator from locale if available. It is not always comma.

pepi124:
For example it is '/' for Persian language.


src/lib_json/json_tool.h
```diff
@@ -92,8 +92,13 @@ static inline void uintToString(LargestUInt value, char*& current) {
  * @see https://github.com/open-source-parsers/jsoncpp/pull/9
  */
 template <typename Iter> Iter fixNumericLocale(Iter begin, Iter end) {
+  char decimalPoint = getDecimalPoint();
```
cdunn2001, line 95: This should be `const`. But not a big deal.

pepi124:
Rebased, tests passed. I have no rights to merge this PR.

src/lib_json/json_tool.h
```diff
@@ -19,6 +19,8 @@
 #include <clocale>
 #endif
 
+#include <algorithm>
```
Youw, line 22: this is actually a heavy dependency to include just for a single (and simple) algorithm function to use

src/lib_json/json_tool.h
```diff
@@ -86,30 +88,25 @@ static inline void uintToString(LargestUInt value, char*& current) {
...
-  for (; begin != end; ++begin) {
-    if (*begin == ',') {
-      *begin = '.';
-    }
+  const char decimalPoint = getDecimalPoint();
+  if (decimalPoint != '\0' && decimalPoint != '.') {
+    std::replace(begin, end, decimalPoint, '.');
   }
-  return begin;
+  return end;
```
Youw, line 101: compared to a previous implementation it has an undesired side-effect:
if used locale-independent version of jsoncpp, the former version would replace ',' with '.' if jsons happen to have ',' as a separator
with the change - it will not happen if current locale happened to be `C` or something
could be non-backward-compatible side-effect

src/lib_json/json_tool.h
BillyDonahue: I don't think it matters enough to optimize around it.
A header like this is likely to be brought in by other headers anyway.

Youw: never directly and not but all compilers
(and it gets better each time)

src/lib_json/json_tool.h
BillyDonahue: This private helper function is only ever used to correct the small buffers produced by `snprintf` of individual numeric values with either `%.*g` or `%.*f` format specifiers.

The idea was that `snprintf` is locale-dependent and JSON isn't, so we're trying to undo the possibly inappropriate decimal point. We don't have to worry about thousands separators, in particular, because we controlled the format string and didn't use a `'` modifier.

To really fix this problem we would have to synchronize the `getDecimalPoint()` call.
We should also make sure that it doesn't change between the time that we call snprintf and the time that we try to fix up the decimal points with this function. This is all kind of unappetizing work, but this PR isn't a valid shortcut IMO. We should instead change the snprintf calls to something that doesn't depend on locale in the first place.

Because we're a library we can't control the application locale. We really need to just use a different formatter.
I think using a `std::ostrstream` with `imbue` is probably the most portable way to go.

There's also `std::to_chars` when the platform is C++17 or beyond.

BillyDonahue (changes requested):

src/lib_json/json_tool.h
Youw: > We should instead change the snprintf calls to something that doesn't depend on locale in the first place.

Sounds briliant.

> There's also `std::to_chars` when the platform is C++17 or beyond.

Tried that not so long ago: some compilers that mostly support C++17 might not have it, e.g. [libc++](https://libcxx.llvm.org/Status/Cxx17.html#id2) still has a partial support of it.

baylesj:

PR #1342, opened 2021-10-20
Fix SEGV crash caused by Value(s) appended to itself.

Ngiong:
Hi,
I found that we could get an infinite recursion of `dupPayload` call 
when we copy a `Value` using copy constructor on a `Value` that appended to itself.
We could make such `Value` by using the `append(Value&& value)` API, as shown in the following example:

```c++
Json::ValueType valuetype0 = Json::ValueType::nullValue;
Json::Value value1{valuetype0};
const Json::Value &value2 = value1.append(std::move(value1));
Json::Value value3 = value2; // infinite recursion here
```

Let me know if there is more I can help with this patch.
Thank you.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -1128,6 +1128,8 @@ Value& Value::append(const Value& value) { return append(Value(value)); }
 Value& Value::append(Value&& value) {
   JSON_ASSERT_MESSAGE(type() == nullValue || type() == arrayValue,
                       "in Json::Value::append: requires arrayValue");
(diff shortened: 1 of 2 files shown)


cdunn2001:
I agree with @BillyDonahue. If we handle this case, we _imply_ that we prevent all manners of this bug. But that would require cycle detection.

What this user needs is a function that would do a deep comparison between two value trees, which the user would then be responsible for calling ahead of any mutating operation. That would be computationally expensive.

A different library could include up-pointers to parents, which would prevent any value from living in two different trees. That's actually a very good idea, but out of scope for this library.

PR #1343, opened 2021-10-21
Fix parsing `[]` in JSONPath

ShooterIT:
There is no need `++` since we already `++current` on parsing number


baylesj:

Issue #1344, opened 2021-10-23
POSITION_INDEPENDENT_CODE and BUILD_SHARED_LIBS=OFF


fjtrujy:
**Describe the bug**
There is an error when building with `BUILD_SHARED_LIBS=OFF` that it is still trying to use `POSITION_INDEPENDENT_CODE ON`

**To Reproduce**
1. I'm doing a cross-compilation to Playstation 2/ Playstation Portable (MIPS CPUs).
2. Configuring with `cmake`
```
cmake -Wno-dev -DCMAKE_TOOLCHAIN_FILE=$PSPDEV/psp/share/pspdev.cmake -DCMAKE_INSTALL_PREFIX=$PSPDEV/psp/sdk -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DJSONCPP_WITH_POST_BUILD_UNITTEST=OFF -DJSONCPP_WITH_TESTS=OFF ..
```
3.  Compiling
```
make VERBOSE=1
```
4. Error
```
[ 14%] Building CXX object src/lib_json/CMakeFiles/jsoncpp_object.dir/json_reader.cpp.obj
cd /Users/fjtrujy/Projects/jsoncpp/build/src/lib_json && /Users/fjtrujy/toolchains/psp/pspdev-new/bin/psp-g++  -I/Users/fjtrujy/Projects/jsoncpp/include -I/Users/fjtrujy/Projects/jsoncpp/src/lib_json/../../include -I/Users/fjtrujy/Projects/jsoncpp/build/include/json -I/Users/fjtrujy/toolchains/psp/pspdev-new/psp/include -I/Users/fjtrujy/toolchains/psp/pspdev-new/psp/sdk/include -DPSP -O2 -G0 -O3 -DNDEBUG -fPIC -Wall -Wconversion -Wshadow -Wextra -Wpedantic -Werror=strict-aliasing -std=c++11 -o CMakeFiles/jsoncpp_object.dir/json_reader.cpp.obj -c /Users/fjtrujy/Projects/jsoncpp/src/lib_json/json_reader.cpp
cc1plus: error: cannot generate position-independent code for ‘-mabi=eabi’
make[2]: *** [src/lib_json/CMakeFiles/jsoncpp_object.dir/json_reader.cpp.obj] Error 1
make[1]: *** [src/lib_json/CMakeFiles/jsoncpp_object.dir/all] Error 2
make: *** [all] Error 2
```

[...]

PR #1347, opened 2021-10-25
Fix POSITION_INDEPENDENT_CODE

fjtrujy:
## Description
This PR just fixed a wrong configuration in the `cmake` files with the `POSITION_INDEPENDENT_CODE`, it should be just enabled when the `BUILD_SHARED_LIBS` config is enabled as well.

The specific details about the issue can be found here:
#1344 

Thanks

### DIFF
--- a/src/lib_json/CMakeLists.txt
+++ b/src/lib_json/CMakeLists.txt
@@ -119,7 +119,7 @@ if(BUILD_SHARED_LIBS)
         OUTPUT_NAME jsoncpp
         VERSION ${PROJECT_VERSION}
         SOVERSION ${PROJECT_SOVERSION}
-        POSITION_INDEPENDENT_CODE ON
+        POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS}
     )
 
     # Set library's runtime search path on OSX
@@ -180,7 +180,7 @@ if(BUILD_OBJECT_LIBS)
         OUTPUT_NAME jsoncpp
         VERSION ${PROJECT_VERSION}
         SOVERSION ${PROJECT_SOVERSION}
-        POSITION_INDEPENDENT_CODE ON
+        POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS}
     )
 
     # Set library's runtime search path on OSX


fjtrujy:
Hello @cdunn2001,
I have rebased the PR and now all the CI checks are green, could you merge the PR?

Thanks

cdunn2001:
I think we already test cmake in AppVeyor. But as long as the tests run fast, more robustness would be welcome.

src/lib_json/CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ if(BUILD_OBJECT_LIBS)
         OUTPUT_NAME jsoncpp
         VERSION ${PROJECT_VERSION}
         SOVERSION ${PROJECT_SOVERSION}
-        POSITION_INDEPENDENT_CODE ON
+        POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS}
```
SpaceIm, line 183: Does it really make sense? I would have remove it, that's all. I mean, BUILD_SHARED_LIBS is orthogonal to BUILD_OBJECT_LIBS.

src/lib_json/CMakeLists.txt
```diff
@@ -119,7 +119,7 @@ if(BUILD_SHARED_LIBS)
...
```
SpaceIm, line 122: it's the BUILD_SHARED_LIBS branch, so it's always ON in this block.

fjtrujy:
@SpaceIm about your suggestions..... in my opinion the best solution is just to put a root level, and just once this:

```
POSITION_INDEPENDENT_CODE ${BUILD_SHARED_LIBS}
```

I didn't want to do that change because I just took a quick way.

Thanks

SpaceIm:
I disagree. I like principle of least surprise:

- shared libs are always PIC
- static and object libs are not PIC by default (unless they are linked to a shared lib internally, therefore build system must manually force PIC).
- PIC can be enabled for object and static libs (and it's a consumer decision through standard CMAKE_POSITION_INDEPENDENT_CODE variable).

fjtrujy:
> I disagree. I like principle of least surprise:
> 
> * shared libs are always PIC
> * static and object libs are not PIC by default (unless they are linked to a shared lib internally, therefore build system must manually force PIC).
> * PIC can be enabled for object and static libs (and it's a consumer decision through standard CMAKE_POSITION_INDEPENDENT_CODE variable).

Ok, I'm fine with whatever solution meanwhile there is a way by flags of disabling the `CMAKE_POSITION_INDEPENDENT_CODE`.

Thanks

cdunn2001:
@SpaceIm , I didn't realize there was controversy. If you feel strongly about this, please submit a PR with a better solution. We try to let the **cmake** community reach consensus.

PR #1351, opened 2021-11-03
Drop compile-time deprecation warning

cdunn2001:
These are only getting in the way. Nobody needs to see these at compile-time (even though I wish people would stop using these).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,11 +72,11 @@ project(jsoncpp
         # 2. ./include/json/version.h
         # 3. ./CMakeLists.txt
         # IMPORTANT: also update the PROJECT_SOVERSION!!
-        VERSION 1.9.4 # <major>[.<minor>[.<patch>[.<tweak>]]]
+        VERSION 1.9.5 # <major>[.<minor>[.<patch>[.<tweak>]]]
         LANGUAGES CXX)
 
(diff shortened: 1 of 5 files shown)


cdunn2001:
I guess it's about time to tag a new release also.

CMakeLists.txt
```diff
@@ -72,11 +72,11 @@ project(jsoncpp
         # 2. ./include/json/version.h
         # 3. ./CMakeLists.txt
         # IMPORTANT: also update the PROJECT_SOVERSION!!
-        VERSION 1.9.4 # <major>[.<minor>[.<patch>[.<tweak>]]]
+        VERSION 1.9.5 # <major>[.<minor>[.<patch>[.<tweak>]]]
         LANGUAGES CXX)
 
 message(STATUS "JsonCpp Version: ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
-set(PROJECT_SOVERSION 24)
+set(PROJECT_SOVERSION 25)
```
scpeters, line 79: I'm wondering why the `SOVERSION` is incremented for a patch release that doesn't break API or ABI? This causes downstream packages to need to be rebuilt on macOS

* https://github.com/Homebrew/homebrew-core/pull/92195
* https://github.com/osrf/homebrew-simulation/issues/1780

I can open a separate issue to discuss this if that is a better venue

scpeters: I started a discussion about this in https://github.com/open-source-parsers/jsoncpp/discussions/1384

Issue #1349, opened 2021-10-28
Large floats not accepted anymore


terokinnunen:
**Describe the bug**
Version 1.9.2 no longer accepts `1+e9999` (or `2e+308` and anything above), regardless of `allowSpecialFloats` setting.

**To Reproduce**
```c++
#include <iostream>
#include <json/json.h>
#include <sstream>

int main(int argc, const char* argv[])
{
    Json::CharReaderBuilder reader;
    reader["allowSpecialFloats"] = true;  // Makes no difference

    std::string s = "1e+9999";
    Json::Value val;
    std::string errors;
    std::stringstream json(s);
    if (Json::parseFromStream(reader, json, &val, &errors)) {
        std::cout << s << " -> " << val.asDouble() << "\n";
    } else {
        std::cout << errors;
    }
    return 0;
}
```

**Expected behavior**
* `1e+9999 -> inf`

Actual behavior:
```
* Line 1, Column 1
  '1e+9999' is not a number.
```

**Desktop (please complete the following information):**
 - OS: Fedora 34
 - Meson version 0.59.1
 - Ninja version 1.10.2

**Additional context**
* Version 1.9.1 accepts value `1e+9999` as infinite, regardless of `allowSpecialFloats` setting. 
* Version 1.9.2 no longer accepts `2e+308` or above, regardless of `allowSpecialFloats` setting.

Specifically breaking change seems to be commit 645cd0412c "Number fixes (#1053)".

cdunn2001:
Sorry, I don't like to change/break old behavior. This was our mistake.

But now, I wonder whether we can recover the old behavior without reverting that large PR. Somehow, we need big numbers to map to `inf`. However, did the old behavior map even bigger, really huge numbers to `inf` also? If not, would it be ok if it did?

In that case, there is probably a relatively simply way to use `inf` for all large numbers. Could you submit a PR for that? Otherwise, I'm not sure when we would find time for this.

terokinnunen:
I'll give it a try to fix.

PR #1353, opened 2021-11-08
Parse large floats as infinity (#1349)

terokinnunen:
Return 1.9.1 functionality where values too large to fit in
double are converted to positive or negative infinity.

Commit 645cd04 changed functionality so that large floats cause
parse error, while version 1.9.1 accepted them as infinite.
This is problematic because writer outputs infinity values
as `1e+9999`, which could no longer be parsed back.

Fixed also legacy Reader even though it did not parse large values
even before breaking change, due to problematic output/parse asymmetry.

`>>` operator sets value to numeric_limits::max/lowest value if
representation is too large to fit to double. [1][2]

> | value in *val*           | description |
> |--------------------------|-------------|
> | numeric_limits::max()    | The sequence represents a value too large for the type of val |
> | numeric_limits::lowest() | The sequence represents a value too large negative for the type of val |

[1] https://www.cplusplus.com/reference/istream/istream/operator%3E%3E/
[2] https://www.cplusplus.com/reference/locale/num_get/get/


terokinnunen:
Mac clang meson static release testing Travis build seems to fail to familiar `'-1e+9999' is not a number.`, not sure why.

terokinnunen:
Fixed macos build. `>>` operator in macos seems to parse values too large to fit to double directly to infinity, unlike in Linux builds (in which value parses to numeric_limits::max/lowest). Accepting also infinite values as result.

PR #1374, opened 2021-12-26
Better support for modern CMake

Andres6936:
Se ha mejorado el soporte con CMake para permitir descargar el proyecto con Fetch, la configuración y el enlace se hace automáticamente para una mejor experiencia de desarrollo, el modo de compilación es dependiente de la opción utilizada y solamente se puede compilar SHARED o STATIC pero no ambas a la vez.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ compile_commands.json
 *.a
 
 # eclipse project files
+.idea/
 .project
 .cproject
 /.settings/
@@ -55,3 +56,4 @@ compile_commands.json
 
 # temps
 /version
+cmake-build-*/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,12 +40,6 @@ foreach(pold "") # Currently Empty
     endif()
 endforeach()
 
-# Build the library with C++11 standard support, independent from other including
-# software which may use a different CXX_STANDARD or CMAKE_CXX_STANDARD.
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-
 # Ensure that CMAKE_BUILD_TYPE has a value specified for single configuration generators.
 if(NOT DEFINED CMAKE_BUILD_TYPE AND NOT DEFINED CMAKE_CONFIGURATION_TYPES)
     set(CMAKE_BUILD_TYPE Release CACHE STRING
@@ -116,9 +110,6 @@ macro(use_compilation_warning_as_error)
         endif()
     endif()
 endmacro()
-
-# Include our configuration header
-include_directories(${jsoncpp_SOURCE_DIR}/include)
 
 if(MSVC)
     # Only enabled in debug because some old versions of VS STL generate
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,11 +98,6 @@ option(JSONCPP_BUILD_SHARED_LIBS "Build jsoncpp_lib as a shared library." OFF)
 # Adhere to GNU filesystem layout conventions
 include(GNUInstallDirs)
 
-set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib" CACHE PATH "Archive output dir.")
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib" CACHE PATH "Library output dir.")
-set(CMAKE_PDB_OUTPUT_DIRECTORY     "${CMAKE_BINARY_DIR}/bin" CACHE PATH "PDB (MSVC debug symbol)output dir.")
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin" CACHE PATH "Executable/dll output dir.")
-
 set(JSONCPP_USE_SECURE_MEMORY "0" CACHE STRING "-D...=1 to use memory-wiping allocator for STL")
 
 configure_file("${PROJECT_SOURCE_DIR}/version.in"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,9 @@ project(jsoncpp
 message(STATUS "JsonCpp Version: ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH}")
 set(PROJECT_SOVERSION 25)
 
+# Defines the project root to be used throughout all scripts.
+SET(JSONCPP_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+
 include(${CMAKE_CURRENT_SOURCE_DIR}/include/PreventInSourceBuilds.cmake)
 include(${CMAKE_CURRENT_SOURCE_DIR}/include/PreventInBuildInstalls.cmake)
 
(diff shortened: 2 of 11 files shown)


example/CMakeLists.txt
```diff
@@ -1,27 +1,6 @@
...
-endif()
-
-foreach(example ${EXAMPLES})
-    add_executable(${example} ${example}/${example}.cpp)
-    target_include_directories(${example} PUBLIC ${CMAKE_SOURCE_DIR}/include)
-    target_link_libraries(${example} jsoncpp_lib)
-endforeach()
-
-add_custom_target(examples ALL DEPENDS ${EXAMPLES})
+ADD_SUBDIRECTORY(readFromStream/)
```
hjmjohnson, line 3: Modern cmake preference is to use lowercase naming.

example/readFromString/CMakeLists.txt
```diff
@@ -1,3 +1,2 @@
-add_executable(readFromString readFromString.cpp)
-target_include_directories(readFromString PUBLIC ${CMAKE_SOURCE_DIR}/include)
-target_link_libraries(readFromString jsoncpp_lib)
\ No newline at end of file
+ADD_EXECUTABLE(readFromString readFromString.cpp)
+TARGET_LINK_LIBRARIES(readFromString PRIVATE jsoncpp::framework)
```
hjmjohnson, line 2:
```bash
#!/bin/bash

# \author Hans J. Johnson

this_script_name=$(basename $0)
CMTMSG=$(mktemp -q /tmp/${this_script_name}.XXXXXX)

cat > ${CMTMSG} << EOF
STYLE: Convert CMake-language commands to lower case

Ancient CMake versions required upper-case commands.  Later command names
became case-insensitive.  Now the preferred style is lower-case.
EOF

SED_BIN=$(which gsed)
if [[ -z "${SED_BIN}" ]];then
  SED_BIN=sed
fi


#Run the following shell code:
## NOTE: MUST USE GNU compliant version of sed
cmake --help-command-list \
| grep -v "cmake version" \
| while read c; do
    echo 's/\b'"$(echo $c | tr '[:lower:]' '[:upper:]')"'\(\s*\)(/'"$c"'\1(/g'
  done > /tmp/convert.sed \
&& git ls-files -z -- bootstrap '*.cmake' '*.cmake.in' '*CMakeLists.txt' '*.wrap' \
   | xargs -0 ${SED_BIN} -i -f  /tmp/convert.sed \
&& rm  /tmp/convert.sed

git add -A && git commit --file ${CMTMSG}
```

src/test_lib_json/CMakeLists.txt
```diff
@@ -1,39 +1,29 @@
 # vim: et ts=4 sts=4 sw=4 tw=0
 
-add_executable(jsoncpp_test
-    jsontest.cpp
-    jsontest.h
-    fuzz.cpp
-    fuzz.h
-    main.cpp
-)
+ADD_EXECUTABLE(jsoncpp_test
```
hjmjohnson, line 3: Why is this change needed?  case change and whitespace change seems unnecessary.

src/test_lib_json/CMakeLists.txt
```diff
@@ -1,39 +1,29 @@
...
+        fuzz.h
+        main.cpp
+        )
 
-
-if(BUILD_SHARED_LIBS)
-    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.12.0)
-        add_compile_definitions( JSON_DLL )
-    else()
-        add_definitions( -DJSON_DLL )
```
hjmjohnson, line 16: Can the compile time defines be removed?  This seems like a possible regression in behavior here.

CMakeLists.txt
```diff
@@ -40,12 +40,6 @@ foreach(pold "") # Currently Empty
     endif()
 endforeach()
 
-# Build the library with C++11 standard support, independent from other including
-# software which may use a different CXX_STANDARD or CMAKE_CXX_STANDARD.
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
hjmjohnson, line 47: Removing this seems like another change of behavior.  Why are these removed here?

example/CMakeLists.txt
Andres6936: Yes sir, I have already made the change along the modified CMakeLists.txt.

src/test_lib_json/CMakeLists.txt
Andres6936: This is usually caused by automatic CLion reformatting.

src/test_lib_json/CMakeLists.txt
Andres6936: No, it does not produce regression, the definitions have already been stated [here](https://github.com/open-source-parsers/jsoncpp/pull/1374/files#diff-8cbbdca03fbe3b64d7024e4899f4abbc6f77db78837901951a5622918504fe79R115).

CMakeLists.txt
Andres6936: No, super discouraged to use the global rules, they produce the overwriting at global level for all the TARGETS defined in this project and those who want to include jsoncpp as a dependency, instead the own definition is used to each TARGET individually.

```cmake
set_target_properties(jsoncpp.framework PROPERTIES CXX_STANDARD 11)
set_target_properties(jsoncpp.framework PROPERTIES CXX_EXTENSIONS OFF)
set_target_properties(jsoncpp.framework PROPERTIES CXX_STANDARD_REQUIRED ON)
```

example/readFromString/CMakeLists.txt
Andres6936: Sir @hjmjohnson , I have already made the preformatting I wanted.

src/test_lib_json/CMakeLists.txt
Andres6936: Any action you would like me to take sir, I have already done some formatting of that section of code, let me know if it is to your approval.

hjmjohnson:
@Andres6936 all these small changes seem to be conceptually related.  Reviewing the changes is difficult because some early changes are updated in later changes.  I would recommend squashing these changes into self-contained patch sets that directly address the desired changes without introducing intermediary temporary steps.  Perhaps squash all of this into 1 commit, or squash the "lower-case" fixups into the commits that introduced the uppercase cmake directives so that the diffs are smaller and easier to understand.

Andres6936:
I will try Sir, although to be frank I don't know how to do it without redoing a PR of everything already done.

.gitignore
```diff
@@ -46,6 +46,7 @@ compile_commands.json
 *.a
 
 # eclipse project files
+.idea/
```
xkszltl, line 49: Recommend to use gitignore.io to general a full and managable intellij config

src/lib_json/CMakeLists.txt
```diff
@@ -105,104 +75,62 @@ list(APPEND REQUIRED_FEATURES
...
-        $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/${JSONCPP_INCLUDE_DIR}>
-        $<BUILD_INTERFACE:${PROJECT_BINARY_DIR}/include/json>
-    )
+if (JSONCPP_BUILD_SHARED_LIBS)
 
-    list(APPEND CMAKE_TARGETS ${STATIC_LIB})
-endif()
+    # CMake link shared library on Windows
+    # Ref: https://stackoverflow.com/a/41618677
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
```
xkszltl, line 107: Better to make it an option rather than set forcefully

Andres6936: In fact it is the only option that exists to obtain a similar behavior in both Linux and Windows without having to touch code or use the export directives in each symbol to be exported, this simulates the default behavior that occurs in Linux to export all symbols by default, which does not happen in Windows.

Although you are right, the code already has the JSON_API directives that solve this problem, I will add the option.

.gitignore
Andres6936: Yes sir, I will

src/lib_json/CMakeLists.txt
xkszltl: I mean, option() with ON as default would allow user overwrite in case there's any compatibility issue with their setup.

Andres6936: In my opinion we should not add an eventual fix to a hypothetical event, for my part it is more convenient to wait for the confirmation of the compatibility problem with the incident report.

Likewise, if you have an article where the compatibility issue with CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS is discussed, I would like to have a document about them to start mitigating the impact of the change.

.gitignore
Andres6936: Done!

PR #1378, opened 2022-01-04
Use default rather than hard-coded 8 for maximum aggregate member alignment

jrtc27:
On CHERI, and thus Arm's Morello prototype, pointers are represented as
hardware capabilities. These capabilities are comprised of not just an
integer address, as is the representation for traditional pointers, but
also bounds, permissions and other metadata, plus a tag bit used as the
validity bit, which provides fine-grained spatial and referential safety
for C and C++ in hardware. This tag bit is not part of the data itself
and is instead kept on the side, flowing with the capability between
registers and the memory subsystem, and any attempt to amplify the
privilege of or corrupt a capability clears this tag (or, in some cases,
traps), rendering them impossible to forge; you can only create
capabilities that are (possibly trivial) subsets of existing ones.

When the capability is stored in memory, this tag bit needs to be
preserved, which is done through the use of tagged memory. Every
capability-sized word gains an additional non-addressable (from the
CPU's perspective; depending on the implementation the tag bits may be
stored in a small block of memory carved out of normal DRAM that the CPU
is blocked from accessing) bit. This means that capabilities can only be
stored to aligned locations; attempting to store them to unaligned
locations will trap with an alignment fault or, if you end up using a
memcpy call, will copy the raw bytes of the capability's representation
but lose the tag, so when it is eventually loaded back as a capability
and dereferenced it will fault.

Since, on 64-bit architectures, our capabilities, used to implement C
language pointers, are 128-bit quantities, this means they need 16-byte
alignment. Currently the various #pragma pack directives, used to work
around (extremely broken and bogus) code that includes jsoncpp in a
context where the maximum alignment has been overridden, hard-code 8 as
the maximum alignment to use, and so do not sufficiently align CHERI /
Morello capabilities on 64-bit architectures. On Windows x64, the
default is also not 8 but 16 (ARM64 is supposedly 8), so this is
slightly dodgy to do there too, but in practice likely not an issue so
long as you don't use any 128-bit types there.

Instead of hard-coding a width, use a directive that resets the packing
back to the default. Unfortunately, whilst GCC and Clang both accept
using #pragma pack(push, 0) as shorthand like for any non-zero value,
MSVC does not, so this needs to be two directives.

### DIFF
--- a/include/json/allocator.h
+++ b/include/json/allocator.h
@@ -9,7 +9,8 @@
 #include <cstring>
 #include <memory>
 
-#pragma pack(push, 8)
+#pragma pack(push)
+#pragma pack()
 
 namespace Json {
 template <typename T> class SecureAllocator {
--- a/include/json/json_features.h
+++ b/include/json/json_features.h
@@ -10,7 +10,8 @@
 #include "forwards.h"
 #endif // if !defined(JSON_IS_AMALGAMATION)
 
-#pragma pack(push, 8)
+#pragma pack(push)
+#pragma pack()
 
 namespace Json {
 
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -23,7 +23,8 @@
 #pragma warning(disable : 4251)
 #endif // if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING)
 
-#pragma pack(push, 8)
+#pragma pack(push)
+#pragma pack()
 
 namespace Json {
 
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -53,7 +53,8 @@
 #pragma warning(disable : 4251 4275)
 #endif // if defined(JSONCPP_DISABLE_DLL_INTERFACE_WARNING)
 
-#pragma pack(push, 8)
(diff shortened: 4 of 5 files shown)


BillyDonahue:
thanks. seems ok on paper. Does the jsoncpp code ever rely on the quantity being exactly 8? with pointer arithmetic or other kludgy tricks?

jrtc27:
It shouldn't do; that'd break on 32-bit architectures, since those would still only use 4 not 8 for the alignment given the 8 here is just a cap, not a value to force. I ran all the tests on Morello and they all pass, too (well, other than the ones that rely on C++ exceptions, something's broken in libunwind/libcxxrt/the toolchain I need to go hunt down...).

BillyDonahue (approved):

jrtc27:
What's this waiting on, please? A review from someone else?

BillyDonahue:
Yeah, I was giving the other maintainers a chance to look at it before merging it.
Then I forgot about it. We can merge it now.

jrtc27:
Great, thanks

PR #1383, opened 2022-01-24
Fixes PreventInSourceBuilds.cmake to work with add_subdirectory

morbo84:
These changes, while still preventing in-source builds for the `jsoncpp` project, allow in-source builds for other projects possibly importing `jsoncpp` via CMake's `add_subdirectory()`. Without these changes, the `AssureOutOfSourceBuilds`  will check for the src and build dirs of the root project, not necessarily of the `jsoncpp` project.


mkruskal-google:
Is there anything blocking this PR?  We're hitting this issue in https://github.com/protocolbuffers/protobuf and had to downgrade to 1.9.4 because of it

PR #1386, opened 2022-01-30
build(meson): use find_program('python3')

Tachi107:
If you really want to be sure to always find python3 when running Meson (and not some other implementation like [Muon](https://muon.build)) it is a bit better to use `find_program('python3')`, as described in https://mesonbuild.com/Reference-manual_functions.html#find_program: _"if the "python3" program is requested and it is not found in the system, Meson will return its current interpreter_

Edit: by better I mean faster and less ambiguous (as this will simply look for python3 in $PATH when Meson is not running through a python interpreter), see https://github.com/mesonbuild/meson/issues/3856#issuecomment-1025015693


Tachi107:
Thanks for the not-so-quick merge! ;)

PR #1397, opened 2022-03-23
add a valueToQuotedString overload 

phlptp:
to take a string length to support things like a string_view more directly.

The valueToQuotedString method is useful for a few other items related to JSON generation.  In many cases we had to use it directly from a std::string_view, which in the present code required conversion to a string, then using the c_str and it computing the length.  So this PR adds an overload that allows direct specification of the string length.  and a few places in the code that could take advantage of it.


src/lib_json/json_writer.cpp
```diff
@@ -490,7 +494,7 @@ void StyledWriter::writeValue(const Value& value) {
         const String& name = *it;
         const Value& childValue = value[name];
         writeCommentBeforeValue(childValue);
-        writeWithIndent(valueToQuotedString(name.c_str()));
+        writeWithIndent(valueToQuotedString(name.c_str(),name.size()));
```
BillyDonahue, line 497:
```suggestion
        writeWithIndent(valueToQuotedString(name.c_str(), name.size()));
```

src/lib_json/json_writer.cpp
```diff
@@ -708,7 +712,7 @@ void StyledStreamWriter::writeValue(const Value& value) {
...
```
BillyDonahue, line 715:
```suggestion
        writeWithIndent(valueToQuotedString(name.c_str(), name.size()));
```

BillyDonahue (changes requested):

PR #1409, opened 2022-06-06
CharReader: Add StructuredError

martinduffy1:
The move from Json::Reader to Json::CharReader omitted the ability to get structured error messages from the parser. Clients which want to extract the line and column information have to parse the returned string. Add a new getStructuredErrors() method which exposes the error as structured data instead of a string.

Based on the work from @kylefromkitware and comments from @BillyDonahue on #1281.

PR #1412, opened 2022-06-15
Expose getLocationLineAndColumn() as a public API

martinduffy1:
Allows better use of the offset data returned by getStructuredErrors()

Based on the work from @KyleFromKitware and comments from @BillyDonahue on #1282

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -608,7 +608,7 @@ bool Reader::decodeDouble(Token& token, Value& decoded) {
       value = -std::numeric_limits<double>::infinity();
     else if (!std::isinf(value))
       return addError(
-        "'" + String(token.start_, token.end_) + "' is not a number.", token);
+          "'" + String(token.start_, token.end_) + "' is not a number.", token);
   }
   decoded = value;
   return true;
@@ -767,24 +767,9 @@ Reader::Char Reader::getNextChar() {
 
 void Reader::getLocationLineAndColumn(Location location, int& line,
                                       int& column) const {
-  Location current = begin_;
-  Location lastLineStart = current;
-  line = 0;
-  while (current < location && current != end_) {
-    Char c = *current++;
-    if (c == '\r') {
-      if (*current == '\n')
-        ++current;
-      lastLineStart = current;
-      ++line;
-    } else if (c == '\n') {
-      lastLineStart = current;
-      ++line;
-    }
-  }
-  // column & line start at 1
-  column = int(location - lastLineStart) + 1;
-  ++line;
+  auto loc = locateInDocument(document_.data(), location - document_.data());
+  line = loc.line;
+  column = loc.column;
 }
 
 String Reader::getLocationLineAndColumn(Location location) const {
@@ -1660,7 +1645,7 @@ bool OurReader::decodeDouble(Token& token, Value& decoded) {
       value = -std::numeric_limits<double>::infinity();
     else if (!std::isinf(value))
       return addError(
-        "'" + String(token.start_, token.end_) + "' is not a number.", token);
+          "'" + String(token.start_, token.end_) + "' is not a number.", token);
   }
   decoded = value;
   return true;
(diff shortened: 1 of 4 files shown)


billhoffman:
@BillyDonahue do you think you can look at this?  It is an important change required by the CMake project to allow for better error reporting in the CMake presets functionality. Thanks.

billhoffman:
@martinduffy1  looks like the branch might need a rebase.

martinduffy1:
https://github.com/open-source-parsers/jsoncpp/pull/1409 is the more important change for this.

src/lib_json/json_reader.cpp
```diff
@@ -1991,6 +1978,32 @@ bool parseFromStream(CharReader::Factory const& fact, IStream& sin, Value* root,
...
+      lastLineStart = current;
+      ++line;
+    } else if (c == '\n') {
+      lastLineStart = current;
+      ++line;
+    }
+  }
+  // column & line start at 1
+  lineAndColumn.column = int(beginDoc + location - lastLineStart) + 1;
+  lineAndColumn.line = ++line;
```
BillyDonahue, line 2002: I think this function can be written more simply than this.
You don't need to introduce variables for `lineAndColumn`, `current`, or `lastLineStart`.

You don't need `endDoc`. It's not handled correctly anyway.
If the `location` indicates a position beyond `endDoc`, the result is silently considered to be a position in an imaginary last line of infinite length. This is not a good failure mode and so don't even try, I think.

```c++
    int line = 1;
    int col = 1;
    const char* last = beginDoc + location;
    for (; beginDoc != last; ++beginDoc) {
      switch (*beginDoc) {
        case '\r':
          if (beginDoc + 1 != last && beginDoc[1] == '\n')
            continue;  // consume CRLF as a single token.
          [[fallthrough]];  // CR without a following LF is same as LF
        case '\n':
          col = 1;
          ++line;
          break;
        default:
          ++col;
          break;
      }
    }
    return {line, col};
```

src/test_lib_json/main.cpp
```diff
@@ -3903,6 +3903,37 @@ JSONTEST_FIXTURE_LOCAL(FuzzTest, fuzzDoesntCrash) {
...
+  }
+};
+
+JSONTEST_FIXTURE_LOCAL(GetLocationLineAndColumnTest, test) {
+  const std::string example = "line 1\nline 2\r\nline 3\rline 4";
+  size_t found_l = -1;
+  size_t found_i = -1;
+  size_t found_n = -1;
+  size_t found_e = -1;
+  size_t found_space = -1;
```
BillyDonahue, line 3922: This test is too clever I think. Trying to reuse these variables and use a fixture function.
It can just hard-code these positions and use a more creative test string.
Just hard-code some inputs and outputs in a table, and run a loop over that table.
It would be less code and easier to maintain.

Improve the name "test".

Needs more coverage.
I think it's missing scenarios where:
    - input has trailing CR,LF,CRLF at end of string or beginning.
    - input with blank lines.
    - input with embedded `\0`.
    - weird stuff like LFCR or CRCRLF.

src/test_lib_json/main.cpp
```diff
@@ -3903,6 +3903,37 @@ JSONTEST_FIXTURE_LOCAL(FuzzTest, fuzzDoesntCrash) {
                              example.size()));
 }
 
+struct GetLocationLineAndColumnTest : JsonTest::TestCase {
+  void testLineAndColumn(const std::string& doc, ptrdiff_t location, int row,
```
BillyDonahue, line 3907: subtly changing size_t to ptrdiff_t here. No reason to do that I think.

src/test_lib_json/main.cpp
```diff
@@ -3903,6 +3903,37 @@ JSONTEST_FIXTURE_LOCAL(FuzzTest, fuzzDoesntCrash) {
...
+    found_l = example.find('l', found_l + 1);
+    found_i = example.find('i', found_i + 1);
+    found_n = example.find('n', found_n + 1);
+    found_e = example.find('e', found_e + 1);
+    found_space = example.find(' ', found_space + 1);
+    testLineAndColumn(example, found_l, row, 1);
+    testLineAndColumn(example, found_i, row, 2);
+    testLineAndColumn(example, found_n, row, 3);
+    testLineAndColumn(example, found_e, row, 4);
+    testLineAndColumn(example, found_space, row, 5);
```
BillyDonahue, line 3933: I don't think the 3rd, 4th, and 5th finds tell you anything the first 2 didn't already tell you.

BillyDonahue (changes requested):

martinduffy1:
Thanks for the feedback @BillyDonahue, I addressed the comments so far.
I'll also mention that #1409 has the more important changes for this and probably makes more sense to progress first, as it allows us to get the offset from the error message that we would use with this API.

martinduffy1:
@BillyDonahue since #1409 is approved now, can you take another look at this?

baylesj:

PR #1420, opened 2022-07-19
Change signature of newCharReader to return uniqu_ptr

AssafPassalCG:
Change signature of newCharReader at CharReaderBuilder to return unique_ptr 
to ensure correct memory management

### DIFF
--- a/doc/jsoncpp.dox
+++ b/doc/jsoncpp.dox
@@ -116,7 +116,7 @@ CharReaders and StreamWriters are not thread-safe, but they are re-usable.
 \code
 Json::CharReaderBuilder rbuilder;
 cfg >> rbuilder.settings_;
-std::unique_ptr<Json::CharReader> const reader(rbuilder.newCharReader());
+std::unique_ptr<Json::CharReader> const reader = rbuilder.newCharReader();
 reader->parse(start, stop, &value1, &errs);
 // ...
 reader->parse(start, stop, &value2, &errs);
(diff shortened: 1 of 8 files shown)


BillyDonahue:
`newCharReader` is a fundamental function of this library. Changing its return type will break all existing user code.

AssafPassalCG:
According to your sample code, the right way to use it is to wrap the
return of the function with unique_ptr,
I understand it might break code,
The reason I made this change is that we had a memory leak because of that
code
and want to prevent it from happening again (in our code, we made the
change),
do you have any suggestions?
maybe add another function to the code?
Thanks
Assaf Passal

BillyDonahue:
Adding another function would be good in the long term.
For the immediate topic, we aren't accepting this PR because it's a breaking change.

A suggestion for a new `std::unique_ptr<CharReader>` function would be better discussed in an Issue thread I think. We may want to make other API changes at the same time when we introduce such a thing. Once we create it, we have to support it, so we would want to make it as comprehensively future-proof as we can at introduction.

For now users who want to capture the raw pointer in a `unique_ptr` can do so by adding some syntax to their `newCharReader` call sites or defining a helper function. It's a little annoying, though.

AssafPassalCG:
I opened https://github.com/open-source-parsers/jsoncpp/issues/1422
I understand the issue with breaking change. I think that you should consider making that change because of the chance that if that will break someone's code, there is a good chance that he/she/they have a memory leak.

PR #1435, opened 2022-09-19
Protect target JsonCpp::JsonCpp against multi-include

### LINKED ISSUE #1356: Build starts to fail with 1.9.5 due to JsonCpp::JsonCpp target not protected against multi-include
**Describe the bug**
Our build suddenly started failing after upgrade to jsoncpp 1.9.5 with the following error:
```
CMake Error at C:/Jenkins/vcpkg/scripts/buildsystems/vcpkg.cmake:578 (_add_library):
   _add_library cannot create imported target "JsonCpp::JsonCpp" because
   another target with the same name already exists.
Call Stack (most recent call first):
  C:/Jenkins/vcpkg/installed/x86-windows/share/jsoncpp/jsoncpp-namespaced-targets.cmake:5 (add_library)
  C:/Jenkins/vcpkg/installed/x86-windows/share/jsoncpp/jsoncppConfig.cmake:31 (include)
  C:/Jenkins/vcpkg/scripts/buildsystems/vcpkg.cmake:788 (_find_package)
```

**To Reproduce**
Have nested CMakeLists which at different places call find_package(jsoncpp)

**Expected behavior**
With 1.9.4 the build succeeds. All previous targets in jsoncpp-targets.cmake are protected against multiple definitions with something equivalent to "if (TARGET <target>) return() endif()". But the new target in jsoncpp-namespaced-targets.cmake is not similarly protected.

**Potential solutions**

Just adding
```
if (TARGET JsonCpp::JsonCpp)
  return()
endif()
```
to jsoncpp-namespaced-targets.cmake seems to prevent the issue. If you want I can open a PR to that effect. #1306 might also fix it, I haven't tested that, yet.

rjoomen:
Fixes #1356, courtesy of @johnwason


Marcus-D-Forte (approved):

Marcus-D-Forte (approved):

rjoomen:
Friendly ping.

JonasVautherin:
Is this project maintained? It's not clear to me.

BillyDonahue:
> Is this project maintained? It's not clear to me.

yes

jsoncpp-namespaced-targets.cmake
```diff
@@ -1,7 +1,7 @@
-if (TARGET jsoncpp_static)
+if (TARGET jsoncpp_static AND NOT TARGET JsonCpp::JsonCpp)
```
BillyDonahue, line 1: ok I had to look this up. `if (TARGET x)` tests for the existence of a definition of a target `x`.

https://cmake.org/cmake/help/latest/command/if.html#target

So if I understand correctly, this is saying that before we call `add_library` to create a `JsonCpp::JsonCpp` target, we are going to check whether one already exists.

I'm a cmake newbie, but make this is better as its own if branch above the others.
Then we are only changing 1 line and we aren' repeating ourselves as much.

```cmake
if (TARGET JsonCpp::JsonCpp)
elseif (TARGET jsoncpp_static)
    ...
elseif (TARGET jsoncpp_lib AND NOT TARGET JsonCpp::JsonCpp)
    ...
endif ()
```

If that doesn't work, that's fine too.

BillyDonahue (approved):

jsoncpp-namespaced-targets.cmake
Levi-Armstrong: @rjoomen Would the requested change work?

ZigRazor (approved):

jsoncpp-namespaced-targets.cmake
rjoomen: I suppose that would work and is slightly cleaner. I modified the PR.

petersteneteg:
This is the right way to do this! Any reason this has not been merged?

JonasVautherin (approved):

baylesj:
Recently we were in a situation where the TravisCI integration was broken and caused no patches to get landed. This unfortunately happened for quite some time and I have now resolved it by migrating the repository to GitHub Actions.

PR #1444, opened 2022-11-16
Proposal: define exceptions when !JSON_USE_EXCEPTION

vslashg:
The current behavior of only defining these classes when exceptions are enabled creates difficulties for large projects with trees of dependencies.

OSS libraries will often reasonably try/catch on these exceptions, which causes those libraries not to build when !JSON_USE_EXCEPTION.  This is easy to miss in those libraries, which don't test for nonstandard configurations of their dependencies, but it makes those libraries difficult to use in projects with large dependency trees that have already configured jsoncpp in this way.


BillyDonahue:
I think I understand the build problem but it sounds like a system that can't compile without these definitions is a system that's expecting JsonCpp to throw them. But if it was built without JSON_USE_EXCEPTION, it won't. So they're anticipating a different API from the one they're linking with.  Would unconditionally providing these definitions (but never actually using them) really help?

baylesj:
> I think I understand the build problem but it sounds like a system that can't compile without these definitions is a system that's expecting JsonCpp to throw them. But if it was built without JSON_USE_EXCEPTION, it won't. So they're anticipating a different API from the one they're linking with. Would unconditionally providing these definitions (but never actually using them) really help?
> 
> I feel like a library in a different configuration like this should be treated like a separate library entirely. If there's a dependency diamond it needs to be resolved more thoroughly.

I agree. Let's close.

PR #1445, opened 2022-11-16
Fix asserts in Value::setComment

vslashg:
The existing asserts seem to not be what was intended; they appear to have been mistranslated in pull/877.

The first assert for `comment.empty()` was previously a check that a provided `const char*` parameter was not null.  The function this replaced accepted empty strings, and the if() statement at the start of this function handles them.

The second assert for `comment[0] == '\0'` was written when `comment` was a `const char*`, and was testing for empty c-string input.  This PR replaces it with `comment.empty()` to match the original intent.


BillyDonahue (approved):
sounds about right. Thanks!

Issue #1447, opened 2022-11-23
CMake: remove ccache babysitting


SpaceIm:
**Is your feature request related to a problem? Please describe.**
Currently jsoncpp unconditionally tries to use ccache if found on system, so build may not be reproducible depending on build machine configuration, or even fails due to local cache corruption in ccache.

**Describe the solution you'd like**
Completely remove ccache micro-management in CMakeLists since it's not a usage requirement of jsoncpp. There is a canonical way to enable ccache usage: set CMAKE_C_COMPILER_LAUNCHER & CMAKE_CXX_COMPILER_LAUNCHER externally, so that users have full control over ccache usage, regardless of what is installed on their system.

PR #1448, opened 2022-11-23
CMake: remove ccache micro management

SpaceIm:
closes https://github.com/open-source-parsers/jsoncpp/issues/1447


juan-lunarg (commented):
I noticed this as well building the Vulkan SDK and would like to see this merged.

dlarsson-adobe:
I would love to see this merged too. Or at least put in some kind of check for whether it's the top level project

Issue #1163, opened 2020-04-28
jsoncpp_readerwriter unit tests failed on Windows


xkszltl:
**Describe the bug**

On windows, we always have 2 tests failed.

```
PS ********\jsoncpp\build> ctest
Test project ********/jsoncpp/build
    Start 1: jsoncpp_readerwriter
1/3 Test #1: jsoncpp_readerwriter ................***Failed    0.95 sec
    Start 2: jsoncpp_readerwriter_json_checker
2/3 Test #2: jsoncpp_readerwriter_json_checker ...***Failed    1.43 sec
    Start 3: jsoncpp_test
3/3 Test #3: jsoncpp_test ........................   Passed    0.01 sec

33% tests passed, 2 tests failed out of 3

Total Test time (real) =   2.41 sec

The following tests FAILED:
          1 - jsoncpp_readerwriter (Failed)
          2 - jsoncpp_readerwriter_json_checker (Failed)
Errors while running CTest
```

On Linux (CentOS 7/Ubuntu 18.04) they works fine.

**Desktop (please complete the following information):**
 - OS: Windows Server 2019
 - Ninja version 1.9.0
 - CMake 3.17.2
 - VS2019
 - Python 3.8.2

xkszltl:
Test log:
<img width="821" alt="image" src="https://user-images.githubusercontent.com/5203025/80526648-a6e7e780-89c5-11ea-9275-ccb3c2f5ad2f.png">

cdunn2001:
You named the versions of everything except **jsoncpp* itself. Why skip that one?

If this were a failure of a specific testcase, we would look into it right away. But the whole suite is broken? To debug that, we would need your help.

Also, we do not support **cmake**. Users do. I suggest trying **meson**. But if you have a fix for **cmake**, please submit a PR. We usually accept those pretty quickly In fact, I just merged one a few minutes ago, so perhaps the tip of `master` will help you. Maybe rebase your change at 58d552e onto that.

I was telling someone last week that it seems like half the Issues here relate to **cmake**. Drives me crazy! Anyway, please re-open if you can point to a more specific failure. Or submit a new PR, based on the current tip of `master.

xkszltl:
@cdunn2001 
jsoncpp version is the latest release tag, 1.9.2 currently.
I don't think this is cmake-related, because there's a post-build test at the end of build, which only runs `jsoncpp_test` and passed.
[...]

PR #1453, opened 2022-12-07
Avoid setting global cmake variables on all systems

mknaranja:
In https://github.com/open-source-parsers/jsoncpp/issues/1163#issuecomment-750770862, the author already warned that the setting of global cmake variables would cause troblems when using jsoncpp as a dependency with CMake.

As the problem in https://github.com/open-source-parsers/jsoncpp/issues/1163 was only related to Windows I first would propose to only do changes for windows and MSVC.

Furthermore, I made the global variables local by removing the CACHE keyword: https://riptutorial.com/cmake/example/11821/variables-and-the-global-variables-cache

I built jsoncpp on Windows with Cmake to see that it still works. There is another problem reported in https://github.com/open-source-parsers/jsoncpp/issues/1451 but that is not related as it was there on Windows before.


mknaranja:
any comments here?

nkh-lab:
For me, more understandable and more universal approach with `if(CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)` described in [issues/1451](https://github.com/open-source-parsers/jsoncpp/issues/1451#issuecomment-1353357360), suggested by @atrelinski

nkh-lab:
So, this is [PR ](https://github.com/open-source-parsers/jsoncpp/pull/1459) based on @atrelinski proposal.

atrelinski:
IMO PR https://github.com/open-source-parsers/jsoncpp/pull/1459 address this problem better and I am for merging PR https://github.com/open-source-parsers/jsoncpp/pull/1459 instead of this one

PR #1455, opened 2022-12-28
Changing 'enum' to 'enum class'

ganeshkamath89:
Fix for Visual Studio warning - Warning C26812: Enum type is unscoped. Prefer enum class over enum.
Useful link: https://stackoverflow.com/questions/63610623/warning-c26812-enum-type-is-unscoped-prefer-enum-class-over-enum

Signed-off-by: ganeshkamath89 <ganeshkamath89@gmail.com>

### DIFF
--- a/include/json/reader.h
+++ b/include/json/reader.h
@@ -156,7 +156,7 @@ public:
   bool good() const;
 
 private:
-  enum TokenType {
+  enum class TokenType {
     tokenEndOfStream = 0,
     tokenObjectBegin,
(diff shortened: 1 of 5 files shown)


BillyDonahue:
i don't think so.


That is not a real warning.
Nothing is wrong here per se.

The PR is an API naming change.

My feeling is that we can't accept it.

ganeshkamath89:
> My feeling is that we can't accept it.

Okay thanks, no problem

PR #1457, opened 2023-01-06
Opportunistically take advantage of C++20 move-out-of stringstream

beevvy:


BillyDonahue (approved):

beevvy:
Pushed a similar change for `json_reader.cpp`.

BTW, I realize it won't have any effect when compiling with `-std=c++11`, which is the default, but it will for projects using the amalgamation, for example.

PR #1458, opened 2023-01-10
feat: adds front and back methods to Value type

jwidauer:
This PR adds the `front` and `back` methods to the `Json::Value` class.
These methods can be used to access the first and last elements of a `Json::Value` container.
For a `Json::Value v` they are equal to `*v.begin()` and `*(--v.end())`.

### DIFF
--- a/include/json/value.h
+++ b/include/json/value.h
@@ -585,20 +585,20 @@ public:
   iterator begin();
   iterator end();
 
-  /// @brief Returns a reference to the first element in the container.
-  /// Calling front on an empty container is undefined behavior.
-  Value const& front() const;
+  /// \brief Returns a reference to the first element in the `Value`.
+  /// Requires that this value holds an array or json object, with at least one element.
+  const Value& front() const;
 
-  /// @brief Returns a reference to the last element in the container.
-  /// Calling back on an empty container is undefined behavior.
-  Value const& back() const;
-
-  /// @brief Returns a reference to the first element in the container.
-  /// Calling front on an empty container is undefined behavior.
+  /// \brief Returns a reference to the first element in the `Value`.
+  /// Requires that this value holds an array or json object, with at least one element.
   Value& front();
 
-  /// @brief Returns a reference to the last element in the container.
-  /// Calling back on an empty container is undefined behavior.
+  /// \brief Returns a reference to the last element in the `Value`.
+  /// Requires that value holds an array or json object, with at least one element.
+  const Value& back() const;
+
+  /// \brief Returns a reference to the last element in the `Value`.
+  /// Requires that this value holds an array or json object, with at least one element.
   Value& back();
 
   // Accessors for the [start, limit) range of bytes within the JSON text from
@@ -940,6 +940,14 @@ public:
 };
 
 inline void swap(Value& a, Value& b) { a.swap(b); }
+
+inline const Value& Value::front() const { return *begin(); }
+
+inline Value& Value::front() { return *begin(); }
+
+inline const Value& Value::back() const { return *(--end()); }
+
+inline Value& Value::back() { return *(--end()); }
 
 } // namespace Json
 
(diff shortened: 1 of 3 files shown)


BillyDonahue:
> a Json::Value container.

They're only valid if the `Value` is holding an array.
If the `Value` is holding an object, or anything scalar, this will fail, right?

jwidauer:
@BillyDonahue Does the previous comment answer your question?

include/json/value.h
```diff
@@ -585,6 +585,22 @@ class JSON_API Value {
   iterator begin();
   iterator end();
 
+  /// @brief Returns a reference to the first element in the container.
```
BillyDonahue, line 588: Surrounding code is using backslashes instead of @ signs. Can you change to that?

include/json/value.h
```diff
@@ -585,6 +585,22 @@ class JSON_API Value {
...
+  /// Calling front on an empty container is undefined behavior.
```
BillyDonahue, line 589: I think the term "container" can be confusing?
Instead of saying what's undefined behavior, maybe state as a precondition.

"Requires that value holds an array or json object, with `size() > 0`."

src/lib_json/json_value.cpp
```diff
@@ -1495,6 +1495,14 @@ Value::iterator Value::end() {
   return iterator();
 }
 
+Value const& Value::front() const { return *begin(); }
```
BillyDonahue, line 1498: Inline these I think.

BillyDonahue (changes requested):

src/lib_json/json_value.cpp
jwidauer: Don't think I can inline these, since they use an incomplete type (the iterators are only forward declared in the header).
I feel like they should be able to be `noexcept` though, but since neither `begin()` nor `end()` are defined `noexcept` I'm not totally sure that's valid, tbh.

include/json/value.h
jwidauer: Sure! Done!

include/json/value.h
jwidauer: I changed it to `Value` instead. Do you think that fits better?
Also changed the explanation to your suggested one.

src/lib_json/json_value.cpp
BillyDonahue: Ah. I didn't realize that about the iterators. They're defined at the bottom of value.h, so the inlining could still happen, but it have to be a the bottom of the file where `swap` is defined. But it wouldn't do us a whole lot of good unless `begin` and `end` were also inlined, so nevermind I guess.

include/json/value.h
```diff
@@ -585,6 +585,22 @@ class JSON_API Value {
...
+  /// Requires that this value holds an array or json object, with at least one element.
+  const Value& front() const;
+
+  /// \brief Returns a reference to the last element in the `Value`.
+  /// Requires that value holds an array or json object, with at least one element.
+  const Value& back() const;
+
+  /// \brief Returns a reference to the first element in the `Value`.
+  /// Requires that this value holds an array or json object, with at least one element.
+  Value& front();
```
BillyDonahue, line 598: Let's have the order here match the order in the .cpp file, which I think is the better ordering:

```c++
    const Value& front() const;
    Value& front();
    const Value& back() const;
    Value& back();
```

BillyDonahue (changes requested):
one more small request. otherwise good to go, I think.

src/lib_json/json_value.cpp
jwidauer: Ah, didn't notice that! I moved them into the header file and underneath swap now, so that, if someone decides to inline `begin` and `end` at some point, there'll be less friction!

BillyDonahue (approved):

BillyDonahue:
Thank you! LGTM

jwidauer:
@BillyDonahue What's the status on this? Is it still something you'd be interested in getting in?

BillyDonahue:
I guess it's safe to say Travis CI isn't going to finish :)

PR #1477, opened 2023-03-28
Update readFromString.cpp print error

gnudles:
Print the error to screen


BillyDonahue (approved):
yup good idea

gnudles:
I did not tested it. Can you try and see if it is formatted well when there is an error?

PR #1486, opened 2023-05-03
Create a jsoncppConfig.cmake file, even if building under meson

wdouglass:
If jsoncpp gets built under meson, no cmake package files get created. This means, if another project wants to use jsoncpp, then jsoncpp must have also been built with cmake.

I've created a shim that allows a `jsoncppConfig.cmake` file to be generated under a meson build. I have found this useful when building under buildroot, which uses meson.

this file only exposes the jsoncpp_lib target, not the jsoncpp_static one. I don't see a way to build the jsoncpp_static target under meson, but I am much less experienced with meson, so I am very open to suggestion here. the attached patch works for me.


eli-schwartz:
> This means, if another project wants to use jsoncpp, then jsoncpp must have also been built with cmake.

Not really true, the other project could use `find_package(PkgConfig)` followed by `pkg_check_modules(...)`.



> this file only exposes the jsoncpp_lib target, not the jsoncpp_static one. I don't see a way to build the jsoncpp_static target under meson, but I am much less experienced with meson, so I am very open to suggestion here. the attached patch works for me.

The static library is built by meson if you use the `-Ddefault_library=static`. The default is "shared". Another permissible value is "both".

Since this hardcodes .so, it will probably not work on macOS or Windows... it also hardcodes the value of `get_option('libdir')`, `get_option('includedir')`...

wdouglass:
That's fair -- i'll do another lap and resubmit

wdouglass:
I have reworked this pr to address your concerns

meson.build
```diff
@@ -63,11 +63,41 @@ import('pkgconfig').generate(
...
+  shared_name = fs.name(jsoncpp_lib.get_shared_lib().full_path())
+  static_name = fs.name(jsoncpp_lib.get_static_lib().full_path())
+endif
+
+if get_option('default_library') == 'shared' or get_option('default_library') == 'both'
+   cmakeconf.set('MESON_SHARED_TARGET', '''
+add_library(jsoncpp_lib IMPORTED SHARED)
+set_target_properties(jsoncpp_lib PROPERTIES
+  IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/''' + get_option('libdir') + '/' + shared_name + '''"
+  INTERFACE_INCLUDE_DIRECTORIES "${PACKAGE_PREFIX_DIR}/''' + get_option('includedir') + '")')
```
eli-schwartz, line 86: I'd suggest using meson's `join_paths('${PACKAGE_PREFIX_DIR}', get_option('XXXdir'), shared_name)` here, since it handles the admittedly unlikely case that libdir or includedir aren't inside the prefix, and also means you can avoid repeatedly entering and then leaving a string literal context.

meson.build
```diff
@@ -63,11 +63,41 @@ import('pkgconfig').generate(
...
+set_target_properties(jsoncpp_static PROPERTIES
+  IMPORTED_LOCATION "${PACKAGE_PREFIX_DIR}/''' + get_option('libdir') + '/' + static_name + '''"
+  INTERFACE_INCLUDE_DIRECTORIES "${PACKAGE_PREFIX_DIR}/''' + get_option('includedir') + '")')
+endif
 
 import('cmake').configure_package_config_file(
   name: 'jsoncpp',
   input: 'jsoncppConfig.cmake.meson.in',
   configuration: cmakeconf)
+install_data('jsoncpp-namespaced-targets.cmake', install_dir : get_option('libdir') + '/cmake/' + jsoncpp_lib.name())
```
eli-schwartz, line 100: I'd use join_paths here too.

meson.build
```diff
@@ -15,7 +15,7 @@ project(
     'cpp_std=c++11',
     'warning_level=1'],
   license : 'Public Domain',
-  meson_version : '>= 0.49.0')
+  meson_version : '>= 0.54.0')
```
eli-schwartz, line 18: The fs module is available since 0.53.0, am I overlooking something else that needs 0.54.0?

wdouglass: I'm not sure -- when i ran the build meson complained that I should be requiring 0.54.0

I'll do another commit (for the join_paths suggestion, thanks!) and try 0.53.0 then

eli-schwartz: In that case it's probably correct -- I haven't checked out the PR and tried it myself -- it should list which feature requires 0.54 though.

wdouglass: looks like the 'name' member function was introduced in 0.54.0

```
WARNING: Project specifies a minimum meson_version '>= 0.53.0' but uses features which were added in newer versions:
 * 0.54.0: {'name'}
```

mizvekov:
Ping, this affects some upstream users, can we merge this?

juan-lunarg:
> Ping, this affects some upstream users, can we merge this?

Currently we have to maintain 2 different code paths since we can't use pkg-config reliably for our Windows builds. We only maintain the pkg-config path to be compliant with package managers that use the meson build instead of the CMake build.
```
find_package(jsoncpp CONFIG)

if (TARGET jsoncpp_static)
    # RPATH causes issues for the Linux SDK tarball due to absolute paths.
    # This can be worked around with relative RPATHS but the tarball doesn't ship jsoncpp.
    # So the easiest solution is to just the static version of jsoncpp.
    target_link_libraries(vkvia PRIVATE jsoncpp_static)
    set_target_properties(vkvia PROPERTIES SKIP_BUILD_RPATH TRUE)
elseif(UNIX)
    # https://github.com/LunarG/VulkanTools/issues/1908
    find_package(PkgConfig REQUIRED)
    pkg_check_modules(jsoncpp REQUIRED IMPORTED_TARGET jsoncpp)
    target_link_libraries(vkvia PRIVATE PkgConfig::jsoncpp)
else()
    message(FATAL_ERROR "Unable to find jsoncpp!")
endif()
```

juan-lunarg:
Update never mind we can use the pkg-config path see:
https://github.com/LunarG/VulkanTools/pull/1938

We still support both paths but now we actively test both paths which was my main issue before.

eli-schwartz:
Thanks, that looks pretty reasonable. :)

PR #1499, opened 2023-07-09
Bump CMake policy version to avoid deprecation warning

roehling:
Starting with CMake 3.27, there will be a warning for compat levels below CMake 3.5.


Hinkiii:
Bruh is this ever gonna get pushed

megapatato:
Anything holding this up? It's almost a year, and depending on CMake policy, this can yield a "deprecation error"

jpr42:
btw CMake 4.0 now has less than 3.5 support fully removed. So great that this got in 👍🏽 
 https://cmake.org/cmake/help/latest/release/4.0.html

PR #1502, opened 2023-07-19
Fix a parser bug where tokens are misidentified as commas.

vslashg:
In the old and new readers, when parsing an object, a comment followed by any non-`}` token is treated as a comma.

The new unit test required changing the runjsontests.py flag regime so that failure tests could be run with default settings.


vslashg:
Sorry, I wasn't expecting my second change to get merged into this PR.  I would still like both to be reviewed, but let me know if you'd like me to split this out somehow.  (The second change relies on test framework changes made to the first.)

baylesj:
LGTM

PR #1503, opened 2023-07-19
Fix out-of-bounds read.

vslashg:
getLocationLIneAndColumn would read past the end of the provided buffer if generating an error message at the end of the stream, if the final character was `\r`.


BillyDonahue (approved):
thanks

Issue #723, opened 2018-01-10
about json5, is there any plan to support


joshuafc:
Hello everyone, I found a library named json5, looks interesting. https://github.com/json5/json5/
the project is write by javascript, support json like 
## examples
```js
{
    foo: 'bar',
    while: true,

    this: 'is a \
multi-line string',

    // this is an inline comment
    here: 'is another', // inline comment

    /* this is a block comment
       that continues on another line */

    hex: 0xDEADbeef,
    half: .5,
    delta: +10,
    to: Infinity,   // and beyond!

    finally: 'a trailing comma',
    oh: [
        "we shouldn't forget",
        'arrays can have',
        'trailing commas too',
    ],
}
```
```js
// This file is written in JSON5 syntax, naturally, but npm needs a regular
// JSON file, so compile via `npm run build`. Be sure to keep both in sync!

{
    name: 'json5',
    version: '0.5.0',
    description: 'JSON for the ES5 era.',
    keywords: ['json', 'es5'],
    author: 'Aseem Kishore <aseem.kishore@gmail.com>',
    contributors: [
        // TODO: Should we remove this section in favor of GitHub's list?
        // https://github.com/json5/json5/contributors
        'Max Nanasy <max.nanasy@gmail.com>',
        'Andrew Eisenberg <andrew@eisenberg.as>',
        'Jordan Tucker <jordanbtucker@gmail.com>',
    ],
    main: 'lib/json5.js',
    bin: 'lib/cli.js',
    files: ["lib/"],
    dependencies: {},
    devDependencies: {
        gulp: "^3.9.1",
        'gulp-jshint': "^2.0.0",
        jshint: "^2.9.1",
        'jshint-stylish': "^2.1.0",
        mocha: "^2.4.5"
    },
    scripts: {
        build: 'node ./lib/cli.js -c package.json5',
        test: 'mocha --ui exports --reporter spec',
            // TODO: Would it be better to define these in a mocha.opts file?
    },
    homepage: 'http://json5.org/',
    license: 'MIT',
    repository: {
        type: 'git',
        url: 'https://github.com/json5/json5',
    },
}
```
About Json5, is there any plan to support handwrite json file like that.

cdunn2001:
No plans, but if someone writes it, we'd surely accept.

PR #1510, opened 2023-09-17
add "allowHexadecimal" reader option

Caldfir:
Adds optional read-support for hexadecimal numbers matching `0x[0-9a-fA-F]{1,16}` (size depends on LargestUInt). New capability is tied to the **allowHexadecimal** reader-option.

Not sure if there is much interest aside from myself. This would contribute to possible future [JSON5](https://spec.json5.org/) support which does seem to have been considered (https://github.com/open-source-parsers/jsoncpp/issues/209 https://github.com/open-source-parsers/jsoncpp/issues/723). Would need to support leading `(+|-)` for this to be compliant with the JSON5 spec.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -875,6 +875,7 @@ public:
   bool failIfExtra_;
   bool rejectDupKeys_;
   bool allowSpecialFloats_;
+  bool allowHexadecimal_;
   bool skipBom_;
   size_t stackLimit_;
 }; // OurFeatures
@@ -914,6 +915,7 @@ private:
     tokenArrayEnd,
     tokenString,
     tokenNumber,
+    tokenHexadecimal,
     tokenTrue,
     tokenFalse,
     tokenNull,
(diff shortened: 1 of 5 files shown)


src/lib_json/json_reader.cpp
```diff
@@ -1639,6 +1671,41 @@ bool OurReader::decodeNumber(Token& token, Value& decoded) {
...
+  for (; current < token.end_; ++current) {
+    Char c = *current;
+    if (c >= 'a')
+      c -= 'a' - 10;
+    else if (c >= 'A')
+      c -= 'A' - 10;
+    else if (c >= '0')
+      c -= '0';
+    else
+      return addError("Contains non-hexadecimal digits.", token, current);
```
BillyDonahue, line 1702: seems that this wouldn't catch nonhex characters like 'z'.

BillyDonahue (changes requested):

src/lib_json/json_reader.cpp
Caldfir: Hmm... I think I had assumed that token.start/token.end were being set correctly by `readHexadecimal()`. It would make the method more robust and flexible to check both bounds though. Will fix.

baylesj:

Caldfir:
@baylesj The commit from Feb 2024 ([7a74a99](https://github.com/open-source-parsers/jsoncpp/pull/1510/commits/7a74a996f3892f04bb33c052e28b9237fdabf45a)) addressed the comments from @BillyDonahue 

Should the issue be reopened?

Issue #1512, opened 2023-09-22
Can't Cross Compiling jsoncpp 00.11.z for aarch64


Derrekito:
**Can't seem to get cross-compiling working**
I'm trying to cross compile from x86_64 host for an ARM64 target.

**To Reproduce**
1.cmake -DCMAKE_INSTALL_PREFIX=/usr/aarch64-linux-gnu \
      -DCMAKE_SYSTEM_NAME=Linux \
      -DCMAKE_SYSTEM_PROCESSOR=aarch64 \
      -DCMAKE_C_COMPILER=aarch64-linux-gnu-gcc \
      -DCMAKE_CXX_COMPILER=aarch64-linux-gnu-g++ \
      ..\
      && make

2. file lib/libjsoncpp.so.00.11.0
lib/libjsoncpp.so.00.11.0: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=9b7e6633cefe354d240f3524a8dda7a25abfd683, not stripped

**Expected behavior**
produce an aarch64 object.

**Desktop (please complete the following information):**
 - OS: ubuntu 20.04

BillyDonahue:
Does this cmake invocation usually work? Can you cross-compile other projects successfully with that?

Derrekito:
I usually avoid cmake, so I cannot answer your question. But I've tried
passing some of these cross compilation flags directly and also by creating
and passing a tool chain file.

https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html

Obviously whatever I'm doing isn't working.

How do I cross compile this project from x86_64 architecture for an ARM64
target?

Derrekito:
I appreciate your response. However, I usually avoid cmake, so I cannot adequately answer your question. But I've tried
passing some of these cross-compilation flags directly and also by creating
and passing a toolchain file.

https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html

Here's my aarch-toochain.cmake file

```
# aarch64-toolchain.cmake

# Set the target system name
set(CMAKE_SYSTEM_NAME Linux)
set(CMAKE_SYSTEM_PROCESSOR aarch64)

# Set the cross compilers
set(bin_path /usr/bin)
set(CMAKE_C_COMPILER ${bin_path}/aarch64-linux-gnu-gcc)
[...]

PR #1513, opened 2023-09-25
Fixes cross compiler issues (#1512)

Derrekito:
## Changes Made
Modification in CMakeLists: The script now checks if CMAKE_CXX_COMPILER is explicitly set before proceeding. This change ensures the project's build system is more robust and adaptable to various build environments, particularly in cross-compilation contexts.

## Observed Side Effect
Test Suite Compatibility: Following this modification, the test suite fails if the object files are not compiled for the host architecture.

## Proposed Mitigations
Compiling for Both Host and Target Architectures: One approach to mitigate this side effect is to ensure that objects are compiled for both the host and target architectures. This could resolve the test failures and maintain cross-compilation support.

## Alternative Approach
Suppressing Tests When Cross-Compiling: As an alternative, I suggest suppressing the execution of tests during cross-compilation. This approach might provide a temporary workaround but requires further discussion to assess its impact on the project's quality assurance processes.

## Seeking Feedback
These changes are crucial for enhancing jsoncpp's support for cross-compilation environments. However, I want the community's thoughts regarding the observed side effects and the proposed mitigation strategies. Would suppressing tests in cross-compilation scenarios be viable, or should we explore the compilation for both architectures in more depth?


Derrekito:
Another approach to reconciling tests and cross-compiled objects is to simply suppress tests when cross-compiling. Thoughts?

PR #1517, opened 2023-11-13
Update CMake minimum to 3.16.3

juan-lunarg:
Fixes warnings from using `find_package(jsoncpp CONFIG)` and simplifies the code

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,31 @@
+name: ci
+
+on:
+  push:
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  # Generally speaking the idiom for open source projects is to support
+  # the version of CMake supported by the oldest LTS release.
+  #
+  # Example: Ubuntu-20.04 and Ubuntu-22.04
+  # Ubuntu-20.04's package manager comes with 3.16.3 by default.
+  # This makes boostrapping the build easier for open source development.
+  #
+  # This idiom really only applies to Linux development. Windows/Mac/Android developers
+  # tend to use newer CMake versions due to Visual Studio / Xcode / Android NDK requirements.
+  #
+  # EX: Visual Studio 2022 requires CMake 3.21
+  test-cmake-minimum:
+    name: Test CMake minimum
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.16.3
+      - run: cmake -S . -B out/ -G Ninja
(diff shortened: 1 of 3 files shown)


juan-lunarg:
> Fixes warnings from using find_package(jsoncpp CONFIG) and simplifies the code

To be clear this is the warning I'm talking about:
```
CMake Deprecation Warning at external/Debug/64/jsoncpp/build/install/lib/cmake/jsoncpp/jsoncppConfig.cmake:2 (cmake_policy):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
Call Stack (most recent call first):
  via/CMakeLists.txt:69 (find_package)
```

I've tested the changes I've made in this PR against `LunarG/VulkanTools` and verified it works.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+  #
+  # EX: Visual Studio 2022 requires CMake 3.21
+  test-cmake-minimum:
+    name: Test CMake minimum
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: lukka/get-cmake@latest
+        with:
+          cmakeVersion: 3.16.3
```
juan-lunarg, line 28: I have found this explicit version testing to be useful in various KhronosGroup repos to ensure our build instructions are actually accurate for linux developers. Let me know what you think.

juan-lunarg: If you are wondering where 3.16.3 comes from:
https://packages.ubuntu.com/focal/cmake

.github/workflows/ci.yml
juan-lunarg, line 1: I tested this action in my fork FWIW:
https://github.com/juan-lunarg/jsoncpp/actions

.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+jobs:
+  # Generally speaking the idiom for open source projects is to support
+  # the version of CMake supported by the oldest LTS release.
+  #
+  # Example: Ubuntu-20.04 and Ubuntu-22.04
+  # Ubuntu-20.04's package manager comes with 3.16.3 by default.
+  # This makes boostrapping the build easier for open source development.
+  #
+  # This idiom really only applies to Linux development. Windows/Mac/Android developers
+  # tend to use newer CMake versions due to Visual Studio / Xcode / Android NDK requirements.
```
juan-lunarg, line 18: Note: This idiom also applies to other packages like gcc/clang/clang-format/etc.

PR #1519, opened 2023-11-21
json_value.cpp bug in the edges of uint/int

YaalLek:
Fixing bug of sending a number that is a bit bigger than max<uint64_t> it returns 0: https://stackoverflow.com/questions/77261400/jsoncpp-do-not-protect-from-uint64-overflow-and-have-weird-behavior/77261716#77261716


baylesj:
This looks like a good patch but needs to have clang format run on it before submission.

YaalLek:
> This looks like a good patch but needs to have clang format run on it before submission.

@baylesj 
Ran clang tidy and pushed changes

PR #1552, opened 2024-09-09
Create github-actions-demo.yml

baylesj:


BillyDonahue:
this looks fun

baylesj:
Since Travis is no longer behaving and Actions seems to be the path forward, I'm going to spend some time trying to get the repo at least to the point where development can occur at all. I didn't realize how bad of a state we had gotten into.

baylesj:
https://github.com/open-source-parsers/jsoncpp/actions/workflows/c-cpp.yml

Going to try to get everything enabled in the actions tab and then we can delete the travis integration entirely.

PR #1567, opened 2024-09-12
Fix flag -DJSONCPP_USE_SECURE_MEMORY:BOOL=TRUE

tfc:
The Cmake description uses `JSONCPP_USE_SECURE_MEMORY` and the source uses `JSONCPP_USING_SECURE_MEMORY`.

Is it known that the tests don't compile with this setting being enabled?

### DIFF
--- a/include/json/config.h
+++ b/include/json/config.h
@@ -127,7 +127,7 @@ using LargestUInt = UInt64;
 
 template <typename T>
 using Allocator =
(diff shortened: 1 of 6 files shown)


baylesj:
The tests compile in all of our currently supported bot configurations + locally on my macbook. 

Can you tell me more about your setup? 

Thanks for catching this!

baylesj:
Thanks for catching this. I'll look at adding a new bot configuration that would catch this.

PR #1573, opened 2024-10-06
performance optimization

bjqiwei:
reduce access to all bytes in the string


bjqiwei:
reduce access to all bytes in the string

bjqiwei:
clang format

baylesj (changes requested):
I would like to see some unit tests added for this patch. This is best practice, and I'm still not entirely sure what the use case for this optimization is.

Can you give a more detailed description of the patch and add tests showing the intended use?

Issue #1352, opened 2021-11-05
Add support for std::string_view to Json::Value::operator[]


sir-exasol:
**Is your feature request related to a problem? Please describe.**
We use `std::string_view` for names of object entries. However, when accessing values with `Json::Value::operator[]`, we always have to convert the name a `std::string` first. This incurs an unnecessary allocation and adds boilerplate code.

**Describe the solution you'd like**
I would like `Json::Value::operator[]` to support `std::string_view`.

**Describe alternatives you've considered**
As mentioned an alternative is to convert the name to `std::string` first, but this is less readable and slower. Converting to `const char*` is in general not possible, because the contents of `std::string_view` might not be null-terminated.

**Additional context**
This feature requires C++17. jsoncpp seems to build fine with C++17 so this shouldn't be a problem as long as the feature is deactivated when compiling with a lower standard. 
https://github.com/open-source-parsers/jsoncpp/pull/999 already implemented this feature, but was closed. It is not clear to me why the PR was closed. If there is additional work necessary, I am happy to help.

BillyDonahue:
The PR was incorrect code.
I don't think it would compile. A key variable was misspelled.

It could create an ambiguity among the overloads of the operator[], which would break some users.

the detection of C++17 was incorrect for the case where jsoncpp is built without c++17 but the client code is built with it. The client would have missing symbol errors.

There are workarounds for these problems but it is not simple.

aberaud:
first step here: https://github.com/open-source-parsers/jsoncpp/pull/1397

PR #1584, opened 2024-12-08
feat: support std::string_view in Value API

evalon32:
This adds direct support for `std::string_view` when available (C++17 and above). The current API can be used with `std::string_view` via the low-level two-pointer methods, but is not ergonomic. E.g., compare:

```
Json::Value node;
std::string foo, bar, baz;
std::string_view foo_sv, bar_sv, baz_sv;

// Efficient & readable:
node[foo][bar][baz];

// Less efficient, less readable:
node[std::string(foo_sv)][std::string(bar_sv)][std::string(baz_sv)];

// Efficient, but a lot less readable:
*node.demand(foo_sv.data(), foo_sv.data() + foo_sv.size())
    ->demand(bar_sv.data(), bar_sv.data() + bar_sv.size())
    ->demand(baz_sv.data(), baz_sv.data() + baz_sv.size());

// After this change, efficient & readable:
node[foo_sv][bar_sv][baz_sv];
```

## Change details (only with C++17 or above)

*   The constructor can take a `std::string_view` parameter. The existing
    overloads taking `const std::string&` and `const char*` are still necessary
    to support assignment from those types.
*   `operator[]`, `get()`, `isMember()` and `removeMember()` take a
    `std::string_view` parameter. This supersedes the overloads taking
    `const std::string&` and `const char*`. The overloads taking a pair of
    pointers (begin, end) are preserved for source compatibility.
*   `getString()` has an overload with a `std::string_view` output parameter.
    The one with a pair of pointers is preserved for source compatibility.

### DIFF
--- a/src/lib_json/json_value.cpp
+++ b/src/lib_json/json_value.cpp
@@ -17,6 +17,10 @@
 #include <sstream>
 #include <utility>
 
+#ifdef JSONCPP_HAS_STRING_VIEW
+#include <string_view>
+#endif
+
 // Provide implementation equivalent of std::snprintf for older _MSC compilers
 #if defined(_MSC_VER) && _MSC_VER < 1900
 #include <stdarg.h>
@@ -420,6 +424,14 @@ Value::Value(const String& value) {
       value.data(), static_cast<unsigned>(value.length()));
 }
 
+#ifdef JSONCPP_HAS_STRING_VIEW
+Value::Value(std::string_view value) {
+  initBasic(stringValue, true);
+  value_.string_ = duplicateAndPrefixStringValue(
+      value.data(), static_cast<unsigned>(value.length()));
+}
+#endif
+
 Value::Value(const StaticString& value) {
   initBasic(stringValue);
   value_.string_ = const_cast<char*>(value.c_str());
@@ -626,6 +638,21 @@ bool Value::getString(char const** begin, char const** end) const {
   *end = *begin + length;
   return true;
 }
+
+#ifdef JSONCPP_HAS_STRING_VIEW
+bool Value::getString(std::string_view* str) const {
+  if (type() != stringValue)
+    return false;
+  if (value_.string_ == nullptr)
+    return false;
+  const char* begin;
+  unsigned length;
+  decodePrefixedString(this->isAllocated(), this->value_.string_, &length,
+                       &begin);
+  *str = std::string_view(begin, length);
+  return true;
+}
+#endif
 
 String Value::asString() const {
   switch (type()) {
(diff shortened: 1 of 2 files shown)


src/lib_json/json_value.cpp
```diff
@@ -1233,12 +1298,18 @@ bool Value::isMember(char const* begin, char const* end) const {
   Value const* value = find(begin, end);
   return nullptr != value;
 }
+#ifdef JSONCPP_HAS_STRING_VIEW
+bool Value::isMember(std::string_view key) const {
+  return isMember(key.data(), key.data() + key.length());
```
baylesj, line 1303: I think this works since it's behind a define and conditionally enabled. Ideally we would be able to have the string_view isMember contain the actual logic and use modern C++ in the `find` implementation.

baylesj (approved):

bcsgh:
FWIW: I suspect this resolved https://github.com/open-source-parsers/jsoncpp/issues/1352 ??

beevvy:
Hi @evalon32, @baylesj,

I just noticed this commit in the master branch, and I think there's a possible problem. jsoncpp is built with -std=c++11 by default, which means that when building json_value.cpp, JSONCPP_HAS_STRING_VIEW is never defined. That means that the newly added member functions won't be part of the resulting library. (I haven't verified it.)

The member functions that are compiled into the library cannot be conditionally enabled. Conditional code should be header-only, for example as inline member functions implemented in terms of other, always enabled member functions.

evalon32:
That's a good point -- if the library and its user are compiled with different C++ versions, we'll get linker errors. I'll try to put together a follow-up PR to fix that.

bcsgh:
Outside of plugins, are there actually any cases that really *need* link compatibility for libraries where the consumer has access to the source? Sure there are legacy reason to continue to support it, but are there any good technical reason for the word at large to not deprecate the cases where people are depending on it? 

(There are several large classes of issues that _just don't exist_ in a build-everything-from-source-and-static-link world.)

beevvy:
Most Linux distributions ship with a single libjsoncpp.so.26 that's used by ~all packages depending on it. For it to be usable as shared library, the ABI must be predictable. Of course, for Linux distributions' case, you could just hard depend on C++17, but that's another topic...

bcsgh:
@beevvy the only implication of your observation is a need to deal with that reality (and that **is** a very valid point).

My _gripe_ is that anyone choose to use that .so (or basically any .os expect were the point is to intentionally get different behavior from different .so's) rather than build from source and statically link. My _claim_ is that the cases where anything else is necessarily (or even desirable) are somewhere between few and vanishing.

And FWIW, every package I've worked on that depends on JsonCpp does _not_ depend on `libjsoncpp.so` and, if I have my way, none ever will.

Issue #1591, opened 2024-12-29
The macro name is the same as the std member


XY-ShadowK:
**Describe the bug**
```c++
#if !defined(isfinite)
#include <float.h>
#define isfinite _finite
#endif
```
The macro name `isfinite` is the same as `std::isfinite`, which can cause compiler to incorrectly replace `std::isfinite` to `std::_finite`.

**To Reproduce**
Just use `std::isfinite` after the inclusion of `jsoncpp.cpp`.

**Expected behavior**

**Desktop (please complete the following information):**
 - OS: Windows 10 22H2 x64

**Additional context**
Please use `std::isfinite` to check values, or use `_isfinite` directly or use other macro name to circumvent std member names.

BillyDonahue:
I agree that's messed up to use a colliding name. We could just use some other name.

> Just use std::isfinite after the inclusion of jsoncpp.cpp.

But why would you be including a .cpp file?
The macros are safe in the json_writer.cpp file if used as intended.

XY-ShadowK:
> But why would you be including a .cpp file? The macros are safe in the json_writer.cpp file if used as intended.

I followed the guidelines of [Amalgamated- (Possibly-outdated)](https://github.com/open-source-parsers/jsoncpp/wiki/Amalgamated-(Possibly-outdated)) to import `jsoncpp`.
If I want to compile a cross-platform project, can I import it this way and build with my own project? Or do I need to follow the instructions of [Build-with-VS-2019-on-Windows](https://github.com/open-source-parsers/jsoncpp/wiki/Build-with-VS-2019-on-Windows) and [Building](https://github.com/open-source-parsers/jsoncpp/wiki/Building) to build and import separately?

BillyDonahue:
I think you misunderstood the guidelines.
The amalgamation script produces a .h and a .cpp file.

You add these to your project.
This means you add the single .cpp file to your Makefile or whatever you're using.
It doesn't mean that you #include it into your own files. It's still a separate compilation.

XY-ShadowK:
As below?
So can it be built cross-paltform with my project?
```c++
cl src/mySrc.cpp jsoncpp-1.9.5/dist/jsoncpp.cpp /I jsoncpp-1.9.5/dist
g++ src/mySrc.cpp jsoncpp-1.9.5/dist/jsoncpp.cpp -I jsoncpp-1.9.5/dist
```

BillyDonahue:
[...]

PR #1593, opened 2025-01-09
Drop pre-C++11 alternatives

BillyDonahue:
We already assume C++11 elsewhere (using `unique_ptr`, etc) so all pre-11 `#ifdef` branches are dead code at this point. Fixes issue #1591 because we can just use `std::isfinite` and `std::isnan`.


baylesj:
This looks great! Thank you for the patch.

baylesj (approved):

PR #1597, opened 2025-01-26
Set up for Bazel module builds.

bcsgh:
Notes:
---

1) This likely doesn't require a release note as it _shouldn't_ have any effect on any released product.
1) The `MODULE.bazel` is forked from the "[bazel-central-registry](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/jsoncpp/1.9.6/MODULE.bazel)" and this allows the BCR to quit having to uses patches.
1) With these changes JsonCpp can be built by Bazel without any changes. Without this, Bazel is unable to locate a correct root to build from and fails.
1) This effectively supersede https://github.com/open-source-parsers/jsoncpp/pull/1534 ;; Addressing comments from there:
    1) The version in `MODULE.bazel` should track the version used everywhere else. This just becomes yet-another-copy of that.
    1) `WORKSPACE` is dead (though one _would_ be needed ... if someone had to JsonCpp, as a root module, with Bazel, and without using modules, which should be nobody ever again).
    1) I'm not including a `.bazelrc` as things should "just work" (unless your `~/.bazelrc` does something annoying).

---

It would be _nice_ if making updates to [BCR](https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/jsoncpp) became part of the normal release process. It looks like that could be rather easy  [with the available tool support](https://github.com/bazelbuild/bazel-central-registry/blob/main/docs/README.md) maybe even automated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ compile_commands.json
 
 # Bazel output paths
 /bazel-*
+/MODULE.bazel.lock
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ compile_commands.json
 
 # temps
 /version
+
+# Bazel output paths
+/bazel-*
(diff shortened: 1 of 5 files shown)


bcsgh:
What else is needed before this is ready to merge?

baylesj:
With this patch:

```

baylesj:
It would be nice to add the tests as a BUILD.bazel target, but that's outside the scope of this patch and could be a followup.

PR #1600, opened 2025-03-16
Make the build configuration under Bazel more correct.

bcsgh:
- Expose `JSON_USE_EXCEPTION` and `JSON_HAS_INT64` as Bazel config flags with defaults that match the existing Bazel build.
- Switch `//:jsoncpp` from using `copts` to `defines` for `JSON_USE_EXCEPTION` and `JSON_HAS_INT64` so that rules that depend on it get the same config.
- Make `src/test_lib_json/fuzz.cpp` respect `JSON_USE_EXCEPTION`.


bcsgh:
Poke?

bcsgh:
Any chance of getting a review of this?

PR #1602, opened 2025-03-18
Add a BUILD.bazel file for //example.

bcsgh:
This is more OCD than anything, but there is a `CMakeList` so might as well do this as well. 

#### Testing

All 5 rules work under `bazel run` (and also do so under all configs enabled via https://github.com/open-source-parsers/jsoncpp/pull/1600 ).


baylesj:
Thank you.

PR #1609, opened 2025-05-27
Some tweaks for reading/writing

helly25:
Some tweaks:
* Make `json.h` an IWYU import header.
* Change `Reader::parse` to take its `document` parameter as `std::string_view`.
* Add `static void StreamWriterBuilder::updateDefaults(const Json::Value& settings);`
  * Allows to set the global configuration.


include/json/reader.h
```diff
@@ -74,7 +74,7 @@ class JSON_API Reader {
    * \return \c true if the document was successfully parsed, \c false if an
    * error occurred.
    */
-  bool parse(const std::string& document, Value& root,
+  bool parse(std::string_view document, Value& root,
```
baylesj: This causes a compile time error in the build since we don't support string_view currently. I believe JsonCpp is on C++11? And `std::string_view` was added in C++17.

We have been talking about a major version change so it's quite possible bumping the standard could be part of that version change. But right now I don't think we can land this specific part of the change unfortunately.

PR #1622, opened 2025-09-09
[docs] Consuming JSONCpp via Conan package manager

uilianries:
Hello!

In 2017, this project had a large instruction on how to use Conan to install JSONCpp, but now it's no longer listed. 

This PR adds a note on how to use Conan again, but much smaller and simpler.

Plus, I fixed the badge that was pointing to Bintray, which is no longer available. 

Regards!


baylesj:
This is great, thank you.

PR #1624, opened 2025-11-06
Update vcpkg install command for jsoncpp

albaropereyra22:
Added --classic option, now required.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,6 +1,28 @@
 licenses(["unencumbered"])  # Public Domain or MIT
 
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+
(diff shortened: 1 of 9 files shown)


baylesj:
Can you explain why `--classic` is now necessary? Does JsonCpp have to be modified to work with manifests / the "new" mode?

albaropereyra22:
They changed the formatting, I’m not an expert, I just looked through help realized there was a classic option tried it and it worked like it’s suppose to in the guide.  Instead of rewriting the guide like it needs to be, I just added the —classic option in the interim.  Show something that works for now until an expert writes the updated guide.

I hope this answers your question.

baylesj:
Attempting to address this issue here:

https://github.com/open-source-parsers/jsoncpp/pull/1633

PR #1625, opened 2025-11-07
Fix "include what you use" issue

victorvianna:
Fix IWYU for istreambuf_iterator.


victorvianna:
Hm, it looks like this should have been fixed by https://github.com/open-source-parsers/jsoncpp/pull/1613/files, what's going on?

victorvianna:
@baylesj PTAL

PR #1689, opened 2026-06-13
fix: avoid quadratic re-scan of comments after a value

hjanuschka:
`OurReader::readComment()` decides whether a comment attaches to the previous value (`commentAfterOnSameLine`) by scanning from the end of that value to the comment via `containsNewLine(lastValueEnd_, commentBegin)`. `lastValueEnd_` only advances when a new value is read, so a long run of comments after a value (error recovery, or a value followed by many comments) re-scans the same growing prefix for every comment -- O(n^2).

A comment can only be on the same line as the last value when no newline separates them, and that gap only grows as more comments are consumed, so it never needs re-examining after the first comment. Setting `lastValueHasAComment_` after the first comment makes the rest skip the scan. Output is unchanged.

Found via a `jsoncpp_fuzzer` timeout (Chromium issue 521541633): a 400KB input scanned 2.24 GB across 8384 `containsNewLine` calls and took ~18s. After the fix: 122 KB / 1 call, ~56ms.

Adds `CharReaderTest/parseCommentsAfterValueNotQuadratic` (a value followed by many trailing comments, bounded parse time): fails multi-second without the fix, passes in ms with it. Existing tests unaffected.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1296,9 +1296,13 @@ bool OurReader::readComment() {
       if (lastValueEnd_ && !containsNewLine(lastValueEnd_, commentBegin)) {
         if (isCppStyleComment || !cStyleWithEmbeddedNewline) {
           placement = commentAfterOnSameLine;
-          lastValueHasAComment_ = true;
         }
       }
+      // The gap between the last value and this comment only grows as more
+      // comments are consumed, so a later comment can never be on the same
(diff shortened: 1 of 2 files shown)


baylesj:
The production changes are good, but I'm reluctant to land a non-deterministic timer based test. Updating the test format to be based on execution count instead of a potentially flaky clock timer, and then landing this change.

Thank you for your CL!

hjanuschka:
@baylesj wow, didnt expect to land this so fast!! really appreciated!

PR #1698, opened 2026-06-17
validate utf-16 surrogate halves in decodeUnicodeCodePoint

SABITHSAHEB:
1. after a high surrogate (D800-DBFF) the next \u escape is consumed and combined without checking it is a low surrogate, so "\uD801\u0041" or "\uD801\uD801" parse to a wrong astral code point and the second escape's real value is lost.
2. a low surrogate that appears on its own (e.g. "\uDC00") is passed straight through and gets written out as the invalid UTF-8 bytes ED B0 80.
Validated the low-surrogate range when completing a pair and rejected an unpaired low surrogate, in both Reader and OurReader. Added reader tests for the two new rejection paths; existing valid-pair cases are unaffected.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -678,6 +678,10 @@ bool Reader::decodeUnicodeCodePoint(Token& token, Location& current,
     if (*(current++) == '\\' && *(current++) == 'u') {
       unsigned int surrogatePair;
       if (decodeUnicodeEscapeSequence(token, current, end, surrogatePair)) {
+        if (surrogatePair < 0xDC00 || surrogatePair > 0xDFFF)
+          return addError("expecting a low surrogate (DC00-DFFF) to complete "
+                          "the unicode surrogate pair",
+                          token, current);
         unicode = 0x10000 + ((unicode & 0x3FF) << 10) + (surrogatePair & 0x3FF);
       } else
         return false;
@@ -685,6 +689,10 @@ bool Reader::decodeUnicodeCodePoint(Token& token, Location& current,
       return addError("expecting another \\u token to begin the second half of "
                       "a unicode surrogate pair",
                       token, current);
+  } else if (unicode >= 0xDC00 && unicode <= 0xDFFF) {
+    return addError("unexpected low surrogate (DC00-DFFF); a high surrogate "
+                    "(D800-DBFF) must come first",
+                    token, current);
   }
   return true;
 }
@@ -1759,6 +1767,10 @@ bool OurReader::decodeUnicodeCodePoint(Token& token, Location& current,
     if (*(current++) == '\\' && *(current++) == 'u') {
       unsigned int surrogatePair;
       if (decodeUnicodeEscapeSequence(token, current, end, surrogatePair)) {
+        if (surrogatePair < 0xDC00 || surrogatePair > 0xDFFF)
+          return addError("expecting a low surrogate (DC00-DFFF) to complete "
+                          "the unicode surrogate pair",
+                          token, current);
         unicode = 0x10000 + ((unicode & 0x3FF) << 10) + (surrogatePair & 0x3FF);
       } else
         return false;
@@ -1766,6 +1778,10 @@ bool OurReader::decodeUnicodeCodePoint(Token& token, Location& current,
       return addError("expecting another \\u token to begin the second half of "
                       "a unicode surrogate pair",
                       token, current);
+  } else if (unicode >= 0xDC00 && unicode <= 0xDFFF) {
+    return addError("unexpected low surrogate (DC00-DFFF); a high surrogate "
+                    "(D800-DBFF) must come first",
+                    token, current);
   }
   return true;
 }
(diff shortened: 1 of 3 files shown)


SABITHSAHEB:
any update?

src/test_lib_json/main.cpp
```diff
@@ -3322,6 +3322,24 @@ JSONTEST_FIXTURE_LOCAL(CharReaderTest, parseString) {
                             "second half of a unicode surrogate pair\n"
                             "See Line 1, Column 12 for detail.\n");
   }
+  {
+    char const doc[] = R"([ "\uD801\u0041" ])";
```
baylesj, line 3337: No tests for the legacy reader change?

src/lib_json/json_reader.cpp
baylesj, line 1750: Both decodeString loops copy any raw byte ≥ `0x20` through unvalidated, so the raw WTF-8 bytes `ED B0 80` (`U+DC00`) still parse into a Value containing exactly the invalid UTF-8 this PR aims to prevent, while `"\udc00"` is now a hard error. The "parser accepted it, therefore strings are valid UTF-8" invariant the PR implies is not actually established — identical content is treated differently depending on byte-level spelling.

baylesj (changes requested):
Thanks for the fix — the validation logic itself looks correct. One design concern before merging, though: this rejection is unconditional in both readers, with no way to opt out.

Unlike #1663 (unescaped control characters, which are grammar-invalid per RFC 8259), lone/mismatched surrogate escapes are syntactically valid JSON under both the RFC 8259 ABNF and ECMA-404 — the spec only calls the resulting value "unpredictable". 

Real-world producers emit them legally: ES2019+ JSON.stringify('\uD83D') yields "\ud83d", and Python's json.dumps does the same. After this change, documents that parsed on every prior release hard-fail in every configuration — including CharReaderBuilder::ecma404Mode, which would now reject text that ECMA-404 defines as conforming, and the legacy Json::Reader, whose Features struct has no knob at all.

Comparable strictness decisions (failIfExtra, rejectDupKeys, allowSpecialFloats) are all routed through CharReaderBuilder settings. Could this be gated behind a setting (e.g. rejectInvalidSurrogates, arguably default-on) so downstream users with lenient-ingestion pipelines have an escape hatch? Alternatively, WHATWG-style U+FFFD replacement in the lenient path would avoid the hard break while still fixing the invalid-UTF-8 output.

src/test_lib_json/main.cpp
SABITHSAHEB: Dropped the legacy Reader change entirely, so there's no behavior change left there to test. It has no Features knob to gate this on, and keeping it lenient avoids breaking input that parsed before. The validation now lives only in OurReader behind the `rejectInvalidSurrogates` setting.

src/lib_json/json_reader.cpp
SABITHSAHEB: Fair point, and the raw-byte path is untouched: `ED B0 80` still passes through, so this doesn't establish a strings-are-valid-UTF-8 invariant and I'm not claiming one. The scope here is just the `\u` escape decode turning into a wrong or invalid code point. Validating raw string bytes is a separate, larger change. Gating behind the setting keeps this from implying more than it does.

SABITHSAHEB:
Makes sense on the spec point: these are well-formed per RFC 8259 and ECMA-404, so an unconditional hard-fail in every config was too aggressive.

Routed it through a `rejectInvalidSurrogates` CharReaderBuilder setting, same as failIfExtra/rejectDupKeys/allowSpecialFloats. Default on (your "arguably default-on"), on in strictMode, off in ecma404Mode so conforming lone surrogates still parse there, and lenient-ingestion pipelines can flip it off. I also reverted the legacy Json::Reader change: it has no Features knob to gate this on, and leaving it as-is keeps input that parsed on prior releases parsing. Added a test for the opt-out path.

SABITHSAHEB:
gentle ping

src/lib_json/json_reader.cpp
baylesj: SGTM, that makes sense.

baylesj:
@greptileai

src/lib_json/json_reader.cpp
```diff
@@ -1768,13 +1769,23 @@ bool OurReader::decodeUnicodeCodePoint(Token& token, Location& current,
     if (*(current++) == '\\' && *(current++) == 'u') {
       unsigned int surrogatePair;
       if (decodeUnicodeEscapeSequence(token, current, end, surrogatePair)) {
+        if (features_.rejectInvalidSurrogates_ &&
+            (surrogatePair < 0xDC00 || surrogatePair > 0xDFFF))
+          return addError("expecting a low surrogate (DC00-DFFF) to complete "
+                          "the unicode surrogate pair",
+                          token, current);
```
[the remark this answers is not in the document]
SABITHSAHEB: That's intentional, see the thread above: the legacy Json::Reader change was dropped during review because its Features struct has no setting to gate this on, and an unconditional hard-fail there would break input that parsed on every prior release. The fix is scoped to CharReader behind rejectInvalidSurrogates. I've updated the PR description so it no longer says both readers.

PR #1703, opened 2026-07-03
reject a lone sign token as a number in decodeNumber

SABITHSAHEB:
a number token that is only a sign ("-", including as an array or object element) never enters decodeNumber's digit loop, so it fell through as integer 0 instead of being rejected; route a sign-only token to decodeDouble, which rejects it, in both Reader and OurReader.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -531,10 +531,9 @@ bool Reader::decodeNumber(Token& token, Value& decoded) {
   if (isNegative)
     ++current;
   // A number token that is only a sign (e.g. "-") carries no digits; the loop
-  // below never runs and would otherwise decode it as zero. Reject it via
-  // decodeDouble, matching how a non-digit inside the token is handled.
+  // below never runs and would otherwise decode it as zero. Reject it.
   if (current == token.end_)
-    return decodeDouble(token, decoded);
+    return false;
   // TODO: Help the compiler do the div and mod at compile time or get rid of
   // them.
   Value::LargestUInt maxIntegerValue =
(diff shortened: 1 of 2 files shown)


src/lib_json/json_reader.cpp
```diff
@@ -530,6 +530,11 @@ bool Reader::decodeNumber(Token& token, Value& decoded) {
   bool isNegative = *current == '-';
   if (isNegative)
     ++current;
+  // A number token that is only a sign (e.g. "-") carries no digits; the loop
+  // below never runs and would otherwise decode it as zero. Reject it via
+  // decodeDouble, matching how a non-digit inside the token is handled.
+  if (current == token.end_)
+    return decodeDouble(token, decoded);
```
baylesj: Can you just return false here instead of invoking decodeDouble?

src/lib_json/json_reader.cpp
```diff
@@ -1575,6 +1580,11 @@ bool OurReader::decodeNumber(Token& token, Value& decoded) {
   if (isNegative) {
     ++current;
   }
+  // A number token that is only a sign (e.g. "-") carries no digits; the loop
+  // below never runs and would otherwise decode it as zero. Reject it via
+  // decodeDouble, matching how a non-digit inside the token is handled.
+  if (current == token.end_)
+    return decodeDouble(token, decoded);
```
baylesj: Same here -- if current == token.end_ and isNegative, can you just return false?

baylesj (changes requested):

src/lib_json/json_reader.cpp
SABITHSAHEB: Done, returns false directly now.

src/lib_json/json_reader.cpp
SABITHSAHEB: Same here, switched to a plain return false.

SABITHSAHEB:
any update?

PR #1709, opened 2026-07-30
fix: make the JSONCPP_USE_SECURE_MEMORY build compile and pass

### LINKED ISSUE #1399: JSONCPP_USING_SECURE_MEMORY uses memset_s which isn't portable
**Describe the bug**

Starting with https://github.com/open-source-parsers/jsoncpp/commit/30170d651c108400b1b9ed626ba715a5d95c5fd2, the library uses memset_s for the secure string allocator.

This is problematic, because:

- not all compilers implement `memset_s`
- the commit suggests this fixes a c++20 issue, but the meson (and cmake) file's c++ standard selection is not incremented from c++11

Apart from that, the unit tests won't build because many spots the test code tries to convert between std::string and the custom allocator string.

**To Reproduce**
1. activate `JSONCPP_USING_SECURE_MEMORY`
2. build the code 

**Expected behavior**

The code should compile with any C++ compiler.
The tests should compile regardless of the string allocator.

 - OS: NixOS
 - Meson version 0.57.1
 - Ninja version 1.10.2

**Additional context**

lenamonj:
`JSONCPP_USE_SECURE_MEMORY=1` does not build. It fails in three separate ways, each hidden behind the previous one, so fixing only the first gets you to the second. No CI job builds this configuration, which is how they accumulated.

**Reproduce on master (`60de77f`), MSVC 19.44 x64:**

```
cmake -S . -B sb -A x64 -DCMAKE_CXX_FLAGS="/DJSONCPP_USE_SECURE_MEMORY=1 /std:c++17"
cmake --build sb --config Debug
```

```
include\json\allocator.h(47,5): error C3861: 'RtlSecureZeroMemory': identifier not found
```

**1. The `_WIN32` branch calls `RtlSecureZeroMemory` without including `<windows.h>`.** Removing that branch lets the portable `volatile std::fill_n` path handle Windows too. The same fill was zeroing `n` bytes rather than `n * sizeof(T)`; `Json::String` is `SecureAllocator<char>` so it is unaffected, but any wider `T` through the public `Json::Allocator<T>` alias was leaving most of the allocation un-wiped.

**2. `jsontestrunner` declares a `std::string` where the surrounding code uses `Json::String`.** Identical types in default builds, different under secure memory, so the runner does not compile. One line.

**3. `CZString`'s move-assignment releases a key with `releasePrefixedStringValue`, but `CZString` keys come from `duplicateStringValue` and carry no length prefix.** The destructor already uses the matching `releaseStringValue`. Under the default allocator the mismatch is survivable; under `SecureAllocator` the suite segfaults. This is the one that only appears after the first two are fixed.

**Verification.** Before: `error C3861`. With only 1 and 2 fixed: builds, then `jsoncpp_test (SEGFAULT)`. With all three: secure build compiles and 3/3 ctest suites pass. The default build is unaffected and also 3/3.

Fixes #1399, which reported both the portability problem and that "the unit tests won't build".

### DIFF
--- a/include/json/allocator.h
+++ b/include/json/allocator.h
@@ -43,10 +43,9 @@ public:
     // unlike memset.
 #if defined(HAVE_MEMSET_S)
     memset_s(p, n * sizeof(T), 0, n * sizeof(T));
-#elif defined(_WIN32)
-    RtlSecureZeroMemory(p, n * sizeof(T));
 #else
-    std::fill_n(reinterpret_cast<volatile unsigned char*>(p), n, 0);
+    std::fill_n(reinterpret_cast<volatile unsigned char*>(p), n * sizeof(T),
+                static_cast<unsigned char>(0));
 #endif
 
     // free using "global operator delete"
--- a/src/jsontestrunner/main.cpp
+++ b/src/jsontestrunner/main.cpp
@@ -328,7 +328,7 @@ int main(int argc, const char* argv[]) {
       return modern_return_code;
     }
 
-    const std::string filename =
+    const Json::String filename =
         opts.path.substr(opts.path.find_last_of("\\/") + 1);
     const bool should_run_legacy = (filename.rfind("legacy_", 0) == 0);
     if (should_run_legacy) {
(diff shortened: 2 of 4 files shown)


baylesj:
@greptileai

baylesj:
We probably want to add a GitHub Action for this build configure to ensure that it does not regress. Added an issue:

https://github.com/open-source-parsers/jsoncpp/issues/1712

PR #1710, opened 2026-07-30
Prevent abort() call while parsing an input strings with too deep nestings

smallSwed:
When `JSON_USE_EXCEPTION` set `0` jsoncpp should not abort when the input string contains object/array which is exceeds the stackLimit. 
The caller unable to validate that the input string exceeds the nesting stackLimit without parsing the json, so abort is a bad response for this case.
When the `JSON_USE_EXCEPTION` is `1` then the caller can catch the thrown exception and handle the issue how they want it. 

With this change I intend to make the non-exception variant case fail gracefully too and handle-able.

### DIFF
--- a/src/lib_json/json_reader.cpp
+++ b/src/lib_json/json_reader.cpp
@@ -1048,10 +1048,19 @@ bool OurReader::parse(const char* beginDoc, const char* endDoc, Value& root,
 }
 
 bool OurReader::readValue() {
-  //  To preserve the old behaviour we cast size_t to int.
-  if (nodes_.size() > features_.stackLimit_)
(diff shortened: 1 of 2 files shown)


smallSwed:
I wasn't sure how to rework the error recovery, but maybe in this case it would be better for the parser to skip the whole recovering all together.
I think that could be achieved by the code chage when the stackLimit exceeded:
```cpp
    ...
    current_ = end_; // short circuit the parsing to prevent error recovery
    return addError(
    ...
```

baylesj:
@greptileai

baylesj:
This seems reasonable and is generally in alignment with our goal of not crashing due to an input error.

